Issue #57, opened 2013-07-01
Console prints messages out of order with Multicore Rendering enabled


essbuh:
With Multicore Rendering enabled, the console can occasionally print messages out of order.

For example:

> ] find client
> "cl_ent_absbox"
> "cl_ent_bbox"
>  client client cheat cheat
> - Displays the client's absbox for the entity under the crosshair.
> - Displays the client's bounding box for the entity under the crosshair.
>   "cl_ent_rbox"
>   client cheat
> - Displays the client's render box for the entity under the crosshair.
> - Find and list all client entities with classnames that contain the specified substring.
>   Format: cl_find_ent <substring>
> 
> "cl_find_ent"
> "cl_find_ent_index"
>  client client cheat cheat
> - Display data for clientside entity matching specified index.
>   Format: cl_find_ent_index <index>
> 
> ] status 
> hostname: Half-Life 2 Deathmatch
> version : 1.0.7.0/24 5337 insecure
> udp/ip  : 192.168.1.5:27015  (public ip: X.X.X.X)
> map     : nmo_broadway at: 682 x, -396 y, 137 z
> #      4 "essbuh"            (...)   05:23       14    0 active loopback
> players : 1 (16 max)
> 
> # userid name                uniqueid            connected ping loss state  adr
> ] status 
> hostname: Half-Life 2 Deathmatch
> version : 1.0.7.0/24 5337 insecure
> udp/ip  : 192.168.1.5:27015  (public ip: X.X.X.X)
> map     : nmo_broadway at: 682 x, -396 y, 137 z
> players : 1 (16 max)
> 
> #      4 "essbuh"            (...)   05:24       14    0 active loopback
> # userid name                uniqueid            connected ping loss state  adr

This does not seem to happen with multicore rendering disabled.

PR #63, opened 2013-07-02
Print chat messages to console in color

saul:
- Chat messages are now printed to console in `CBaseHudChatLine::Colorize`
- Removed `RemoveColorMarkup` function (no longer used)

**Note**: as the chat message are printed to console in chunks of color, this feature depends on #57 being fixed, otherwise the message is printed out of order.

### DIFF
--- a/mp/src/game/client/hud_basechat.cpp
+++ b/mp/src/game/client/hud_basechat.cpp
@@ -47,41 +47,6 @@ Color g_ColorYellow( 255, 178, 0, 255 );
 Color g_ColorGrey( 204, 204, 204, 255 );
 
 
-// removes all color markup characters, so Msg can deal with the string properly
-// returns a pointer to str
-char* RemoveColorMarkup( char *str )
-{
-	char *out = str;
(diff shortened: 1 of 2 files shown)


mp/src/game/client/hud_basechat.cpp
```diff
@@ -1394,7 +1348,7 @@ void CBaseHudChatLine::InsertAndColorizeText( wchar_t *buf, int clientIndex )
 	wchar_t *txt = m_text;
 	int lineLen = wcslen( m_text );
 	Color colCustom;
-	if ( m_text[0] == COLOR_PLAYERNAME || m_text[0] == COLOR_LOCATION || m_text[0] == COLOR_NORMAL || m_text[0] == COLOR_ACHIEVEMENT || m_text[0] == COLOR_CUSTOM || m_text[0] == COLOR_HEXCODE || m_text[0] == COLOR_HEXCODE_ALPHA )
```
saul, line 1397: This conditional was checking whether `m_text[0]` was a color code other than `COLOR_USEOLDCOLORS`.

PR #73, opened 2013-07-05
Added nav-mesh support

### LINKED ISSUE #39: NavMesh commands missing
I've plugged in the sample episode 2 mod, and was going to test an old navmesh from my 2007-engine mod, but I can't seem to find any of the commands. I do think if I tried, I could get them to work, but this kind of thing was readily available with the old engine, so it's a pretty big downgrade. I haven't analyzed all the differences, but I've noticed the HidingSpotList is missing, as well.

saul:
**Changes**:
- Several TF/NextBot-specific features have been `#ifdef`'d out of the nav-mesh code.
- Added a `nav_mesh.vpc` script that is included by `server_hl2mp.vpc` if `$SOURCESDK` is defined.
- Nav edit mode no longer uses `NDEBUG_PERSIST_TILL_NEXT_SERVER` as the overlay duration. This is too long and causes overlapping, increased fillrate and flickering. All durations have been set to 0 (i.e., until the next frame).

This pull request fixes #39.
## 

**Issues**:
- [x] The `tier1` _code_ in the repository is missing the LZMA decoder. Therefore if you build your own `tier1`, there are unresolved external symbol errors at link time. **Workaround**: Use the Valve-compiled `tier1.lib` supplied in the repository. **Fixed** in 9fcd93a and b791326.
- [x] Port changes over to the `sp` branch. **Will be completed by @JoeLudwig when pushing.**
## 

**Note to Valve devs**: when merging this pull request into your own codebase, be sure to add the `USE_NAV_MESH` pre-processor definition to your TF VPC script (if you're not including `nav_mesh.vpc`). This is to ensure the nav-mesh code gets built in `gameinterface.cpp` (see the changes for more info).

### DIFF
--- a/mp/src/game/server/nav_file.cpp
+++ b/mp/src/game/server/nav_file.cpp
@@ -1407,18 +1407,17 @@ NavErrorType CNavMesh::Load( void )
 		}
 	}
 
-	if ( IsX360() )
+#ifdef _XBOX
+	// 360 has compressed NAVs
+	CLZMA lzma;
+	if ( lzma.IsCompressed( (unsigned char *)fileBuffer.Base() ) )
 	{
-		// 360 has compressed NAVs
-		CLZMA lzma;
-		if ( lzma.IsCompressed( (unsigned char *)fileBuffer.Base() ) )
-		{
-			int originalSize = lzma.GetActualSize( (unsigned char *)fileBuffer.Base() );
-			unsigned char *pOriginalData = new unsigned char[originalSize];
-			lzma.Uncompress( (unsigned char *)fileBuffer.Base(), pOriginalData );
-			fileBuffer.AssumeMemory( pOriginalData, originalSize, originalSize, CUtlBuffer::READ_ONLY );
-		}
+		int originalSize = lzma.GetActualSize( (unsigned char *)fileBuffer.Base() );
+		unsigned char *pOriginalData = new unsigned char[originalSize];
+		lzma.Uncompress( (unsigned char *)fileBuffer.Base(), pOriginalData );
+		fileBuffer.AssumeMemory( pOriginalData, originalSize, originalSize, CUtlBuffer::READ_ONLY );
 	}
+#endif
 
 	// check magic number
 	unsigned int magic = fileBuffer.GetUnsignedInt();


mp/src/game/server/nav_area.cpp
```diff
@@ -2778,7 +2778,7 @@ void CNavArea::Draw( void ) const
 	NavEditColor color;
 	bool useAttributeColors = true;
 
-	const float DebugDuration = NDEBUG_PERSIST_TILL_NEXT_SERVER;
```
JoeLudwig, line 2781: What is this change about? Was NDEBUG_PERSIST_TILL_NEXT_SERVER undefined?  You could include public/engine/ivdebugoverlay.h  to get it.

mp/src/game/server/nav_colors.cpp
```diff
@@ -79,8 +79,8 @@ void NavDrawLine( const Vector& from, const Vector& to, NavEditColor navColor )
 	const Vector offset( 0, 0, 1 );
 
 	Color color = NavColors[navColor];
-	NDebugOverlay::Line( from + offset, to + offset, color[0]/2, color[1]/2, color[2]/2, true, 0.1f );
-	NDebugOverlay::Line( from + offset, to + offset, color[0], color[1], color[2], false, 0.15f );
```
JoeLudwig, line 83: Why change these colors?  (Though I admit that dividing color values by 2 is awfully strange.)

mp/src/game/server/nav_edit.cpp
```diff
@@ -732,8 +732,10 @@ void CNavMesh::DrawEditMode( void )
 	static ConVarRef host_thread_mode( "host_thread_mode" );
 	host_thread_mode.SetValue( 0 );
 
+#ifdef TERROR
 	static ConVarRef sb_perf_collect( "sb_perf_collect" );
 	sb_perf_collect.SetValue( 0 );
+#endif
```
JoeLudwig, line 738: TF2 has these lines but doesn't have sb_perf_collect. SetValue on a convar ref will basically throw away the value if the convar doesn't exist. Was this causing some problem?

mp/src/game/server/nav_file.cpp
```diff
@@ -1315,18 +1315,17 @@ void CommandNavCheckFileConsistency( void )
 		}
 	}
 	
-	if ( IsX360() )
+#ifdef _XBOX
```
JoeLudwig, line 1318: Turns out the LZMA decoder is public domain. I'll add it to the SDK instead of adding these ifdefs.

mp/src/game/server/nav_colors.cpp
saul, line 83: I did not change the colours, only the debug overlay duration (see my pull request for more info on durations).

mp/src/game/server/nav_edit.cpp
saul, line 738: It seems like a leftover from L4D being merged into main. `sb_perf_collect` is a Left4Dead survivor bot ConVar, hence the `#ifdef TERROR`.

As you stated, this simply removes a ConVarRef warning in the console.

mp/src/game/server/nav_colors.cpp
JoeLudwig, line 83: So you did. I misread the diff.

mp/src/game/server/nav_area.cpp
saul, line 2781: `NDEBUG_PERSIST_TILL_NEXT_SERVER` is _supposed_ to make debug overlays only persist until the next time the edit interface is drawn again. However, the duration is too long and causes the debug overlays to overlap several times.

Adding a debug ConVar (e.g., `nav_overlay_duration_test`) and replacing the line with:

```
const float DebugDuration = nav_overlay_duration_test.GetBool() ? NDEBUG_PERSIST_TILL_NEXT_SERVER : 0.0f;
```

And then toggling the value of `nav_overlay_duration_test` will show exactly what I mean (notice the flickering too).

mp/src/game/server/nav_edit.cpp
JoeLudwig, line 738: I'm going to solve this with my delete key. :)

PR #179, opened 2013-09-04
Linux port of VBSP, VVIS, and VRAD v2 (Part 1)

sortie:
Hi,

I have ported fgdlib, raytrace, vbsp, vvis, and vrad to Linux and I'd like to upstream this effort so others can benefit from it as well. My ported tools are able to compile maps that can be correctly run by the Linux 2013 SDK base. This is version 2 of the #160 pull request, see it for more details.

This first part fixes a considerable number of non-portable constructs and warnings in the Source SDK codebase. The main port requires most of these smaller fixes.  I have arranged this pull request such that the most important fixes come first, although it'd be preferable if all of it was merged. I realize how the Source SDK is currently built such that some of these warnings are suppressed, nonetheless it would be good to actually fix these warnings as some can be considered minor bugs. There is a few non-trivial C++ 'bugs' in the codebase that is only exposed at -Wextra that would be well worth fixing as language doesn't mandate what is expected.

I've read the CONTRIBUTING document and I hereby assert that I agree to its terms and that this is my own personal work. If you think I can improve this patch set so it's more suitable for upstreaming, please let me know. It'll probably take some time to review the changes, so I figured I'd send them off before the next part is ready.

The following parts will be arriving soon, I need to improve the port a bit, but it currently works well. I am able to correctly compile sdk_d1_trainstation_01.vmf and play the first map of HL2 normally under Linux. Thanks for releasing fgdlib and raytrace such that this work was possible!

### DIFF
--- a/mp/src/utils/vbsp/portals.cpp
+++ b/mp/src/utils/vbsp/portals.cpp
@@ -1178,8 +1178,8 @@ void FindPortalsLeadingToArea_R(
 		if( !p->nodes[0]->occupied || !p->nodes[1]->occupied )
 			continue;
 	
-		if( p->nodes[1]->area == iDestArea && p->nodes[0]->area == iSrcArea ||
-			p->nodes[0]->area == iDestArea && p->nodes[1]->area == iSrcArea )
+		if( (p->nodes[1]->area == iDestArea && p->nodes[0]->area == iSrcArea) ||
+			(p->nodes[0]->area == iDestArea && p->nodes[1]->area == iSrcArea) )
 		{
 			// Make sure the plane normals point the same way.
 			plane_t *pMapPlane = &g_MainMap->mapplanes[p->onnode->planenum];
--- a/mp/src/utils/vbsp/textures.cpp
+++ b/mp/src/utils/vbsp/textures.cpp
@@ -251,7 +251,7 @@ int	FindMiptex (const char *name)
 			g_bHasWater = true;
 		}
 		const char *pShaderName = GetMaterialShaderName(matID);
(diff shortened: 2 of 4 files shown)


floube:
Hi @sortie 

would it be possible for you to upload the executables for vbsp, vvis and vrad which run on Linux? Unfortunately the generated makefile isn't working since some C++ files still use code from the Win32 API.

Cheers

thorsummoner:
Tried compiling vbsp on ubuntu 14.04 LTS, `gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4`, 

```
$ ln -sf chunkfile.h public/ChunkFile.h
$ CPLUS_INCLUDE_PATH=./public:./public/tier0:./public/tier1:./utils/common make utils/vbsp/vbsp
```

this pr's head introduces these additional compiler errors for vbsp.cpp, compared to the current master.
```
In file included from ./public/tier1/utlmemory.h:20:0,
                 from ./public/tier1/utlrbtree.h:12,
                 from ./public/tier1/utlsymbol.h:17,
                 from ./public/filesystem.h:13,
                 from ./utils/common/cmdlib.h:35,
                 from utils/vbsp/vbsp.h:13,
                 from utils/vbsp/vbsp.cpp:8:
./public/mathlib/mathlib.h:1220:2: error: #error Unknown architecture
 #error Unknown architecture
./public/tier0/platform.h:1089:1: error: ‘DLL_IMPORT’ does not name a type
 PLATFORM_INTERFACE char *   Plat_asctime( const struct tm *tm, char *buf );
 ^
```

I cant even get vbsp to compile though, not to mention all the supporting libraries.

PR #257, opened 2014-04-14
Fixed clang-found semantic error in c_te_decal.cpp

hikari-no-yume:
Fixes build for OS X users!


Bitl:
This fix was recently implemented in the latest Source SDK build, without any refrence to this pull request 

https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/game/client/c_te_decal.cpp#L142

PR #300, opened 2014-11-25
Changed delay parameter for ent_fire

EricBecker:
Originally, the delay param was an integer, and supplying any decimal value as a delay (e.g. 2.5 seconds) would simply round down (e.g., to **2** seconds). This was basically due to the fact that it was set as an integer.


ZehM4tt:
Theres no difference to it if you still use atoi on assignment?

PR #313, opened 2015-02-26
Fixed several +use bugs.

tgnottingham:
This request involves several fixes to the +use functionality. Apologies for the large diff. I found it too tempting to fix all the +use issues at once, since they were all in one (large!) function. I primarily wanted to issue this pull request to see if there was any possibility of any part of it being accepted. If that is the case, please don't hesitate to ask me to break it up into more granular pull requests.

* Fixed not being able to +use objects inside or beyond clip brushes.
    * E.g. in DoD:S, you could not pickup live grenades, etc. if they are in a clip brush. This is actually the issue that motivated me to look at the +use code to begin with.
* Fixed being able to +use NPCs out of +use range in some cases.
    * E.g. if standing close enough to +use one NPC, you could aim over his/her shoulder and +use another NPC that was very far away.
* Fixed sv_debug_player_use showing incorrect overlays in some cases.
    * E.g. when using an NPC behind another NPC, the debug visualization showed the trace line hitting the front NPC.
* Fixed some questionable math used to determine object distances.
    * E.g. the third argument in "nearestDist = CalcDistanceToLine( point, searchCenter, forward );" was in all likelihood intended to be "searchCenter + forward", so that nearestDist was the distance from "point" to the line described by the forward view vector starting at "searchCenter" (the eye position). This and other similar calculations were changed to consistently use a different but simple distance approximation.
* Refactored and commented code used to find the entity being +used.
    * E.g. removed initial 8-direction line trace, since it was likely not working as intended. Unless the initial forward trace hit a usable object (in which case it was returned immediately), the last usable object hit in the remaining 7 traces was considered the nearest object, even if it was not the nearest. The initial forward line trace, followed the sphere/cone-based query, etc. seemed adequate.
    * Made logging behavior and text consistent.
    * Revised comments.

### DIFF
--- a/mp/src/game/shared/baseplayer_shared.cpp
+++ b/mp/src/game/shared/baseplayer_shared.cpp
@@ -1065,205 +1065,218 @@ float IntervalDistance( float x, float x0, float x1 )
 	return 0;
 }
 
+float EstimatedDistanceSquared(const Vector &point, const CBaseEntity *pEntity)
+{
+	Vector nearestPoint;
+	pEntity->CollisionProp()->CalcNearestPoint( point, &nearestPoint );
+	return (nearestPoint - point).LengthSqr();
+}
+
 CBaseEntity *CBasePlayer::FindUseEntity()
 {
-	Vector forward, up;
-	EyeVectors( &forward, NULL, &up );
-
-	trace_t tr;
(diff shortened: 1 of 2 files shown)


Filip98:
your efforts are not wasted

PR #331, opened 2015-08-15
Lag compensation now supports pose parameters

AltimorTASDK:
Fixes the lack of pose parameter lag compensation shown here in CSGO:
https://www.youtube.com/watch?v=I0_xRnxe-sE

Hitboxes are now pretty tight doing the same thing I did in the video. I took that picture with 75ms of lag added by client interpolation.

### DIFF
--- a/mp/src/game/server/player_lagcompensation.cpp
+++ b/mp/src/game/server/player_lagcompensation.cpp
@@ -41,6 +41,7 @@ ConVar sv_unlag_fixstuck( "sv_unlag_fixstuck", "0", FCVAR_DEVELOPMENTONLY, "Disa
 // Purpose: 
 //-----------------------------------------------------------------------------
 #define MAX_LAYER_RECORDS (CBaseAnimatingOverlay::MAX_OVERLAYS)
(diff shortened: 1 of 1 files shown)


formatme:
it's been right here, hellooo

AltimorTASDK:
@formatme This bug was fixed a while ago, the recent bug was different

mp/src/game/server/player_lagcompensation.cpp
```diff
@@ -112,6 +114,7 @@ struct LagRecord
 	LayerRecord				m_layerRecords[MAX_LAYER_RECORDS];
 	int						m_masterSequence;
 	float					m_masterCycle;
+	float					m_poseParameters[MAX_POSE_PARAMETERS];
```
sigsegv-mvm, line 117: You're probably going to want to add a `for` loop in the copy constructor to copy the `m_poseParameters` values over from `src` (similar to what's done for `m_layerRecords` in the copy ctor).

XutaxKamay:
9 years after, I noticed something `flags |= LC_POSE_PARAMS_CHANGED;` is missing right?

Because it checks if the flags but it does not set them at all...

PR #337, opened 2015-09-15
Convert all ISO-8859-1 text to utf-8

dkreuter:
This commit replaces ISO-8859-1 encoded text with utf-8 (excluding protobuf, which has been converted upstream)

Only sp/game/mod_episodic/resource/mod_episodic_english.txt and sp/game/mod_hl2/resource/mod_hl2_english.txt were utf-16 + bom before.


thorsummoner:
awesome <3

PR #343, opened 2015-10-09
Enable precache for overrides in scripts/mvm_level_sounds.txt

sigsegv-mvm:
*This is a TF-specific fix, so I'm not 100% confident that you want me to submit a pull request here. But since the relevant code is actually located in the public part of the Source SDK, it made a certain amount of sense to submit the fix here. I've also emailed the TF team with a link to this pull request to make sure they're aware of it.*

Based on commented-out lines in TF2's game_sounds_manifest.txt, it's clear that
mvm_level_sounds.txt was originally intended to be precached; when the file was
changed to be loaded as a level-specific sound script override, the precache
flag was apparently forgotten.

This patch fixes several sounds which have never worked in MvM mode:
- SV_StartSound: player/spy_shield_break.wav not precached (0)
- SV_StartSound: player/medic_charged_death.wav not precached (0)
- SV_StartSound: )player/flame_out.wav not precached (0)
- SV_StartSound: physics/concrete/concrete_impact_flare1.wav not precached (0)

### DIFF
--- a/mp/src/game/shared/SoundEmitterSystem.cpp
+++ b/mp/src/game/shared/SoundEmitterSystem.cpp
@@ -278,7 +278,7 @@ public:
 			V_strncpy( scriptfile, "scripts/mvm_level_sounds.txt", sizeof( scriptfile ) );
 			if ( filesystem->FileExists( "scripts/mvm_level_sounds.txt", "GAME" ) )
 			{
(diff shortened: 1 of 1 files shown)


sigsegv-mvm:
Looks like this was fixed in TF2's 20151028 update.

Nephyrin:
Yeah, this was fixed on the TF2 side -- thanks for bringing this to our attention!

PR #363, opened 2015-12-28
Fix false positive in netadr_t::IsReservedAdr

willox:
For some reason the range "192.168.0.0 -> 192.255.255.255" is treated as reserved.

It looks like it was only a typo.


LestaD:
+1

PR #379, opened 2016-06-10
Remove useless if condition.

Li4ick:
If conditions are short circuited. The "armor" condition doesn't do anything.


GuestSneezeOSDev:
Hello there Source SDK 2013 will not be using any PR's thank you

PR #380, opened 2016-06-11
Fixed infinite recursion by using incorrect key on value conversation.

ZehMatt:


SC1040-TS2:
Necroposter here. Based on some digging, this issue is also present in https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/game/shared/baseentity_shared.cpp#L411 as well.

This reply is being made for further documentation purposes for singleplayer developers reading this repository.

PR #391, opened 2016-09-20
Fix smoothing groups not working on detail brushes.

tgnottingham:
Before:
detail_before

After:
detail_after

This is a low risk change. As far as I can tell, there is one code path that executes the modified function.


SirYodaJedi:
Bump. Could this be updated to the new codebase? Would love to see this pushed upstream.

Unusuario2:
This PR should be closed in favour of #1346

PR #395, opened 2016-10-22
Fixed a compilation error in VS2015.

PazerOP:
Fixed ambiguous call to abs()


Filip98:
is the code now fully compatible with VS2015 and backwards compatible with VS2013?

joeangry:
It can't be until Valve releases updated libraries for things we don't have the source code for. It might compile fine but it won't link.

Filip98:
damn, forgot about that.. well we're doomed until source 2, nothing new

joeangry:
So exactly what is the point of your commits @PazerOP?

LestaD:
@PazerOP Maybe squash all your commits in one?

PazerOP:
What the fuck, why did all these commits get added to this pull request

PR #396, opened 2016-11-05
Fixed a memory allocation error

PazerOP:
Fixed a case where reallocated memory would be wiped clean, breaking a
lot of code that (correctly) assumes data will still be there after the
realloc.


SC1040-TS2:
> Oops.

Necroposter here. Was this ever fixed to your knowledge, or was the PR closed because it received no attention from Valve?

PR #400, opened 2016-11-27
Add credits ConVar for easy editing.

Drixevel:
Not all mods will call for this but mine does.

Adds the ConVar 'cl_credits_file' which by default is 'scripts/credits.txt' to edit if need be.


Drixevel:
I must have misclicked somewhere, wanted to push this to the wrong place.

PR #416, opened 2017-04-11
sp/src/game/shared/ragdoll_shared.cpp: PVS-Studio: NULL Pointer Dereference

SvyatoslavRazmyslov:
We have found and fixed a bug using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/

Analyzer warnings: 

- V595 The 'pRagdoll' pointer was utilized before it was verified against nullptr. Check lines: 883, 884. ragdoll_shared.cpp 883
- V595 The 'pRagdoll' pointer was utilized before it was verified against nullptr. Check lines: 924, 925. ragdoll_shared.cpp 924

We suggests having a look at the emails, sent from @pvs-studio.com.

### DIFF
--- a/sp/src/game/shared/ragdoll_shared.cpp
+++ b/sp/src/game/shared/ragdoll_shared.cpp
@@ -877,15 +877,16 @@ void CRagdollLRURetirement::Update( float frametime ) // EPISODIC VERSION
 	
 		for ( i = m_LRU.Head(); i < m_LRU.InvalidIndex(); i = next )
 		{
-			CBaseAnimating *pRagdoll = m_LRU[i].Get();
-
 			next = m_LRU.Next(i);
-			IPhysicsObject *pObject = pRagdoll->VPhysicsGetObject();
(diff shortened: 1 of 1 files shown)


SC1040-TS2:
> We have found and fixed a bug using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/
> 
> Analyzer warnings:
> 
> * V595 The 'pRagdoll' pointer was utilized before it was verified against nullptr. Check lines: 883, 884. ragdoll_shared.cpp 883
> * V595 The 'pRagdoll' pointer was utilized before it was verified against nullptr. Check lines: 924, 925. ragdoll_shared.cpp 924
> 
> We suggests having a look at the emails, sent from @pvs-studio.com.

Necroposter here. This also seems to be present in https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/game/shared/ragdoll_shared.cpp#L883 as well. It seems to be more widespread than previously thought.

PR #419, opened 2017-04-16
Fixed alignment error when LTCG is enabled

KrazyKasrkin:
Fixed an alignment error/warning in mathlib when building the SDK with Link-time Code Generation enabled (/LTCG)

Needed to be consistent with the `power2_n` definition in color_conversion.cpp


SC1040-TS2:
> Fixed an alignment error/warning in mathlib when building the SDK with Link-time Code Generation enabled (/LTCG)
> 
> Needed to be consistent with the `power2_n` definition in color_conversion.cpp

For documentation purposes, this also needs to be duplicated for the SP mathlib.h code.

PR #434, opened 2017-09-11
Fixed bug in CCommand constructor that could append junk to the comma…

KrazyKasrkin:
…nd string

* Missed copying the null terminator
* Since m_pArgSBuffer is not initialized this problem can occur when the (command string buffer length) + 1 char in m_pArgSBuffer array happens to be non-zero.

### DIFF
--- a/mp/src/tier1/convar.cpp
+++ b/mp/src/tier1/convar.cpp
@@ -359,7 +359,7 @@ CCommand::CCommand( int nArgC, const char **ppArgV )
 		{
 			*pSBuf++ = '\"';
 		}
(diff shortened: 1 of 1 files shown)


SC1040-TS2:
> …nd string
> 
> * Missed copying the null terminator
> * Since m_pArgSBuffer is not initialized this problem can occur when the (command string buffer length) + 1 char in m_pArgSBuffer array happens to be non-zero.

Necroposter here. This also seems to affect the SP/src/tier1/convar.cpp as well. A programmer should have been pinged for this.

PR #448, opened 2018-03-07
Fix Linux soundscapes

TotallyMehis:
The soundscape keyvalues loaded using IFileSystem::LoadKeyValues returns incorrect data and fails to setup soundscapes correctly. This means soundscapes don't work at all, in any map.

[This has been reported to other mods](http://steamcommunity.com/app/265630/discussions/2/535151589882790386/), surprisingly nobody has brought it up here. Don't know if this has any effect on SP. Haven't investigated if it's because of IFileSystem::TYPE_SOUNDSCAPE.

### DIFF
--- a/mp/src/game/server/soundscape_system.cpp
+++ b/mp/src/game/server/soundscape_system.cpp
@@ -62,7 +62,7 @@ void CSoundscapeSystem::AddSoundscapeFile( const char *filename )
 	MEM_ALLOC_CREDIT();
 	// Open the soundscape data file, and abort if we can't
 	KeyValues *pKeyValuesData = new KeyValues( filename );
-	if ( filesystem->LoadKeyValues( *pKeyValuesData, IFileSystem::TYPE_SOUNDSCAPE, filename, "GAME" ) )
+	if ( pKeyValuesData->LoadFromFile( filesystem, filename, "GAME" ) )
 	{
 		// parse out all of the top level sections and save their names
 		KeyValues *pKeys = pKeyValuesData;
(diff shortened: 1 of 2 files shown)


AdRiAnIlloO:
Hi,
I spent hours finding the root of the issue, which I had too. And it is far from what it seems.

The problem is that such mod developers (and I) added a call to `KeyValues::SetStringGrowable(true)` 
`KeyValues::SetStringGrowable(true)` which triggers undefined behaviour on KeyValue reads of KV's created before such call, resulting in empty key that we see in consolees-related case that we see in console. That function changes the KeyValues access algorithm **globally** from string tree to string table (hash). This is well documented and warned within the `KeyValues.h` header.

`CSoundscapeSystem::Init()` is called when game/server starts and during level startups. The first time it creates a KeyValues for caching the soundscape entries on the next level startups (given that soundscapes and other sysre freed when level is shutdown). In this sense, our `KeyValues::SetStringGrowable(true)` is naturally done after first `CSoundscapeSystem::Init()` execution and making the problem manifest on subsequent calls to the latter.

Removing the mentioned `SetStringGrowable` calls solves the problem. I highly doubt there are other source of bad codes causing this concrete issue.

Greetings,
Adrián

TotallyMehis:
Did you mean [KeyValues::SetUseGrowableStringTable](https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/public/tier1/KeyValues.h#L79)? At no point do we use it in the game code.

AdRiAnIlloO:
Hi back. Yes, I corrected it. I'm sorry for assumming such situation. I want to inform you that after dealing with my KV's conflicting issue, I don't get the soundscapes error output in console in my Linux server, if useful.

toizy:
I use 

`if (pKeyValuesData->LoadFromFile(filesystem, filename, "GAME"))`

in 'CSoundscapeSystem::AddSoundscapeFile' instead. I have no idea if this is a right correction, but it worked for me so I just let it go.

PR #459, opened 2018-07-05
Fix smooth friction sounds using the incorrect soundhandle

Kefta:
Seems like a simple copy-paste typo


thegrb93:
Nice catch

TotallyMehis:
Duplicate of #441
You also didn't fix the MP branch.

PR #502, opened 2020-09-09
Fix typo in logical expression

GamerDude27:
A bitwise operator (`|`) is used instead of a logical one (`||`).
This means that the right operand will be evaluated regardless of the result of the left operand.


dajeed (approved):

EricS-Valve:
Manually merged

PR #586, opened 2024-08-31
Fixed out of bounds read with predef_flexcontroller_names array

tyabus:
Found by PVS-Studio


tyabus:
Fixed in the latest SDK update

PR #590, opened 2024-09-14
Fix for swept box rays that were never really working properly with OBBs. (IntersectRayWithOBB)

XutaxKamay:
IntersectRayWithOBB has a bug where it would somehow ignore angle information inside the matrix.

Why? I don't know. Maybe some developers not wanting to use VectorTransform somehow ? (or overworked)

This is probably present in all Source games (including Source 2) if it hasn't been fixed yet. (?)

This fixes some edge cases where for example a game mod wants to have thicker bullets or other stuffs that I'm not aware of (I discovered it by accident by working on the engine).


Bitl:
This seems like a similar fix to https://github.com/ValveSoftware/source-sdk-2013/pull/552

does this do anything different in comparison to that

XutaxKamay:
> This seems like a similar fix to #552
> 
> does this do anything different in comparison to that

It was different at first, but I haven't noticed the commits from June 2024, before the fix it was only for the plane calc and not for the missing transformation

PR #616, opened 2025-02-18
gameplay: allow non-giants to be tickled in MvM

mastercoms:
Holiday Punch is currently completely non-functional in the mode otherwise, and common bots can already be stunned by other sources. it's only giants that are supposed to be immune.


Mikusch:
I am not a fan of this, robots do not have animations for laughing and would just awkwardly stand still while playing a laugh voice line. There are animations for bonk stuns inflicted by Sandman, Thermal Thruster, Sapper, etc. so it's not really comparable.

mastercoms:
Right, but that patch was in response to a video iirc that showed an exploit with stunning boss bots, I think this one: https://youtu.be/zAqyQnpl2ds?t=39. Before that point, it was okay to stun normal robots using the holiday punch for years despite there being no animation. Like is said in the original post, normal bots get stunned from other sources so it's not necessarily a deliberate balance choice. As it stands the holiday punch does nothing in MvM which is awkward.

PR #618, opened 2025-02-18
[TF2] gameplay: automatically collect MvM money in respawn room

mastercoms:
MvM money is automatically collected in various cases where its position is unfair for players to collect themselves.

There is a case where robots can drop money in their own spawn, which can prove hard or impossible to acquire, so add an additional check for money that is in a spawn room, just like it is already done for trigger_hurt.

### DIFF
--- a/src/game/server/tf/entity_currencypack.cpp
+++ b/src/game/server/tf/entity_currencypack.cpp
@@ -17,6 +17,7 @@
 #include "particle_parse.h"
 #include "player_vs_environment/tf_population_manager.h"
 #include "collisionutils.h"
+#include "func_respawnroom.h"
 #include "tf_objective_resource.h"
 
 //=============================================================================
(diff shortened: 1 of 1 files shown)


Mikusch:
In my opinion it would be cleaner to expand the above check to also check for the nav flags, something like this:
```cpp
CTFNavArea *area = (CTFNavArea *)TheNavMesh->GetNavArea( GetAbsOrigin() );
if ( !area || area->HasAttributeTF(TF_NAV_SPAWN_ROOM_BLUE | TF_NAV_SPAWN_ROOM_RED) )
{
	TFGameRules()->DistributeCurrencyAmount( m_nAmount );
	m_bTouched = true;
	UTIL_Remove( this );

	return;
}
```

mastercoms:
Sure, I just thought it appropriate since `trigger_hurt` has the same check.

flurbury:
Hi, there seems to be a regression with this; many custom MvM maps opt to group the func_respawnroom brushes into a single entity, causing them to have a massive bounding box spanning the entire map. This is most noticable on Hoovydam.

mastercoms:
Ack, that's no good! My apologies. The solution @Mikusch proposed would likely be suitable for fixing this.

PR #621, opened 2025-02-18
[TF2] fix: restore tf_parachute_maxspeed_onfire_z customization

mastercoms:
The BASE Jumper rising when on fire was removed in a balance update, and the cvar which controls this behavior has since been unused. Instead, set `tf_parachute_maxspeed_onfire_z`'s default value to the current game behavior for the BASE jumper (no difference from `tf_parachute_maxspeed_z`), and integrate it into the new code so that servers may integrate the old, novel behavior of rising when on fire by modifying the variable if they so please.


mastercoms:
Closing this for now since the code has some checks now which conflict with this behavior.

PR #622, opened 2025-02-18
[TF2] fix: scout balls computing critical hits incorrectly

mastercoms:
Scout balls were computing crits in their own entity think outside of the player command, which made the random generation invalid, since crit calcs are only valid during attacks. This caused random crits to be computed incorrectly, and also allowed for an exploit where you could store a random crit with the ball and permanently crit every ball thereafter.

Technically, you could seed more crit chances by using primary fire and rolling more crits before the ball actually launched in the think. However in practice, there is no way to influence the crit between the secondary attack command and the ball think running, because the primary attack time is delayed by 0.25 seconds, which is longer than the launch delay.

For safety, there's two approaches that could solve this:

1. `tf_scout_bat_launch_delay` could be integrated into the primary attack delay,
2. `tf_scout_bat_launch_delay` could get a max limit of 0.25 seconds to avoid setting it higher.

But I'm not sure if this is required (as any modifications to gameplay variables should come at the risk of the community server owner / modder). Regardless, I've added option 1 as an additional commit.

### DIFF
--- a/src/game/shared/tf/tf_weapon_bat.cpp
+++ b/src/game/shared/tf/tf_weapon_bat.cpp
@@ -271,6 +271,8 @@ void CTFBat_Wood::SecondaryAttack( void )
 		SecondaryAttackAnim( pPlayer );
 		SendWeaponAnim( ACT_VM_PRIMARYATTACK );
 
+		CalcIsAttackCritical();
+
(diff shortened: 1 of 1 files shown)


SomeBloodyMinger:
This enables Scout a decent melee option other than the atomizer tbh. If they remove it, they should buff the wrap assassin at-least in some way (especially in the absence of the old sandman). Granted it was never useless, just underwhelming on its own.

PR #625, opened 2025-02-19
[TF2] fix: add a cvar to disable equip action item prompts

mastercoms:
Veteran players may want to equip another action item like the contracker, or in the case of some community MvM servers, equip the grappling hook or spell book for custom gameplay. So follow their request of adding an option to remove the prompt.


misyltoad:

mastercoms:
Yes, I have. I accept and agree to them.

mastercoms:
Since it's more an advanced setting that maybe shouldn't be in the UI, I didn't want to archive it to make sure the player is explicitly setting it in their own config. If it does make it to the UI or that isn't a concern, then yes it should be archived.

PR #406, opened 2017-01-27
Performance+feature fixes for glow outlines

PazerOP:
Improved performance and fixed "glow when occluded" and "glow when unoccluded" functionality.

PR #627, opened 2025-02-19
[TF2] gameplay: detonator jumps aren't considered self dmg jumps

mastercoms:
Only blast damage is considered for the `bSelfBlastDmg` modifier. This affects a few things: applying the `rocket_jump_dmg_reduction` attribute, blocking self bleed, but most importantly, it allows to bypass the no knockback attribute on QuickFix ubers. Before this change, Pyros would be blocked from blast jumping with their own Detonator flares while under QuickFix uber.

### DIFF
--- a/src/game/shared/tf/tf_gamerules.cpp
+++ b/src/game/shared/tf/tf_gamerules.cpp
@@ -7405,7 +7405,8 @@ float CTFGameRules::ApplyOnDamageAliveModifyRules( const CTakeDamageInfo &info, 
 			}
 		}
 
-		if ( pAttacker == pVictimBaseEntity && (info.GetDamageType() & DMG_BLAST) &&
+		if ( pAttacker == pVictimBaseEntity &&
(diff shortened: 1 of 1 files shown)


JohnCSimon:
@mastercoms - from the contribution guide
```

Valve is happy to accept pull requests and issues in the source-sdk-2013
repository in these cases:
 * Changes that fix bugs in the SDK deployment process itself. The repository
   should build out of the box, and anything that prevents that is a pull
   request we want.
 * **High priority bugs in HL2,** the Episodes, or HL2MP that can be fixed in
   client.dll or server.dll.

For other changes we suggest that you issue a pull request to one of these
fine community-maintained repositories instead: 
  https://developer.valvesoftware.com/wiki/Source-sdk-2013-community-repos

```

dunno how high priority these are

mastercoms:
That's an old file that's been there for ages, and I figured in any case, the diffs for changes would be easier to pull in with this vs. other ways community fixes have been contributed in the past. Plus, patches that affect TF2 have been engaged with in the past, including #406.

wgetJane:
imo still worth having these pull requests up for ppl that wanna make tf2-based mods, even if valve doesn't plan on merging them

PR #634, opened 2025-02-19
[TF2] Fix huntsman arrows treating hitboxes as axis-aligned boxes rather than as oriented boxes

wgetJane:
fixes ValveSoftware/Source-1-Games/issues/5904

unfixed:

https://github.com/user-attachments/assets/fc104f0a-b448-422c-9851-449a9d989a83

fixed:

https://github.com/user-attachments/assets/bb9ccfe9-05a7-493f-aa5d-44d73efc6c97

### DIFF
--- a/src/game/server/tf/tf_projectile_arrow.cpp
+++ b/src/game/server/tf/tf_projectile_arrow.cpp
@@ -832,7 +832,7 @@ void CTFProjectile_Arrow::ArrowTouch( CBaseEntity *pOther )
 			Ray_t ray;
 			ray.Init( start, position );
 			trace_t tr;
-			IntersectRayWithBox( ray, position+pbox->bbmin, position+pbox->bbmax, 0.f, &tr );
+			IntersectRayWithOBB( ray, position, angles, pbox->bbmin, pbox->bbmax, 0.f, &tr );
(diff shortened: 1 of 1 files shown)


wgetJane:
> I... Personally think the huntsman needs to have this hitbox behaviour fixed in general. To NOT use the melee hitbox period.

i think that's a different issue imo

this is a bug fix and that's a balance change

movercell:
> I... Personally think the huntsman needs to have this hitbox behaviour fixed in general. To NOT use the melee hitbox period.

Honestly that will probably make it too hard to use for a normal player

Trico-Everfire:
> > I... Personally think the huntsman needs to have this hitbox behaviour fixed in general. To NOT use the melee hitbox period.
> 
> Honestly that will probably make it too hard to use for a normal player

That's more a fundemental problem with the weapon/the player's own skill than it is with the issue at hand. I don't think teleporting instant kill arrows are an intended feature and if they are that has got to be the dumbest thing since random crits.

wgetJane:
pls do not turn this pr's comments into a balance debate, this is for a bug fix

wgetJane:
lol forgot that renaming branches closes these

PR #637, opened 2025-02-19
[TF2] Fix sniper laser dot position being calculated inaccurately on client

wgetJane:
unfixed:

https://github.com/user-attachments/assets/6fa1b486-0f5c-4557-83a8-8283d806b84a

fixed:

https://github.com/user-attachments/assets/5c947721-107b-48f8-a41f-410c98eb856d

### DIFF
--- a/src/game/shared/tf/tf_weapon_sniperrifle.cpp
+++ b/src/game/shared/tf/tf_weapon_sniperrifle.cpp
@@ -1541,8 +1541,8 @@ bool CSniperDot::GetRenderingPositions( C_TFPlayer *pPlayer, Vector &vecAttachme
 		{
 			// Take the owning player eye position and direction.
 			vecAttachment = pPlayer->EyePosition();
-			QAngle anglesEye = pPlayer->EyeAngles();
-			AngleVectors( anglesEye, &vecDir );
+			vecDir = GetAbsOrigin() - vecAttachment;
+			VectorNormalize( vecDir );
 		}
 
 		trace_t tr;


mastercoms:
I think this makes spectating Snipers in first person look weird.

wgetJane:
> I think this makes spectating Snipers in first person look weird.

weird how? like it's delayed?

probably would be fixed by just using the old behaviour if the local player's observer mode is in_eye or something and observer target is pPlayer

mastercoms:
The Sniper dot moves ahead/more precisely than the crosshair, and yeah, I feel like that would be a fine fix. A better but more complicated fix would be to use the TF player specific high resolution eye angles for the spectator view angles instead of the default.

wgetJane:
> The Sniper dot moves ahead/more precisely than the crosshair, and yeah, I feel like that would be a fine fix. A better but more complicated fix would be to use the TF player specific high resolution eye angles for the spectator view angles instead of the default.

oh i didnt even know that the first person spec view angles has been that inaccurate this whole time

imo that seems like the way to go

mastercoms:
Here is the "hack" just in case

```cpp
// Take the owning player eye position and direction.
vecAttachment = pPlayer->EyePosition();
C_TFPlayer* pLocalPlayer = C_TFPlayer::GetLocalTFPlayer();
if ( pLocalPlayer && pLocalPlayer->GetObserverTarget() == pPlayer )
{
	// match up to the view that the local player is observing.
	QAngle anglesEye = pPlayer->EyeAngles();
	AngleVectors( anglesEye, &vecDir );
}
else
{
	vecDir = GetAbsOrigin() - vecAttachment;
	VectorNormalize( vecDir );
}
```

wgetJane:
implemented that "hack" since it should be completely fine even if the more complicated fix is implemented afterwards

PR #645, opened 2025-02-19
Fix sdk_container script causing the linux build to fail

zJanny:
The podman run command had an invalid syntax for the mount options which caused the script to fail, and the .ccache directory might not have been created which also caused the command to fail. This pull request fixes this

### DIFF
--- a/src/sdk_container
+++ b/src/sdk_container
@@ -24,13 +24,14 @@ function run_in_sniper() {
 
         image="registry.gitlab.steamos.cloud/steamrt/sniper/sdk:latest"
         echo "Running script inside $image"
+        mkdir ~/.ccache #create directory because it might not exists which causes the mount to fail
         exec podman run \
             --env "$check_env=$script_file" \
             --env "CCACHE_DIR=${CCACHE_DIR:-$HOME/.ccache}" \
             --userns=keep-id \
             --rm -it \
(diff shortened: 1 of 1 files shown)


src/sdk_container
```diff
@@ -24,13 +24,14 @@ function run_in_sniper() {
 
         image="registry.gitlab.steamos.cloud/steamrt/sniper/sdk:latest"
         echo "Running script inside $image"
+        mkdir ~/.ccache #create directory because it might not exists which causes the mount to fail
```
ekohl: I'd use `-p` so it doesn't fail when it already exists
```suggestion
        mkdir -p ~/.ccache #create directory because it might not exists which causes the mount to fail
```

But really, you want it to match what's actually mounted so:
```suggestion
        mkdir -p "${CCACHE_DIR:-$HOME/.ccache}"
```

KingKrouch:
Yeah I noticed the .ccache folder issue when trying to compile the code.
I also noticed, at least on NixOS, it's impossible to run the bash scripts until you modify the headers of "sdk_container" and "buildallprojects" to have this towards the top instead of ``#!/bin/bash``:

``#!/usr/bin/env bash``

xVemu:
Why closed?

zJanny:
> Why closed?

This has already been merged here:
https://github.com/BitlDevelopmentStudios/source-sdk-2013-bds-base/commit/339918c50f832c9c37da115eb673d9ee90cc6448

PR #654, opened 2025-02-19
[TF2] fix: cheat-protect commands which control panel visibility

mastercoms:
some VGUI exploits to gain information which you normally cannot see (explained outside of commit)

ValveSoftware/Source-1-Games#2488

should also be done for vgui_togglepanel: https://github.com/mastercomfig/tf2-patches-old/commit/a02595b7aa1bb909e0734a37358863c145c99470

### DIFF
--- a/src/game/client/game_controls/baseviewport.cpp
+++ b/src/game/client/game_controls/baseviewport.cpp
@@ -79,7 +79,7 @@ void hud_autoreloadscript_callback( IConVar *var, const char *pOldValue, float f
 
 static ConVar cl_leveloverviewmarker( "cl_leveloverviewmarker", "0", FCVAR_CHEAT );
 
-CON_COMMAND( showpanel, "Shows a viewport panel <name>" )
+CON_COMMAND_F( showpanel, "Shows a viewport panel <name>", FCVAR_CHEAT )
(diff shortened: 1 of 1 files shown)


ldesgoui:
I assume you tested this in TC2, but this isn't gonna break VGUI right? Like it's not using this codepath normally?

Niterux:
I don't agree with this because the panels that are shown and hidden themselves should be fixed IMO, if any other way to show and hide panels gets discovered, those panels will still give an advantage, unless they are patched. This tech is also used in huds to add custom menus.
[VillageGreenPreserver/hudmenu-base](https://github.com/VillageGreenPreserver/hudmenu-base)

ldesgoui:
Yeah that's fair. If a dev comes around and fixes `specgui`, please do get in contact with me or ETF2L AntiCheat staff, there is one more case that's exploitable and I'd rather not make it public for the time being. Thanks!

mastercoms:
> I assume you tested this in TC2, but this isn't gonna break VGUI right? Like it's not using this codepath normally?

Yes, some custom HUDs do rely on this functionality to create custom menus, but the main game had no issues with this patch.

PR #655, opened 2025-02-19
add very naive logic for launching in steam runtime

Jan200101:
the logic added here is very naive and just copies the SDK 2013 mp lookup. it then runs `$RUNTIME_DIR/_v2-entry-point -- hl2.sh` to ensure everything is setup right

`_v2-entry-point` is labeled as a prototype so this might break in the future

`$RUNTIME_DIR/run -- $RUNTIME_DIR/pressure-vessel/bin/steam-runtime-launcher-interface-0 container-runtime ./hl2.sh` might be better? unsure

there are better ways of handling this in the code, this is mostly just to get things working without changing everything

### DIFF
--- a/src/launcher_main/main.cpp
+++ b/src/launcher_main/main.cpp
@@ -87,6 +87,7 @@ static void *Launcher_GetProcAddress( void *pHandle, const char *pszName )
 #endif
 
 static const AppId_t k_unSDK2013MPAppId = 243750;
+static const AppId_t k_unSniperRuntimeAppID = 1628350;
 
 #ifdef MOD_LAUNCHER
 static const AppId_t k_unMyModAppid = MOD_APPID;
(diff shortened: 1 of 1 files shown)


Jan200101:
its worth noting that there is a comment in `hl2.sh` that notes the dependency on the sniper runtime however there is no check to enforce this

```
# determine platform
UNAME=$(uname)
if [ "$UNAME" == "Linux" ]; then
    # TF2 requires the sniper container runtime
    . /etc/os-release

    # prepend our lib path to LD_LIBRARY_PATH
    export LD_LIBRARY_PATH="${GAMEROOT}/bin/$plat_dir":$LD_LIBRARY_PATH
fi
```

wgetJane:
imo why not just a bash file in the game output folder?

Jan200101:
the runtime can be installed anywhere so making use of the steam api to get it, like how its done for the actual engine build, is a lot more reliable

PR #661, opened 2025-02-19
[TF2] Add ConVar for changing player max speed limit

CookieCat45:
Replaces the hardcoded max speed limit of 520 on players with a new cvar: `tf_maxspeed_limit`


darkitxu84:
Unrelated: I love how the comment says an extra 3% when it's really 30%. How could that happen?

0x5066:
floating point probably...?

FlaminSarge:
Nice.

FlaminSarge:
Closes https://github.com/ValveSoftware/Source-1-Games/issues/4572

PR #669, opened 2025-02-19
[TF2] Update 10 second stuck MvM bots to use CommitSuicide over TakeDamage for rare spawn softlocks

Mentrillum:
This is very rare to happen nowadays. But what would happen is an MvM bot would get stuck in world geometry while trying to leave spawn. If the bot stays stuck for 10 seconds, the game prints a dev message saying `A robot is interpenetrating a solid - killed!` and the game applies 999999.9 damage, the issue here is that since the bot is inside of the spawn that damage gets nullified because the bot still has the uber spawn condition. CommitSuicide does not do this and instantly kills the bot even with spawn uber protection. As such, this change is here to prevent those rare moments.

For reference on this bug, see https://github.com/ValveSoftware/Source-1-Games/issues/4620


henke37:
This addresses a known trolling method where players intentionally causes bots to get stuck.

misyltoad:

Mentrillum:
Yes I have, for this PR, #675, and any future PRs I do here I accept the terms.

PR #682, opened 2025-02-20
[TF2] Fix prediction for force-a-nature jumps

wgetJane:
fixes this issue partially: https://github.com/ValveSoftware/source-1-games/issues/6017

unfixed, net_fakelag 100:

https://github.com/user-attachments/assets/10c4b937-72a5-4c46-b198-43318ac1546a

fixed, net_fakelag 100:

https://github.com/user-attachments/assets/7f37d57e-bdf0-4411-b3a2-6cd2004947eb

CAVEAT: this does not fix prediction for the "stun" that gets self-applied by FaN-jumping, which reduces your airstrafing control

i feel like it's a separate issue to fix, though i would personally just remove the self-stun since it seems pretty odd to have in the first place, but that can be easily interpreted as a balance change

this still makes FaN jumps feel a LOT smoother, but the screen jitter would still happen if you try to airstrafe immediately upon a FaN jump at high ping

btw i also tested this in other people's random frog fortress servers, so this works completely clientside as well


FlaminSarge:
Heads up, I was planning on opening a PR for https://github.com/mastercomfig/tf2-patches-old/pull/462

Should I pull in these changes into that as well for the sake of Valve being able to pick up both changes easily without messy rebase/merge?

wgetJane:
sounds good, though i'll keep this pr open until yours is sure to be merged

maybe if possible, take my commit from this pr first for your pr so it's less potentially messy

src/game/shared/tf/tf_player_shared.cpp
```diff
@@ -472,6 +472,7 @@ BEGIN_PREDICTION_DATA_NO_BASE( CTFPlayerShared )
 	DEFINE_PRED_FIELD( m_bIsTargetedForPasstimePass, FIELD_BOOLEAN, FTYPEDESC_INSENDTABLE ), // does this belong here?
 	DEFINE_PRED_FIELD( m_askForBallTime, FIELD_FLOAT, FTYPEDESC_INSENDTABLE ),
 	DEFINE_PRED_ARRAY( m_flItemChargeMeter, FIELD_FLOAT, LAST_LOADOUT_SLOT_WITH_CHARGE_METER, FTYPEDESC_INSENDTABLE ),
+	DEFINE_FIELD( m_bScattergunJump, FIELD_BOOLEAN ),
```
FlaminSarge: @wgetJane I'm unclear on why this (and m_flDisguiseCompleteTime) don't use the predicted field macros despite being in the predicted prop block. Also seems like the prop is defined differently from other predicted props (in terms of where in tf_player_shared.h it lives/etc.). Do you have an explanation of that (or an explanation of how the prop macros work in general, I have no idea what I'm doing)?

Would also be useful to describe why the matrix calcs have been swapped out in the PR desc, just for posterity.

wgetJane: afaik this is how you declare non-networked predicted variables (basically just smoothing without error correction from the server), and to me it didn't seem like it needed to be networked since it resets when you land anyway, though it would work fine as a networked var as well

the matrix stuff was swapped out because EntityMatrix only exists on the server for some reason, but the calculations have the exact same results (im sure, since i tested this as a client-only fix)

misyltoad: I think you want `DEFINE_PRED_FIELD` without `FTYPEDESC_INSENDTABLE` for non-networked predicted variables, otherwise you get `FTYPEDESC_KEY | FTYPEDESC_SAVE` in the flags.

Those should be harmless, but probably good to be consistent?

FlaminSarge: > the matrix stuff was swapped out because EntityMatrix only exists on the server for some reason, but the calculations have the exact same results (im sure, since i tested this as a client-only fix)

Makes sense, just figured it might be useful for community projects if you included that note in the PR description, since it's not immediately evident why it was changed from the changes themselves.

FlaminSarge: > I think you want `DEFINE_PRED_FIELD` without `FTYPEDESC_INSENDTABLE` for non-networked predicted variables, otherwise you get `FTYPEDESC_KEY | FTYPEDESC_SAVE` in the flags.

Would `FTYPEDESC_NOERRORCHECK` be appropriate here? Though that seems CLIENT_DLL-only.

wgetJane: i assumed that that flag only makes sense with FTYPEDESC_INSENDTABLE, i'll check if it even does anything without it

FlaminSarge: I see it getting used without INSENDTABLE but only inside CLIENT_DLL blocks

src/game/shared/tf/tf_weapon_shotgun.cpp
```diff
@@ -329,27 +328,29 @@ void CTFScatterGun::FireBullet( CTFPlayer *pPlayer )
...
 
 		// Knock the firer back!
-		if ( !(pOwner->GetFlags() & FL_ONGROUND) && !pPlayer->m_bScattergunJump )
+		if ( !(pOwner->GetFlags() & FL_ONGROUND) && !pPlayer->m_Shared.m_bScattergunJump )
 		{
-			pPlayer->m_bScattergunJump = true;
+			pPlayer->m_Shared.m_bScattergunJump = true;
 
+#ifndef CLIENT_DLL
 			pOwner->m_Shared.StunPlayer( 0.3f, 1.f, TF_STUN_MOVEMENT | TF_STUN_MOVEMENT_FORWARD_ONLY );
```
misyltoad, line 335: Any reason we can't predict this?

src/game/shared/tf/tf_weapon_shotgun.cpp
```diff
@@ -329,27 +328,29 @@ void CTFScatterGun::FireBullet( CTFPlayer *pPlayer )
...
+#endif
 
 			float flForce = AirBurstDamageForce( pOwner->WorldAlignSize(), 60, 6.f );
 
 			Vector vecForward;
 			AngleVectors( pOwner->EyeAngles(), &vecForward );
 			Vector vecForce = vecForward * -flForce;
 
-			EntityMatrix mtxPlayer;
```
misyltoad, line 344: Maybe we should just bring EntityMatrix over for client as well?

wgetJane: sounds fair, i was trying to minimise how many files i touch for this fix

something to note about EntityMatrix::InitFromEntity is that it uses GetAbsAngles but the client should use EyeAngles otherwise there'd be pred errors when you FaN jump while moving your mouse

src/game/shared/tf/tf_weapon_shotgun.cpp
wgetJane: there isnt, i just imagined it could be a separate issue since a few other things (like thermal thruster) apply self-stun

i'll see if i can implement predicted self-stun for this pr as well

misyltoad:
Had to merge manually, but ty

FlaminSarge:
(For bookkeeping: https://github.com/ValveSoftware/source-sdk-2013/commit/a62efecf624923d3bacc67b8ee4b7f8a9855abfd )

PR #690, opened 2025-02-20
Fix the caber, and fix intended extended melee hits

Reabstraction:
In TF2, the game expects us to be charging whilst attacking to get a range bonus from 48 (Bottle)/72 (Sword) to 128
Unfortunately, this is bugged
You see, melee hits aren't instant
But your charge ends the instant you try to swing
As such, we can never use the extended range, since we aren't "charging" when the hit connects
This can be worked around by swinging as soon as we charge, but this means giving up a crit and requiring us to be very close

This pull request fixes this, by introducing a class variable called `m_bWasCharging`, which is set if you swing whilst charging, and unset when the hit is done or you swap weapons

I also fixed the caber being excluded from a few achievements

### DIFF
--- a/src/game/shared/tf/achievements_tf_demoman.cpp
+++ b/src/game/shared/tf/achievements_tf_demoman.cpp
@@ -626,7 +626,8 @@ class CAchievementTFDemoman_MeleeKillWhileJumping : public CBaseTFAchievement
 		{
 			if ( m_bStickyJumping )
 			{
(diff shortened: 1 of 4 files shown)


Reabstraction:
LMK if a CLA needs to be signed, or if there are any issues with the pull request

wgetJane:
imo i would interpret this as also a balance change instead of as only a bug fix

wgetJane:
@Reabstraction this feels like it should be 3 different prs:
- caber achievements fix
- caber mini-crits (seems like a balance change?)
- extended charge melee range (bug fix but also seems like a balance change)

PR #692, opened 2025-02-20
[HL2] Fix Linux compilation

Blixibon:
This pull request fixes HL2 and Episodic projects not compiling on Linux in the `master` branch after the latest commit.

### DIFF
--- a/src/game/server/hl2/npc_playercompanion.cpp
+++ b/src/game/server/hl2/npc_playercompanion.cpp
@@ -3538,7 +3538,7 @@ void CNPC_PlayerCompanion::InputClearAllOuputs( inputdata_t &inputdata )
 			typedescription_t *dataDesc = &dmap->dataDesc[i];
 			if ( ( dataDesc->fieldType == FIELD_CUSTOM ) && ( dataDesc->flags & FTYPEDESC_OUTPUT ) )
 			{
-				CBaseEntityOutput *pOutput = (CBaseEntityOutput *)((int)this + (int)dataDesc->fieldOffset[0]);
+				CBaseEntityOutput *pOutput = (CBaseEntityOutput *)((intp)this + (int)dataDesc->fieldOffset[0]);
 				pOutput->DeleteAllElements();
 				/*
 				int nConnections = pOutput->NumberOfElements();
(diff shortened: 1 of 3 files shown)


SrcEngGuy:
Isn't `NULL` by default defined? If it isn't why not just add these lines of code to some include?
```cpp
#ifdef _LINUX || _OSX
#define NULL 0
#endif
````

**Comment by Blixibon** [comment]
> Isn't `NULL` by default defined? If it isn't why not just add these lines of code to some include?
> 
> ```c++
> #ifdef _LINUX || _OSX
> #define NULL 0
> #endif
> ```

The issue was that `NULL` is ambiguous in these cases, which causes errors in the updated SDK.

For example:
```
game/shared/hl2/hl2_gamerules.cpp: In function 'CAmmoDef* GetAmmoDef()':
game/shared/hl2/hl2_gamerules.cpp:1817:84: error: call of overloaded 'AddAmmoType(const char [8], int, AmmoTracer_t, NULL, NULL, NULL, int, int)' is ambiguous
 1817 |   def.AddAmmoType("Battery",   DMG_CLUB,     TRACER_NONE,   NULL, NULL, NULL, 0, 0 );
      |                                                                                    ^
In file included from game/shared/hl2/hl2_gamerules.cpp:10:
game/shared/ammodef.h:86:10: note: candidate: 'void CAmmoDef::AddAmmoType(const char*, int, int, int, int, int, float, int, int, int)'
   86 |  void    AddAmmoType(char const* name, int damageType, int tracerType, int plr_dmg, int npc_dmg, int carry, float physicsForceImpulse, int nFlags, int minSplashSize = 4, int maxSplashSize = 8 );
      |          ^~~~~~~~~~~
In file included from game/shared/hl2/hl2_gamerules.cpp:10:
game/shared/ammodef.h:87:10: note: candidate: 'void CAmmoDef::AddAmmoType(const char*, int, int, const char*, const char*, const char*, float, int, int, int)'
   87 |  void    AddAmmoType(char const* name, int damageType, int tracerType, char const* plr_cvar, char const* npc_var, char const* carry_cvar, float physicsForceImpulse, int nFlags, int minSplashSize = 4, int maxSplashSize = 8 );
      |          ^~~~~~~~~~~
```

PR #694, opened 2025-02-20
Changed float to int for MaxClients()

Northalicious:
Changed float to int for MaxClients() as part of vscript as it makes more sense for this function


ficool2:
This can cause unexpected issues in user code that expects a float from this function. Breaking ABI for script functions is not a good idea

PR #696, opened 2025-02-20
Added MaxClientsEx() to fix MaxClients()

Northalicious:
MaxClients() mistakenly returns a float, so I made a new function for compatibility that returns an int.
(Apologies for the re-submit, I'm still fairly new to github's specifics)

### DIFF
--- a/src/game/server/vscript_server.cpp
+++ b/src/game/server/vscript_server.cpp
@@ -1006,7 +1006,7 @@ static float Time()
 	return gpGlobals->curtime;
 }
 
-static float MaxClients()
+static int MaxClients()
 {
 	return gpGlobals->maxClients;
 }
--- a/src/game/server/vscript_server.cpp
+++ b/src/game/server/vscript_server.cpp
@@ -1006,7 +1006,12 @@ static float Time()
 	return gpGlobals->curtime;
 }
 
-static int MaxClients()
+static float MaxClients()
+{
+	return gpGlobals->maxClients;
+}
+
+static int MaxClientsEx()
 {
 	return gpGlobals->maxClients;
 }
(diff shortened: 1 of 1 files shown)


ficool2:
This is useless and it's not worth bloating the functions list for this (especially when the float return on MaxClients is harmless most of the time, and if not you just call .tointeger)

darkitxu84:
Is there any practical use in using this function instead of the original one? Besides, shouldn't it be an unsigned integer?

Northalicious:
> Is there any practical use in using this function instead of the original one? Besides, shouldn't it be an unsigned integer?

You make a good point. Changed to an unsigned int. The only reason to not use the original is that if applying the number to other numbers you'd have to convert it using .ToInt()

treacherousfiend:
I don't think this is particularly necessary, if you really are worried about this being an integer you can just set up a function in vscript, a variable, or even a constant if you do a minor workaround to get the value as an integer. I just do `getconsttable().MAX_CLIENTS <- MaxClients().tointeger();` so I have a constant to use instead of calling the function constantly, even though it is cheap.

> Besides, shouldn't it be an unsigned integer?

Bit of a side note, while in theory this number shouldn't be negative so unsigned should be fine, as far as I can tell the only language valve has used for vscript that supports unsigned integers is Lua. Also vscript struggles in some scenarios with handling uint for the languages that don't support it. All of the non-functional/bugged exposed functions in TF2's CTFBot class are broken because of it improperly handling uint, so I'd avoid uint wherever possible when coding in regards to vscript.

For reference: 
Functions in vscript call for int (as uint is unsupported): https://developer.valvesoftware.com/wiki/Team_Fortress_2/Scripting/Script_Functions#CTFBot
C++ code expects uint or enum that I presume the compiler turns into uint since it has no negative values: https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/game/server/tf/bot/tf_bot.h#L340-L365

PR #698, opened 2025-02-20
[TF2] Add NetProps preventing Hale from earning Heavy achievements

Lizard-Of-Oz:
Custom characters, such as Saxton Hale in VSH, technically are heavily modified existing classes.

This leads to Hale earning Heavy-specific achievements, which introduces the feeling of jankiness.

To solve this, in this PR I propose two NetProps for the Player that can temporarily disable achievements and contracker quests for an individual player. The idea is to then change those NetProps on demand via VScript for the custom character.

- When those NetProps are set to 0 (the default value) they allow achievements/quests as normal.
- When set to 1, they disable class-specific achievements/quests. There's all-class achievements such as "Accumulate 1000 total kills" and it makes sense that Hale could earn them too.
- When set to 2, earning achievements/quests is disabled for that player entirely.

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -3756,6 +3756,8 @@ IMPLEMENT_CLIENTCLASS_DT( C_TFPlayer, DT_TFPlayer, CTFPlayer )
 	RecvPropInt( RECVINFO( m_iPlayerSkinOverride ) ),
 	RecvPropBool( RECVINFO( m_bViewingCYOAPDA ) ),
 	RecvPropBool( RECVINFO( m_bRegenerating ) ),
+	RecvPropInt( RECVINFO( m_nRestrictAchievements ) ),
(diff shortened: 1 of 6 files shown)


darkitxu84:
Maybe it's a little unnecessary, but if the variable only goes from 0 to 2 it would make more sense to use a data type uint8_t (1 byte) instead of short (2 bytes)

Lizard-Of-Oz:
Initially I wanted to add a NetProp to disable player stat collection as well, to prevent players from having 3 headshots as Scout, but the stat tracking code is too intimidating. If anyone can figure out how to do this, it'd be appreciated.

Lizard-Of-Oz:
Upon further thinking, I've got a better idea I'll PR in the following days: make EVENT HOOKS for achievements, stats and contracker quests, so that VScript can cancel them. After all, this PR was intended to be used in conjunction with VScript, and the new idea seems like a straight upgrade.

PR #699, opened 2025-02-20
added more names to tf_bot.cpp

Circuitella:
I Added A Few Names, Some Being Internal References To The Engine, Some Voice Line References, Others Meet The Team Video References And 1 For No Reason, Having More Names Could Make Offline Play Feel More Alive Which Would Be Cool

### DIFF
--- a/src/game/server/tf/bot/tf_bot.cpp
+++ b/src/game/server/tf/bot/tf_bot.cpp
@@ -195,6 +195,22 @@ const char *GetRandomBotName( void )
 		"Saxton Hale",
 		"RageQuit",
 		"Screamin' Eagles",
+		"No Bones",
+		"Im Not Real",
+	
+		"Sexy Mann",
+		"This Sucks On Ice",
+		"SEDUCE ME!",
+		"Bloody Hell",
+		"Explode",
+		"Kill",
+		"sv_cheats 1",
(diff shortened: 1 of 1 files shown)


src/game/server/tf/bot/tf_bot.cpp
```diff
@@ -218,7 +234,7 @@ const char *GetRandomBotName( void )
 		"Tiny Baby Man",
 		"Big Mean Muther Hubbard",
 		"Force of Nature",
-
+
```
Margen67: whitespace
```suggestion

```

src/game/server/tf/bot/tf_bot.cpp
```diff
@@ -195,6 +195,22 @@ const char *GetRandomBotName( void )
 		"Saxton Hale",
 		"RageQuit",
 		"Screamin' Eagles",
+		"No Bones",
+		"Im Not Real",
+
```
WilsonNet:
```suggestion
```

FlaminSarge:
In keeping with the original file, the new names should most likely be added at the end of the list, with one extra newline separating them from the other names. I believe that's why the extra whitespace exists for previous sections of the list: each time new names were added they separated them with a newline.

PR #703, opened 2025-02-21
[TF2] Prevent disguise kit resetting multiple times in high ping situations

nosoop:
Fixes ValveSoftware/Source-1-Games#2755.

`CTFWeaponPDA_Spy::Deploy` may be called multiple times across multiple ticks if the player latency is high enough.
Feels like there should be a better fix for this, but this seems non-invasive enough.

Might require a bit of code review.  Some light testing suggests that this works, but I haven't really dug into the VGUI side of things to understand potential side-effects of setting flags in `CHudMenuSpyDisguise::SetVisible`.


nosoop:
@wgetJane Yeah; setting that on deploy is way simpler.  Thanks!

FlaminSarge:
This doesn't happen to need `InPrediction()` like the minigun fix needed, right?

wgetJane:
> This doesn't happen to need `InPrediction()` like the minigun fix needed, right?

the event "spy_pda_reset" seems to be only used for the hud, so probably not

PR #706, opened 2025-02-21
[TF2] Add FCVAR_SERVER_CANNOT_QUERY to cl_disablehtmlmotd

megascatterbomb:
Updates cl_disablehtmlmotd by adding the `FCVAR_SERVER_CANNOT_QUERY` flag to prevent community servers from "nagging" for it to be enabled.

Originally I proposed disabling the cvar outright by adding dev only, however the discussion below made it clear that several other useful features rely on html motds. This updated proposal will hopefully serve as an acceptable compromise.

<details>
<summary> Original Proposal </summary>
Also requires removing/commenting out this option in `user_default.scr` (outside this repo):

```
// To be removed from user_default.scr L601-L607
	"cl_disablehtmlmotd"
	{
		"#Valve_DisableHTMLMOTD"
		"#Tooltip_DisableHTMLMOTD"
		{ BOOL }
		{ "0" }
	}
```

This PR proposes changing `cl_disablehtmlmotd` to a dev-only cvar forced to 1, preventing forced advertising within community servers.

In practice, HTML Motds are slow to load, and are used for two purposes:

- Communicating server info like the rules (which can be done with a non-html motd).
- Advertising to players.

Even when users have set `cl_disablehtmlmotd 1`, they are often nagged to revert this cvar. For this reason, I believe just changing the cvar's default to 1 is insufficient; the dev-only flag is required to dissuade this practice entirely.

Example of advertising with HTML Motd:

image

Examples of nags:

image

Screenshot 2025-02-20 230751
</details>

### DIFF
--- a/src/game/client/game_controls/vguitextwindow.cpp
+++ b/src/game/client/game_controls/vguitextwindow.cpp
@@ -32,7 +32,7 @@ extern INetworkStringTable *g_pStringTableInfoPanel;
 
 #define TEMP_HTML_FILE	"textwindow_temp.html"
 
-ConVar cl_disablehtmlmotd( "cl_disablehtmlmotd", "0", FCVAR_ARCHIVE, "Disable HTML motds." );
+ConVar cl_disablehtmlmotd( "cl_disablehtmlmotd", "0", FCVAR_ARCHIVE | FCVAR_SERVER_CANNOT_QUERY, "Disable HTML motds." );
 
 //=============================================================================
 // HPE_BEGIN:


Dessensce:
no more ads, valve

ficool2:
A better solution is to just add `FCVAR_SERVER_CANNOT_QUERY` to `cl_disablehtmlmotd` so servers cannot easily query whether the client disabled it

KaelaSavia:
That would also break
- !radio feature people use to listen to music
- !backpack command trade servers use
- easy way to join community server's steam group as !group
- custom inventories like slag gaming [no idea if they still exist]
and probably some other stuff i don't remember

FCVAR_SERVER_CANNOT_QUERY would be better choice as above said

KaelaSavia:
> Thanks for the feedback, I've updated the cvar to use FCVAR_SERVER_CANNOT_QUERY instead.
> 
> I've kept the default value for the cvar at 1. Hopefully this is an acceptable compromise for countering ads without completely breaking the other features @KaelaSavia mentioned.


I'd consider that part anti-feature personally.

With cvar marked as FCVAR_SERVER_CANNOT_QUERY and cl_disablehtmlmotd set to 1 there's no way to tell player why commands using motd like !inventory, !music, !steamgroup etc. do not work.

Only way to let players know why would be nagging them in chat or in radio menu every time they press button to open page with text 'Enable Motd' even though he might have motd already enabled because you can't tell with FCVAR_SERVER_CANNOT_QUERY set.

Or having admins or other players derail chat every time someone asks why 'X' doesn't work because motd was enabled by default for decades in all source games

KaelaSavia:
Would be great if they could also take care of servers ping spoofing at same time. Joining server with 30 ping only for it to turn out 300 is not fun!

PR #718, opened 2025-02-21
Add colored console chat history

ficool2:
Self-explanatory, makes chat history in the console use colors.
screenshot

This PR also fixes the bug where saying a full chat message containing only spaces breaks color codes afterwards (as it made the console become solid white). One more small issue this PR addresses is removing an unsafe const cast and instead copying arguments from `say`/`say_team` command into a buffer first, since `CheckChatText` writes to it in-place,

NOTE: this PR requires the spew bug in materialsystem to be fixed as well, otherwise the console chat can bug out and be positioned out of order.

### DIFF
--- a/src/game/client/hud_basechat.cpp
+++ b/src/game/client/hud_basechat.cpp
@@ -792,7 +792,8 @@ void CBaseHudChat::MsgFunc_SayText( bf_read &msg )
 	CLocalPlayerFilter filter;
 	C_BaseEntity::EmitSound( filter, SOUND_FROM_LOCAL_PLAYER, "HudChat.Message" );
 
-	Msg( "%s", szString );
+	// TERROR: color console echo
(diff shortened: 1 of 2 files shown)


Ashetf2:
ficool could you fix words not being highlighted properly in the console? (or maybe it's fixed here)

ficool2:
> ficool could you fix words not being highlighted properly in the console? (or maybe it's fixed here)

https://github.com/ValveSoftware/source-sdk-2013/pull/761

KaelaSavia:
Can we have convar to disable colored console for accessibility reasons?
 -Thanks

Niterux:
This has a lot of issues because of https://github.com/ValveSoftware/Source-1-Games/issues/1985. Before this is merged that issue should probably be fixed, lots of times names and colors will end up on separate lines in the console and such.

PR #727, opened 2025-02-21
Fix radial fog not being enabled in 3D sky on official maps

ficool2:
Self-explanatory, [matches the fog controller](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/game/server/fogcontroller.cpp#L375)


ficool2:
Yes, I have read the contributiom terms and agree to them.

PR #731, opened 2025-02-21
[TF2] Add client predicted autobhop cvars.

kiwitf2:
Adds two cvars based on whats seen with cs:s and csgo (for parity plus it would be VERY nice to have).
Autobunnyhopping also works whilst crouched

"sv_enablebunnyhopping" ( def. "0" )
 game notify replicated
 - Removes speed cap from bunnyhopping.

"sv_autobunnyhopping" ( def. "0")
 game notify replicated
Whether players will automatically bunnyhop whilst holding +JUMP.
   0   - Players will not automatically jump while holding +JUMP.
   1   - Players will automatically jump while holding +JUMP.
   2   - Players will automatically jump while holding +JUMP, and will be able to jump while crouching.

### DIFF
--- a/src/game/shared/tf/tf_gamemovement.cpp
+++ b/src/game/shared/tf/tf_gamemovement.cpp
@@ -57,6 +57,9 @@ ConVar  tf_parachute_maxspeed_onfire_z( "tf_parachute_maxspeed_onfire_z", "10.0f
 ConVar  tf_parachute_aircontrol( "tf_parachute_aircontrol", "2.5f", FCVAR_DEVELOPMENTONLY | FCVAR_REPLICATED, "Multiplier for how much air control players have when Parachute is deployed" );
 ConVar	tf_parachute_deploy_toggle_allowed( "tf_parachute_deploy_toggle_allowed", "0", FCVAR_REPLICATED | FCVAR_DEVELOPMENTONLY );
 
+ConVar 	sv_enablebunnyhopping( "sv_enablebunnyhopping", "0", FCVAR_REPLICATED | FCVAR_NOTIFY, "Removes speed cap from bunnyhopping." );
+ConVar 	sv_autobunnyhopping( "sv_autobunnyhopping", "0", FCVAR_REPLICATED | FCVAR_NOTIFY, "Players will automatically jump while holding space." );
+
 ConVar  tf_halloween_kart_aircontrol( "tf_halloween_kart_aircontrol", "1.2f", FCVAR_CHEAT | FCVAR_REPLICATED, "Multiplier for how much air control players have when in Kart Mode" );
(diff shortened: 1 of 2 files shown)


src/game/shared/tf/tf_gamemovement.cpp
```diff
@@ -1237,7 +1243,7 @@ bool CTFGameMovement::CheckJumpButton()
 		return false;
 
 	// Cannot jump again until the jump button has been released.
-	if ( mv->m_nOldButtons & IN_JUMP )
+	if ( mv->m_nOldButtons & IN_JUMP && (!sv_autobunnyhopping.GetBool() || !bOnGround ) )
```
Bitl: Having !bOnGround in this statement means that this code will break parachutes/ghosts/scout dashes. After this statement, there's an if statement that checks if scouts can air dash, which is completely skipped when this code is executed. Removing || !bOnGround fixes the issue.

kiwitf2: tested ghost and parachutes and they seem to work fine whether i use your change or not

Removing `|| !bOnGround` (your "fix") causes the scout to dash immediately after the bhop. 

Having `|| !bOnGround` was intentional as it lets you hold spcae to bhop and then release and tap it if you wanted to dash

wgetJane:
have you tested how this behaves with scout's doublejump?

src/game/shared/tf/tf_gamemovement.cpp
```diff
@@ -1223,7 +1229,7 @@ bool CTFGameMovement::CheckJumpButton()
 	ToggleParachute();
 
 	// Cannot jump will ducked.
-	if ( player->GetFlags() & FL_DUCKING )
+	if ( player->GetFlags() & FL_DUCKING && !sv_autobunnyhopping.GetBool())
```
qawab: I could be wrong but I believe this if-statement conflicts with the game's design choice. Normally, players are blocked from jumping while crouched.

kiwitf2: considering that being able to crouch while bhopping is really nice, i might make sv_autobunnyhopping allow for crouch bhopping if it's set to 2, ty for feedback

src/game/shared/tf/tf_gamemovement.cpp
FlaminSarge, line 1104: Would it not be better to instead have a cvar to change the max speed factor?

kiwitf2: this is based on how it is in counter strike and im fairly sure there its just enable/disable

FlaminSarge: Is it? I still think it's far more flexible to be able to set the factor by cvar instead.

kiwitf2: ¯\_(ツ)_/¯ i just made this with that in mind, i made this pr a while ago and was and am happy enough with how it is rn tbh, don't really wanna change it all that much

PR #735, opened 2025-02-22
[TF2] Fix player not being able to deploy parachute after landing and becoming airborne without jump button

FlaminSarge:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/2455

As described in above bug, TF_COND_PARACHUTE_DEPLOYED is not removed upon landing, but only upon using the jump button while grounded. This can lead to a case where the player has used the parachute, landed, and then becomes airborne without the jump button again, resulting in them being unable to deploy the parachute despite being fully grounded.

This patch moves the removal of the condition to the same case as TF_COND_PARACHUTE_ACTIVE's removal when grounded. It also prevents using the jump button to add the parachute conditions while grounded, in order to prevent the client from receiving the parachute condition for a frame every time they jump.

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -3102,14 +3102,8 @@ void CTFPlayerShared::ConditionThink( void )
 		RemoveCond( TF_COND_KNOCKED_INTO_AIR );
 		RemoveCond( TF_COND_AIR_CURRENT );
 
-		if ( InCond( TF_COND_PARACHUTE_ACTIVE ) )
-		{
-			RemoveCond( TF_COND_PARACHUTE_ACTIVE );
-		}
-		if ( InCond( TF_COND_PARACHUTE_DEPLOYED ) )
-		{
-			RemoveCond( TF_COND_PARACHUTE_DEPLOYED );
-		}
+		RemoveCond( TF_COND_PARACHUTE_ACTIVE );
+		RemoveCond( TF_COND_PARACHUTE_DEPLOYED );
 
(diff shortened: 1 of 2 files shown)


FlaminSarge:
Alternative to "Don't allow activating parachute on ground": prevent ToggleParachute entirely when on the ground. However, this has the opportunity to introduce complications with parachute cond removal for halloween karts/etc., so limiting the behavior to preventing just activation is safer.

src/game/shared/tf/tf_player_shared.cpp
```diff
@@ -3090,9 +3090,16 @@ void CTFPlayerShared::ConditionThink( void )
 
 	VehicleThink();
 
-	if ( m_pOuter->GetFlags() & FL_ONGROUND && InCond( TF_COND_PARACHUTE_ACTIVE ) )
+	if ( m_pOuter->GetFlags() & FL_ONGROUND )
 	{
-		RemoveCond( TF_COND_PARACHUTE_ACTIVE );
+		if ( InCond( TF_COND_PARACHUTE_ACTIVE ) )
```
misyltoad: Are these InConds' necessary?

FlaminSarge: ... They are not (unless there's some optimization I'm unaware of, or if Assert is particularly taxing):
```
void CTFPlayerShared::RemoveCond( ETFCond eCond, bool ignore_duration )
{
	Assert( eCond >= 0 && eCond < TF_COND_LAST );
	Assert( eCond < m_ConditionData.Count() );

	if ( !InCond( eCond ) )
		return;
```

Let me know if it turns out they're needed re: optimization, but I'll remove them for now.

PR #736, opened 2025-02-22
Restore functionality to tf_parachute_maxspeed_onfire_z but set its value to match tf_parachute_maxspeed_z

FlaminSarge:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/3960

This allows servers to restore the original functionality with a value of '10.0f' (or whatever other value they desire).

Of note that high positive values for this cvar will cause the case to be skipped on next evaluation (due to z-velocity < 0 check), but this implementation exactly matches that of 2014 pre-Tough-Break.


A1exKH (approved):
LGTM.

PR #751, opened 2025-02-23
[TF2] Remove cheat/dev flags from viewmodel sway convars

NachosChipeados:
Closes https://github.com/ValveSoftware/Source-1-Games/issues/2120

### DIFF
--- a/src/game/shared/tf/tf_viewmodel.cpp
+++ b/src/game/shared/tf/tf_viewmodel.cpp
@@ -62,8 +62,8 @@ CTFViewModel::~CTFViewModel()
 void DrawEconEntityAttachedModels( CBaseAnimating *pEnt, CEconEntity *pAttachedModelSource, const ClientModelRenderInfo_t *pInfo, int iMatchDisplayFlags );
 
 // TODO:  Turning this off by setting interp 0.0 instead of 0.1 for now since we have a timing bug to resolve
-ConVar cl_wpn_sway_interp( "cl_wpn_sway_interp", "0.0", FCVAR_CLIENTDLL | FCVAR_CHEAT | FCVAR_DEVELOPMENTONLY );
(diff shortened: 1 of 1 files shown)


TorutheRedFox:
you should also add FCVAR_ARCHIVE so it persists, and something similar to the following should also be added to user_default.scr:

```
"cl_wpn_sway_interp"
{
	"#TF_wpn_sway_option"
	"#Tooltip_wpn_sway_option"
	{ SLIDER 0 0.1 }
	{ "0" }
}
```

NachosChipeados:
Note that theres currently some odd behavior when using flipped viewmodels with the sway convars. I'll see how i can fix that

NachosChipeados:
The flipped viewmodels bug has been fixed, so i'd say this is safe to merge now 👍

wgetJane:
have you tested this with net_fakelag and net_fakejitter?

NachosChipeados:
I don't remember, and atm i can't really test it again because my computer is broken

0x5066:
Seeing this PR, is there any technical reason why the viewmodel sway is so different compared to HL2 and TF2/HL2DM?
Does it have anything to do with networking or something of that sort?

TorutheRedFox:
the sway seems to have been reprogrammed for TF2, and it got disabled as it was causing issues around launch that don't seem to occur anymore

teufortressIndustries:
Virtually every single Source 1 game has weapon sway, I don't see a reason to not add this.

GrubPL:
> The flipped viewmodels bug has been fixed, so i'd say this is safe to merge now 👍 

Is the flipped viewmodel sway fix TF2 specific?
Because I'm pretty sure CS:S also has this issue

TorutheRedFox:
> Is the flipped viewmodel sway fix TF2 specific?
> Because I'm pretty sure CS:S also has this issue

it is tf2 specific as it mainly targets two tf2 viewmodel cvar that were marked as dev only and one of them to 0 to nullify the effects of the sway

PR #759, opened 2025-02-23
bsplib: Ensure WriteBSPFile resets g_pBSPHeader as it points to a local function variable

### LINKED ISSUE #758: [cleanup] Global g_pBSPHeader pointer points to destroyed stack variable after WriteBSPFile completition
Global ```g_pBSPHeader``` pointer points to local stack variable ```outHeader``` in the beginning of ```WriteBSPFile```:
https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/utils/common/bsplib.cpp#L2635-L2637

But it is not reset to ```nullptr``` when ```WriteBSPFile``` completed. ```g_pBSPHeader``` used extensively in other APIs, so it is very easy to dereference a stall pointer.

Suggest to reset ```g_pBSPHeader``` to ```nullptr``` immediately before ```outHeader``` is destroyed, so ```g_pBSPHeader``` will not point to stall data.

dimhotepus:
Closes #758


dimhotepus:

PR #761, opened 2025-02-23
Fix unreliable console and chat mouse selection

ficool2:
The console and chat have a notorious reputation for being difficult to select on single-lines. This is due to RichText splitting the buffer when calculating line breaks, which means it doesn't account for extra splitting that happens in the line painting loop, and this makes the selection sometimes skip `RichText::DrawString`. This PR addresses the issue by making the selection buffer be split while painting instead of when calculating line breaks, and now selection works perfectly.

Before:

https://github.com/user-attachments/assets/60b8c93d-ea97-4974-8eb0-8bec08aeed7b

After:

https://github.com/user-attachments/assets/09c9dae0-2e8c-45c6-ada0-0d10f3beebc5

NOTE to anyone merging this PR: this requires GameUI to be recompiled, or overriding the console dialog yourself.

### DIFF
--- a/src/vgui2/vgui_controls/RichText.cpp
+++ b/src/vgui2/vgui_controls/RichText.cpp
@@ -862,6 +862,12 @@ void RichText::Paint()
 		if ( m_LineBreaks.IsValidIndex( lineBreakIndexIndex ) && m_LineBreaks[lineBreakIndexIndex] < iLim )
 			iLim = m_LineBreaks[lineBreakIndexIndex];
 
+		// Stop in selection range
+		if ( iLim >= selection0  && i < selection0 )
+			iLim = selection0;
+		if ( iLim >= selection1 && i < selection1 && i >= selection0 )
+			iLim = selection1;
+
 		// Handle non-drawing characters specially
 		for ( int iT = i; iT < iLim; iT++ )
 		{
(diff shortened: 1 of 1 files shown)


AdamTadeusz:
@ficool2 How would one go about recompiling GameUI?

Ashetf2:
Tested it. Highlighting text in console doesn't work but it does in the chat

https://github.com/user-attachments/assets/6e3111cb-9054-4f8c-a87f-7e010c3d651c

https://github.com/user-attachments/assets/07953df3-826f-4ed1-bbd2-b012865842ea

ficool2:
> Tested it. Highlighting text in console doesn't work but it does in the chat
> Desktop.2025.05.18.-.13.30.53.04.DVR.mp4
> Desktop.2025.05.18.-.13.33.18.06.DVR.mp4

`NOTE to anyone merging this PR into their own mod: this requires GameUI to be recompiled, or overriding the console dialog yourself..`

Ashetf2:
I did recompile gameui wdym

FrozenDragon0:
This would be great to see not just for TF2 but all source games aswell, as most other source games such as HL2, Portal, Portal 2 etc suffer this same problem.

PR #767, opened 2025-02-23
Fix match hud breaking if initially disabled by user

ficool2:
If the user has `tf_use_match_hud` set to 0, the match HUD status at the top has a broken layout because it doesn't set itself up. This is because the code only reloads if the toggle status of the match HUD has changed, but it's initial value is false. Users typically have to run `hud_reloadscheme` to force a layout refresh. This PR addresses that by giving the state variable a value of -1 initially so that the layout is always reset the first time around.

Screenshot of how the layout gets broken currently:
image

### DIFF
--- a/src/game/client/tf/tf_hud_match_status.cpp
+++ b/src/game/client/tf/tf_hud_match_status.cpp
@@ -269,7 +269,7 @@ CTFHudMatchStatus::CTFHudMatchStatus(const char *pElementName)
 	: CHudElement(pElementName)
 	, BaseClass(NULL, "HudMatchStatus")
 	, m_pTimePanel( NULL )
-	, m_bUseMatchHUD( false )
+	, m_nUseMatchHUD( -1 )
 	, m_eMatchGroupSettings( k_eTFMatchGroup_Invalid )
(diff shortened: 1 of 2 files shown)


TorutheRedFox:
an alternate fix would be to make bReload a member variable that's initialized to true and set to false when finished reloading

TorutheRedFox:
simply adding ``m_bUseMatchHUD = ShouldUseMatchHUD();`` to the end of CTFHudMatchStatus's constructor seems to work too

Ashe-tf:
Valve pls add ☝

mastercoms:
I think this is breaking scheme initialization on some parts of the HUD, like the KOTH timer. Perhaps it would be best to change this to an int, and set it to -1 for uninitialized state.

TorutheRedFox:
> I think this is breaking scheme initialization on some parts of the HUD, like the KOTH timer. Perhaps it would be best to change this to an int, and set it to -1 for uninitialized state.

technically you could also just set the bool to -1 as a bool is just an integer (with a typical size of 8 bits) like any other

PR #769, opened 2025-02-23
[TF2] Add support for red miniboss background for support and mission

HyperionCoding:
Self-explanatory, see picture

kuva

I decided to keep support and mission giants separated because missions shows up only when they are active. Mission icons are hidden when the wave starts and removing icons from between other icons would look odd.

This does not break Spy icon flashing.


misyltoad:
Hello, we fixed this a much simpler way in upstream TF2. So I am going to close this for now.

HyperionCoding:
> Hello, we fixed this a much simpler way in upstream TF2. So I am going to close this for now.

The upstream solution has multiple problems and only seems to add background check without doing the reordering. Also mission support does not get miniboss background.

PR #770, opened 2025-02-23
[TF2] Add ConVar to enable/disable MvM upgrade sorting by price

HyperionCoding:
Currently custom upgrade files that change the price of upgrades will cause upgrades to show up in different places and throw off players who often buy upgrades by muscle memory. This PR addresses that by adding ConVar that allows disabling the sorting. If disabled, the upgrades will appear in the order defined in the upgrades file.

Default behaviour with custom upgrade file:

kuva

Sorting disabled and upgrades re-ordered in the upgrade file:

kuva

The convar flags may need to be changed.


A1exKH (approved):
LGTM.

PR #772, opened 2025-02-24
[TF2] Allow set_scattergun_has_knockback attribute to take on values greater than 1

FlaminSarge:
This increases the number of times self-knockback may be applied via the attribute.
This also changes m_bScattergunJump to m_iScattergunJump and updates any relevant code accordingly.

Changes CTFPlayerShared's bool m_bScattergunJump to int m_iScattergunJump.
Renames CTFScattergun's HasKnockback to GetKnockbackCount to return the attribute value directly.
Changes all usages of HasKnockback to `GetKnockbackCount() > 0`
Changes CTFScatterGun's FireBullet knockback logic to compare the value of m_iScattergunJump to GetKnockbackCount in order to determine whether to apply self-knockback.
Changes CanScatterGunKnockBack in tf_weapon_shotgun to be consistent with elsewhere on `> 0` check, fixing an issue where values of -1 could cause the regular-damage-zero-z-force behavior but not actually apply the extra knockback.

~~This version of the PR includes #682 for easy rebasing/inclusion in projects that already include that PR.~~
Rebased onto master now that the above PR is included in master.

_Should_ be demo-safe.

Implements https://github.com/mastercomfig/tf2-patches-old/pull/462


FlaminSarge:
@wgetJane let me know if I missed anything/messed up on this

wgetJane:
looks good

PR #773, opened 2025-02-24
[TF2] Allow set_scattergun_has_knockback attribute to take on values greater than 1 - no prediction fix

FlaminSarge:
This is an implementation of https://github.com/ValveSoftware/source-sdk-2013/pull/772 that does not include the prediction fixes for Scattergun jumps from #682, for projects that do not include that commit to use (it is highly recommended to use that fix, though).

Instead of the relevant changes in CTFPlayerShared, the modified scattergun jump property is in CTFPlayer.


FlaminSarge:
#682 was merged and #772 has been rebased, this PR is redundant.

PR #776, opened 2025-02-24
[TF2] Allow player_sapped_object event to fire for world-built sentries, setting ownerid to -1

FlaminSarge:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/4587

Adds a check against valid ownerid for the one case that's missing it.


A1exKH (approved):
LGTM.

PR #779, opened 2025-02-24
all: Fix dangerous macro expressions as macroses are tricky with evaluation order

### LINKED ISSUE #778: [bug] Dangerous macro expressions waiting to explode
When you define function-like macro, you should:

0. Consider using ```{constexpr}``` function first, which does points below obsolete :)
1. Ensure side effects of N-times evaluation are expected:
```
#define abs(x) x >= 0 ? x : -x

int v = abs(readFromFile(f));  // Evaluates as readFromFile(f) >= 0 ? readFromFile(f) : -readFromFile(f), which is unlikely you expect.
```
2. Wrap entire macro body in parentheses, consider: 
```
#define twice(x) x + x

int a = twice(12) * 3;  // Evaluates as 12 + 12 * 3 = 48, not as (12 + 12) * 3 = 72 you might expect.
```
3. Wrap every macro parameter in parentheses, consider:
```
#define linear(k, x, b) (k * x + b)

int v = linear(3 + 1, 2, 5);  // Evaluates as 3 + 1 * 2 + 5 = 10, not as (3 + 1) * 2 + 5 = 13 you might expect. 
```

Linked PR fixes some most suspicious places.

dimhotepus:
Closes #778

### DIFF
--- a/src/public/filesystem.h
+++ b/src/public/filesystem.h
@@ -143,8 +143,8 @@ enum
 };
 typedef uint32 PathTypeQuery_t;
 
-#define IS_PACKFILE( n ) ( n & ( PATH_IS_PACKFILE | PATH_IS_MAPPACKFILE ) )
-#define IS_REMOTE( n )   ( n & PATH_IS_REMOTE )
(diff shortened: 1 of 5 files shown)


dimhotepus:

PR #797, opened 2025-02-25
[TF2] Fix prediction for jumping when huntsman is charged/released

wgetJane:

(in the vids below, i press jump immediately after charging 3 times, then jump immediately after releasing 3 times, then jump immediately after holstering 3 times)

unfixed, net_fakelag 100:

https://github.com/user-attachments/assets/fca19788-6847-4e5e-b1ad-0991ddb84d9d

fixed, net_fakelag 100:

https://github.com/user-attachments/assets/f88c1153-b433-41fd-8479-9d69af8aba47

this pr adds CTFWeaponBase::OwnerCanJump (similar to the existing CTFWeaponBase::OwnerCanTaunt)

### DIFF
--- a/src/game/shared/tf/tf_weapon_compound_bow.cpp
+++ b/src/game/shared/tf/tf_weapon_compound_bow.cpp
@@ -654,26 +654,6 @@ void CTFCompoundBow::SetArrowAlight( bool bAlight )
 //-----------------------------------------------------------------------------
 // Purpose: 
 //-----------------------------------------------------------------------------
-void CTFCompoundBow::SetInternalChargeBeginTime( float flChargeBeginTime )
-{
-#ifdef GAME_DLL
-//	float flCurrentChargeBeginTime = GetInternalChargeBeginTime();
-//	if ( flCurrentChargeBeginTime == 0.f && flChargeBeginTime > 0.f )
-//	{
-//		DisableJump();
-//	}
-//	else if ( flCurrentChargeBeginTime > 0.f && flChargeBeginTime == 0.f )
-//	{
-//		EnableJump();
-//	}
-#endif // GAME_DLL
-
-	BaseClass::SetInternalChargeBeginTime( flChargeBeginTime );
-}
-
-//-----------------------------------------------------------------------------
-// Purpose: 
-//-----------------------------------------------------------------------------
 bool CTFCompoundBow::OwnerCanJump( void )
 {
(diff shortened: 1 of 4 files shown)


misyltoad:

src/game/shared/tf/tf_weapon_compound_bow.cpp
```diff
@@ -657,16 +657,24 @@ void CTFCompoundBow::SetArrowAlight( bool bAlight )
...
-	float flCurrentChargeBeginTime = GetInternalChargeBeginTime();
-	if ( flCurrentChargeBeginTime == 0.f && flChargeBeginTime > 0.f )
-	{
-		DisableJump();
-	}
-	else if ( flCurrentChargeBeginTime > 0.f && flChargeBeginTime == 0.f )
-	{
-		EnableJump();
-	}
+//	float flCurrentChargeBeginTime = GetInternalChargeBeginTime();
```
misyltoad: Let's just remove this function override, now.

wgetJane:
> Hello, have you seen and read the contributing terms here?
> 
> https://github.com/ValveSoftware/source-sdk-2013/blob/master/CONTRIBUTING
> 
> These are visible when you make a PR now, so I am asking for the older PRs. Thanks.

yeah looks good and i agree with these terms

mastercoms:
This also fixes a bug where picking up a charged Huntsman off the ground makes you unable to jump.

wgetJane:
> This also fixes a bug where picking up a charged Huntsman off the ground makes you unable to jump.

oh nice so https://github.com/ValveSoftware/Source-1-Games/issues/4013 is fixed as well

mastercoms:
Wait sorry, no, still needs a weapon reset call. But it does help a lot since the disable / enable jump isn't stateful.

wgetJane:
tested before and after this pr, not sure if this is the correct way to reproduce that bug:
- spawn as sniper with huntsman
- charge huntsman
- killbind while huntsman is charged
- respawn
- pick up dropped huntsman

before this pr: you can't jump until you die, or drop the bow, or change your primary, or charge the bow again
after this pr: you can jump as soon as it is picked up

mastercoms:
Ah interesting, maybe it is fixed, but to me charge time seems like it'd persist and not sure why it'd be updating like that. Maybe because it's not stateful, there's something idle or something that updates charge time and we don't need a full reset to reset state?

PR #798, opened 2025-02-25
Fix for MvM Administrator lines being truncated by countdown

rabscootle:
This fix is intended to resolve MvM announcer lines being truncated by the final 5 seconds of setup countdown. This was achieved by adding code to ensure that the announcer wave start line has finished playing before playing countdown audio.

Video of fixed bug:
https://github.com/user-attachments/assets/6ead54a2-ebd9-4e98-a4f9-ec398e3f8947


wgetJane:
btw to embed videos, make sure to have a newline directly both above and below the link

PR #799, opened 2025-02-25
Set m_mousespeed default value to 0 (disable mouse acceleration entirely)

Tiagoquix:
This makes sures no other ConVar related to mouse acceleration is enabled by default. Related ConVars are disabled by default, with the exception of `m_mousespeed` that is addressed in this PR.

This is an addition to `m_rawinput` being enabled by default.

---------

For reference, a popular FPS config. for TF2, called "mastercomfig", has an add-on called Flat Mouse, which disables all acceleration-related mouse ConVars.

https://github.com/mastercomfig/mastercomfig/blob/7eb031ad8cd8e096b623819909e18a64c88ec16a/config/cfg/addons/flat-mouse.cfg

It also used to enable `m_rawinput` when its default was `0`.


A1exKH (approved):
LGTM.

PR #801, opened 2025-02-25
Fix bug where credits dropped in the robots' spawn become unobtainable in MVM

bradleycha:
In rare circumstances, a bot killed just outside of its spawn may drop credits within the robot spawn.  This can lead to the credits becoming unobtainable, sometimes even with Scout's money magnet mechanic.

This fix checks if the credits come to rest within the robots' spawn, and if they do, automatically collects them.  I also did some minor refactoring of CCurrencyPack::ComeToRest(), as I feel that the function is a bit of a mess.

### DIFF
--- a/src/game/server/tf/entity_currencypack.cpp
+++ b/src/game/server/tf/entity_currencypack.cpp
@@ -18,6 +18,7 @@
 #include "player_vs_environment/tf_population_manager.h"
 #include "collisionutils.h"
 #include "tf_objective_resource.h"
+#include "func_respawnroom.h"
 
 //=============================================================================
 //
@@ -171,6 +172,28 @@ void CCurrencyPack::BlinkThink( void )
 	SetContextThink( &CCurrencyPack::BlinkThink, gpGlobals->curtime + TF_CURRENCYPACK_BLINK_DURATION, "CurrencyPackBlinkThink" );
 }
 
+//-----------------------------------------------------------------------------
+// Despawn and collect the currency
+//-----------------------------------------------------------------------------
+void CCurrencyPack::DespawnAndCollect( void )
+{
+   TFGameRules()->DistributeCurrencyAmount( m_nAmount );
+	m_bTouched = true;
+	UTIL_Remove( this );
+}
+
+//-----------------------------------------------------------------------------
+// Check if the currency's origin is within some bounding box
+//-----------------------------------------------------------------------------
+bool CCurrencyPack::IsWithinCollidableBoundingBox( CBaseEntity *pEntity )
+{
+   Vector origin = GetCollideable()->GetCollisionOrigin();
+
+   Vector vecMins, vecMaxs;
+	pEntity->GetCollideable()->WorldSpaceSurroundingBounds( &vecMins, &vecMaxs );
+
+   return IsPointInBox( origin, vecMins, vecMaxs );
+}
 
 //-----------------------------------------------------------------------------
 // Become touchable when we are at rest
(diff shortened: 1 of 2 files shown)


catuhana:
#618 ?

TheShermanTanker:
Since we're doing this, can we remove robot spawn invulnerability? The comment on that states that the robots only get free uber in robot spawn to prevent money from being dropped in the spawn where players cannot pick it up. With this change, that functionality is now obsolete. Look at tf\_bot\_behavior.cpp and tf\_populators.cpp for more details

bradleycha:
> #618 ?

I wasn't aware the that PR when I made this code, thanks for letting me know.

While that PR does collect money in spawn, this code includes a few additional checks to ensure only active spawns for the robots will trigger the fail-safe.  I do this out of fear for weird edge-cases where bots could die just outside of red's spawn, which would trigger the fail-safe and collect money, even though it's perfectly reachable.

TheShermanTanker:
> > Since we're doing this, can we remove robot spawn invulnerability? The comment on that states that the robots only get free uber in robot spawn to prevent money from being dropped in the spawn where players cannot pick it up. With this change, that functionality is now obsolete. Look at tf_bot_behavior.cpp and tf_populators.cpp for more details
> 
> This is a bad idea. The bot spawns are clearly not intended to be combat zones. Most classes would not even be able to deal damage to bots in that area.
> 
> Your suggestion would also open up the possibility of the bomb being dropped in inaccessible areas e.g. on Mannworks, if the bomb is dropped on one side and the next wave of bots spawn on the other side, then it's impossible to lose as those bots cannot get back into their spawn once they drop down.

In that case I retract my suggestion. That aside, I think Event\_Killed has logic for trigger\_hurt money collection that can be modified to accommodate this fix, it will likely be simpler to implement it there:

Currently it looks like this:
```cpp
// Give money directly to the team if a trigger killed us
if ( info.GetDamageType() )
{
	CBaseTrigger *pTrigger = dynamic_cast< CBaseTrigger *>( info.GetInflictor() );
	if ( pTrigger )
	{
		bDropPack = false;	
		TFGameRules()->DistributeCurrencyAmount( nDropAmount, NULL, true, true );
	}
}
```

You might be able to change it to the following:

```cpp
// Give money directly to the team if a trigger killed us or if we were destroyed in spawn
if ( GetLastKnownArea() != nullptr && GetLastKnownArea()->HasAttributeTF( TF_NAV_SPAWN_ROOM_BLUE ) )
{
	bDropPack = false;
	TFGameRules()->DistributeCurrencyAmount( nDropAmount, nullptr, true, true );
}
else if ( info.GetDamageType() )
{
	CBaseTrigger *pTrigger = dynamic_cast< CBaseTrigger *>( info.GetInflictor() );
	if ( pTrigger )
	{
		bDropPack = false;	
		TFGameRules()->DistributeCurrencyAmount( nDropAmount, nullptr, true, true );
	}
}
```

This would circumvent the need to implement the complex check in this Pull Request and is in my opinion simpler to do

TheShermanTanker:
> > Since we're doing this, can we remove robot spawn invulnerability? The comment on that states that the robots only get free uber in robot spawn to prevent money from being dropped in the spawn where players cannot pick it up. With this change, that functionality is now obsolete. Look at tf_bot_behavior.cpp and tf_populators.cpp for more details
> 
> This is a bad idea. The bot spawns are clearly not intended to be combat zones. Most classes would not even be able to deal damage to bots in that area.
> 
> Your suggestion would also open up the possibility of the bomb being dropped in inaccessible areas e.g. on Mannworks, if the bomb is dropped on one side and the next wave of bots spawn on the other side, then it's impossible to lose as those bots cannot get back into their spawn once they drop down.

In that case I retract my suggestion.

PR #802, opened 2025-02-25
[TF2] Fix prediction errors and bugs related to Revenge Crits

### LINKED ISSUE #6059: [TF2] Manmelter faulty crit charge consumption
By holding down alt-fire first, then primary fire, the Manmelter consumes it's critical charges without launching it's fireball. This happens once per frame/tick, until it's altfire is released. Presumably the crits are instead being consumed on the alt-fire? There's no error information in the console; a player can be flamed by a flamethrower and extinguished by the Manmelter and nothing new will appear in the console.

FlaminSarge:
* Fixes Manmelter revenge crits [depleting when not able to fire](https://youtu.be/EJ3lHsF7Q6Q)
* Fixes revenge crit counter rubberbanding due to prediction (for revolver, shotgun, flaregun)
* Only decrements revenge crits if the projectile successfully fires (for all 3 weapon types)
* ~~Only plays Absorb effects on flaregun when actually extinguishing players (even if no revenge crits gained due to self-burn check); this removes the clientside "did revenge crits increase" check that used to play it erroneously when rubberbanding occurred~~ undid this change as m_iRevengeCrits is correctly predicted now
* Fixes Manmelter beeping rapidly if attempting to fire underwater
* Replace CTFPlayerShared IncrementRevengeCrits with AddRevengeCrits to handle both inc/dec cases
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/6059
Fixes issue 2 on https://github.com/ValveSoftware/Source-1-Games/issues/4022


FlaminSarge:
@wgetJane I need some help on the m_iRevengeCrits prediction behavior for this; even with these changes, the client seems to still be predicting the count dropping by more than 1 (often by 4-5) then rubberbanding back up to the correct value. I was under the impression that moving the decrement to FireProjectile would be enough to fix that, but it doesn't seem like that's the case.

wgetJane:
@FlaminSarge havent checked yet but i would imagine that thats caused by m_iRevengeCrits not being a predicted variable, so it gets decremented several times due to smoothing (predicted methods get called multiple times for smoothing purposes)

so either check for IsFirstTimePredicted() before decrementing or declare it as a predicted field (with INSENDTABLE since its probably a networked var)

FlaminSarge:
> @FlaminSarge havent checked yet but i would imagine that thats caused by m_iRevengeCrits not being a predicted variable, so it gets decremented several times due to smoothing (predicted methods get called multiple times for smoothing purposes)
> 
> so either check for IsFirstTimePredicted() before decrementing or declare it as a predicted field (with INSENDTABLE since its probably a networked var)

Seems like the latter did the job, thanks for the help.

FlaminSarge:
Updated to undo the DoAbsorbEffects-related changes now that the revengecrits prop is predicted. Tested everything, should be good to go. I accept CONTRIBUTING.md/etc.

PR #803, opened 2025-02-25
[TF2] Fix prediction for hud charge meters of stickybomb launcher, loose cannon, huntsman, etc

wgetJane:

affected items:
\- Stickybomb Launcher
\- Loose Cannon
\- Huntsman / Fortified Compound
\- Dragon's Fury (hud meter is hidden by default without a custom hud)
\- Sapper (mvm)

(btw in the vids below, red crosshair = mouse1 is held down)

unfixed, net_fakelag 200:

https://github.com/user-attachments/assets/ae9c8363-91ec-4183-be45-ed65e348c252

fixed, net_fakelag 200:

https://github.com/user-attachments/assets/650e68ed-79b3-4eb3-b103-b1908811aebb

(note: the fix from https://github.com/ValveSoftware/source-sdk-2013/pull/777 is also active in the second vid)

imo this solution doesnt feel too elegant, so tell me if u can see a better way to fix this


JustYourOrdinaryNormie:
you could also just swap out curtime with TICKS_TO_TIME(pPlayer->m_nTickBase)

PR #804, opened 2025-02-25
Team Fortress 2: Visibly show other sappers

OfficerSpy:
When a sapper is built on a player, only the default sapper is visible. Let the other sappers visibly appear as well.

This is a bug fix for a minor problem that is currently only present in Mann vs. Machine Mode.


ficool2:
A better fix rather than horribly hardcoding model names like this would be to override the `ValidateEntityAttachedToPlayer` method on C_ObjectSapper to always validate (see `C_PointWorldText::ValidateEntityAttachedToPlayer`)

PR #809, opened 2025-02-25
studio: Clamp root LOD in range 0...MAX_NUM_LODS - 1 as Clamp is pure function

### LINKED ISSUE #808: [bug] Clamp doesn't clamp root LOD in range 0...MAX_NUM_LODS - 1 which can cause issues with LOD
```Clamp``` is pure function which means its only side effect is the result:
https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/public/tier0/basetypes.h#L97-L106

It means ```Clamp``` here doesn't do anything meaningful:
https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/public/studio.h#L3188-L3189

Need to store ```Clamp``` result in the ```rootLOD``` variable.

dimhotepus:
Closes #808

### DIFF
--- a/src/public/studio.h
+++ b/src/public/studio.h
@@ -3186,7 +3186,7 @@ inline void Studio_SetRootLOD( studiohdr_t *pStudioHdr, int rootLOD )
 	}
 
 	Assert( rootLOD >= 0 && rootLOD < MAX_NUM_LODS );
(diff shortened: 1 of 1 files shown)


dimhotepus:

PR #811, opened 2025-02-25
sha1: Ensure compiler will not remove memset for clearing security-sensitive data

### LINKED ISSUE #810: [bug] SHA1 has a cleanup code to clear security sensitive data but compiler removes it due to optimization in Release mode
```finalcount``` is local buffer used for security-sensitive data, so it is cleared before function returns:

https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/tier1/checksum_sha1.cpp#L219
https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/tier1/checksum_sha1.cpp#L238-L245

Problem is compiler removes memset during optimizations as he can prove local buffer is not accessed after. So memset is dead code and eliminated.

See this SO question on details and links:
https://stackoverflow.com/questions/15538366/can-memset-function-call-be-removed-by-compiler

See standart on how to ensure such ```memset``` is not removed:
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1381.pdf

dimhotepus:
Closes #810

### DIFF
--- a/src/tier1/checksum_sha1.cpp
+++ b/src/tier1/checksum_sha1.cpp
@@ -209,6 +209,14 @@ bool CSHA1::HashFile(char *szFileName)
 }
 #endif
 
+// Compiler removes memset when it proves no side effects are present.
+// See https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1381.pdf
+static void* secure_memset( void* v, int c, size_t n ) {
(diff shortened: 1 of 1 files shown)


dimhotepus:

PR #818, opened 2025-02-25
all: Ensure pointers casted from one type to another have same underlying alignment

### LINKED ISSUE #817: [bug] Ensure pointer casts between different types will work correctly as such types must have same alignment
Objects of different types may be aligned differently, and casting a pointer type may break the alignment. Dereferencing such an incorrect pointer may cause a crash, while performing operations on it may cause data loss.

See https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf (page 56, statement 7):
> A pointer to an object type may be converted to a pointer to a different object type. If the resulting pointer **is not correctly aligned** for the referenced type, **the behavior is undefined.**

Consider the following example:

```
void foo(void) {
  char ch = '1';
  int *int_ptr = (int *)&ch;
  char *char_ptr = (char *)int_ptr;
}
```

Reading the address of the 'ch' variable and writing it to a pointer of type 'int' could result in data loss. 
When casting back, the alignment may change.

Especially useful when your target architecture is not x86-64, like M1+, Android, Windows ARM, etc as they have different alignment requirements.

See 
https://blog.quarkslab.com/unaligned-accesses-in-cc-what-why-and-solutions-to-do-it-properly.html
https://www.reddit.com/r/C_Programming/comments/14yva4f/ub_caused_by_pointer_casting/
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf

dimhotepus:
Closes #817

### DIFF
--- a/src/public/tier0/memoverride.cpp
+++ b/src/public/tier0/memoverride.cpp
@@ -594,7 +594,7 @@ ALLOC_CALL void *__cdecl _aligned_malloc_base( size_t size, size_t align )
 
 inline void *MemAlloc_Unalign( void *pMemBlock )
 {
-	unsigned *pAlloc = (unsigned *)pMemBlock;
+	alignas(unsigned **) unsigned *pAlloc = (unsigned *)pMemBlock;
 
 	// pAlloc points to the pointer to starting of the memory block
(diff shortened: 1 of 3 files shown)


dimhotepus:

PR #824, opened 2025-02-25
Add typing bubble above players

ficool2:
There is an unused particle in TF2 named `speech_typing`, and it appears to have been forgotten about given no code for it ever existed. This PR restores it so it appears when the player has their chat box open, Would make for a nice quality-of-life improvement just like the existing voice chat bubbles above player heads. 

https://github.com/user-attachments/assets/cf4d0fac-f541-4a75-be8e-83ed9a5d7eca

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -4259,6 +4259,8 @@ void C_TFPlayer::SetDormant( bool bDormant )
 		{
 			ShowBirthdayEffect( false );
 		}
+
+		ToggleTypingEffect( false );
 	}
 
 	if ( IsDormant() && !bDormant )
@@ -8138,6 +8140,21 @@ void C_TFPlayer::UpdateTypingEffect()
 		&& !InFirstPersonView()
 		&& ( !m_Shared.IsStealthed() || !IsEnemyPlayer() ) )
 	{
+		ToggleTypingEffect( true );
+	}
+	else
+	{
+		ToggleTypingEffect( false );
+	}
+}
+
+//-----------------------------------------------------------------------------
+// Purpose: 
+//-----------------------------------------------------------------------------
+void C_TFPlayer::ToggleTypingEffect( bool bToggle )
+{
+	if ( bToggle )
+	{
 		if ( !m_pTypingEffect )
 		{
 			m_pTypingEffect = ParticleProp()->Create( "speech_typing", PATTACH_POINT_FOLLOW, "head" );
@@ -8152,6 +8169,7 @@ void C_TFPlayer::UpdateTypingEffect()
 		}
 	}
 }
+
 
 //-----------------------------------------------------------------------------
 // Purpose: 
--- a/src/game/client/tf/c_tf_player.h
+++ b/src/game/client/tf/c_tf_player.h
@@ -231,6 +231,7 @@ public:
 	void StopSaveMeEffect( bool bForceRemoveInstantly = false );
 
 	void UpdateTypingEffect();
+	void ToggleTypingEffect( bool bToggle );
 
 	void CreateTauntWithMeEffect();
 	void StopTauntWithMeEffect();


src/game/shared/in_buttons.h
```diff
@@ -37,5 +37,6 @@
 #define IN_GRENADE1		(1 << 23)	// grenade 1
 #define IN_GRENADE2		(1 << 24)	// grenade 2
 #define	IN_ATTACK3		(1 << 25)
+#define	IN_TYPING		(1 << 26)
```
Tripperful, line 40: IN_ is used for actual inputs, adding this here is inappropriate.
You mentioned the voice indicator in the description yourself, so maybe it's worth implementing this textual one in a similar manner to voice?

ficool2: Initially I implemented a new field in `CUserCmd` but I opted against it, since the `buttons` field conveniently has several unused bits and all 32-bits get networked. The alternative is to send string commands to the server, which is not only more code, I also can't vouch for their reliability unlike player commands

KaelaSavia:
I'd recommend adding client-side convar that lets users disable it, TF2 is laggy already as it is, last thing we need is another particle above each player.

Other than that, are there going to be two speech bubbles if player is typing and using voice chat at same time?

AzureWoof:
It would also be nice to change it so that cloaked Spies aren't revealed to enemies with both this particle and the voice chat particle when alltalk is on.

ficool2:
> I'd recommend adding client-side convar that lets users disable it, TF2 is laggy already as it is, last thing we need is another particle above each player.
> 
> Other than that, are there going to be two speech bubbles if player is typing and using voice chat at same time?

Added toggle convar and made the chat particle hidden when using voice in last commit

> It would also be nice to change it so that cloaked Spies aren't revealed to enemies with both this particle and the voice chat particle when alltalk is on, as long as that's in-scope for this request.

I already made sure to account for this, I'll make a separate PR for the voice chat particle side of it

Void48:
Could this please be team-only? The VC icon is team-only (I get why), and I feel like it’d be better not to signal to the enemy team that you’re typing, since it just makes it easier for them to kill you. Without the icon, no one knew if someone was at a disadvantage for a second. Realistically this probably doesn’t matter much, but I still think it would be better if it were only visible to teammates. Sorry for the rambling... :fishsticks: 

"You can still be killed if you're standing still while typing" - yes, I'm aware. However I type fast, so currently it doesn't really affect me or other people who type quickly. And the typing bubble appears in advance before someone even starts typing

> Summary: Make the typing indicator team-only when "**alltalk**" is disabled.

src/game/shared/in_buttons.h
megascatterbomb: I'm concerned that if the input array has a constant 1 from someone leaving their chatbox open, that could be a trivial bypass for anti-afk mechanisms. If the anti-afk checks in the game check only for changes in value then I'd be okay with this, but in any case I agree with Tripperful that it's not ideal.

On the other hand implementing something new might break demos which is definitely worse.

Mooshua: > IN_ is used for actual inputs, adding this here is inappropriate
> @Tripperful

The base game already allocates `IN_SCORE` for the scoreboard, how is this inappropriate?

> I'm concerned that if the input array has a constant 1 from someone leaving their chatbox open, that could be a trivial bypass for anti-afk mechanisms.
>@megascatterbomb

[The game doesn't care about anything other than the movement buttons, thankfully.](https://github.com/ValveSoftware/source-sdk-2013/blob/238a15a79dd2c3c151e98eea8b39e5f161a93703/src/game/server/tf/tf_player.cpp#L2348).

Mooshua:
Should the concommand shenanigans for activating/toggling this shiny new input be added to [in_main](https://github.com/ValveSoftware/source-sdk-2013/blob/238a15a79dd2c3c151e98eea8b39e5f161a93703/src/game/client/in_main.cpp#L1453)? Competitive players might have fun tricking people with that if so

src/game/shared/in_buttons.h
Tripperful: > The base game already allocates `IN_SCORE` for the scoreboard, how is this inappropriate?

+showscores is an actual input bound to the Tab key by default which you hold down. It's different for what you're proposing for the chat. Adding such random stuff to the IN bitmask is bad because there are only just so many bits to play with, it must only deal with actual inputs

ficool2: > there are only just so many bits to play with

There has only been [one extra bit added since 2007](https://github.com/Source-SDK-Archives/source-sdk-orangebox/blob/master/game/shared/in_buttons.h), this is not getting exhausted anytime soon

Bitl:
I re-implemented this PR in my fork as I felt the need to allocate an input bit to this functionality was unnecessary, as it isn't a key.

https://github.com/BitlDevelopmentStudios/source-sdk-2013-bds-base/commit/c25f464ea26c532393a9a7038ffb83ff855ecb94

changes done:

- m_bTyping is implemented in the base player classes, rather than tf_player. It is also a public networked variable used for the below context.
- m_bTyping is turned on or off depending on the situation in ClientModeShared::CreateMove

All other aspects are virtually the same, but it doesn't use an input bit for it. It should function the same way.

ficool2:
> I re-implemented this PR in my fork as I felt the need to allocate an input bit to this functionality was unnecessary, as it isn't a key.
> 
> 
> 
> https://github.com/BitlDevelopmentStudios/source-sdk-2013-bds-base/commit/c25f464ea26c532393a9a7038ffb83ff855ecb94
> 
> 
> 
> changes done:
> 
> 
> 
> - m_bTyping is implemented in the base player classes, rather than tf_player. It is also a public networked variable used for the below context.
> 
> - m_bTyping is turned on or off depending on the situation in ClientModeShared::CreateMove
> 
> 
> 
> All other aspects are virtually the same, but it doesn't use an input bit for it. It should function the same way.

Your implementation doesn't work in multiplayer, the client is not networking the typing state to the server (networkvars are one-way from server -> client)

Flpstrike:
Awesome!

GrubPL:
I'd like to note that the speech bubble doesn't show up while taunting, I doubt that's intended behavior for this PR 
feel free to correct me on that one btw, just thought I'd point it out incase it is unintended

PR #828, opened 2025-02-25
Hide voice labels for cloaked enemy spies

ficool2:
When using voice chat, a sprite will appear above the player's head. Unfortunately, this does even when cloaked as a spy which obviously reveals your position. This PR changes that so the sprite does not appear for enemy spies that are cloaked.

### DIFF
--- a/src/game/shared/voice_status.cpp
+++ b/src/game/shared/voice_status.cpp
@@ -227,11 +227,10 @@ void CVoiceStatus::DrawHeadLabels()
 			continue;
 
 		C_BasePlayer *pPlayer = dynamic_cast<C_BasePlayer*>(pClient);
-		if( !pPlayer )
+		if ( !pPlayer )
 			continue;
 
-		// Don't show an icon for dead or spectating players (ie: invisible entities).
-		if( pPlayer->IsPlayerDead() )
+		if ( !pPlayer->ShouldShowHeadLabel() )
 			continue;
 
 		// Place it 20 units above his head.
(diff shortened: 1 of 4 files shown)


Lizard-Of-Oz:
Another problem is the _color_ of the speech bubble - it reveals the Spy's real team. The color difference is big enough to notice.
Since `C_TFPlayer::GetHeadLabelMaterial` appears to run only on the client, I think something like this would work:
```
if (player is spy && he's enemy to localPlayer && he's disguised as enemy)
  render his speech bubble in localPlayer's team's color
else
  proceed as normal
 ```

File: src/game/shared/voice_status.cpp [hunk 1]
```diff
@@ -227,11 +227,10 @@ void CVoiceStatus::DrawHeadLabels()
 			continue;
 
 		C_BasePlayer *pPlayer = dynamic_cast<C_BasePlayer*>(pClient);
-		if( !pPlayer )
+		if ( !pPlayer )
 			continue;
 
-		// Don't show an icon for dead or spectating players (ie: invisible entities).
-		if( pPlayer->IsPlayerDead() )
```
henke37, line 234: Is the removal of this check intentional?

Bitl: Yes, as it seems pPlayer->IsPlayerDead() is already checked in the src/game/client/c_baseplayer.h for this PR. pPlayer->ShouldShowHeadLabel() is more than enough for this.

FlaminSarge:
It may be worth including a replicated server cvar for this to preserve the legacy behavior (and for that to also affect the "Player is typing" bubble if that gets merged).

Bitl:
> If it doesn't already exist, it may be worth including a replicated server cvar for this to preserve the legacy behavior (and for that to also affect the "Player is typing" bubble if that gets merged). If the cvar does exist, additional values for changing the behavior back to legacy would be nice.

Wrote a basic implementation here if ficool2 wishes to implement this

https://github.com/BitlDevelopmentStudios/source-sdk-2013-bds-base/commit/fa152bef573bb6291ae0316397c0d3d61bf93268

PR #829, opened 2025-02-25
ignore this

weaponcubemap:


Margen67:
What about `SpawnBlood` or `UTIL_BloodDrips`?

wgetJane:
have you tested this? i think blood effects dont get networked to the attacker by default, so wouldn't this just make blood effects never visible for the attacker?

wgetJane:
turning off blood impact effects for our own attacks sounds weird, there would be no feedback for attacks that hit, this pr doesnt feel like it was tested

effects that are predicted/replicated by the client like blood impacts, bullet sparks, bullet decals, etc are not networked to the attacker that causes those effects

instead of just deleting the client's own blood effects altogether, it should also be networked to them

this is what tf2 does, blood impact effects are serverside and spawn on the exact hitpos of the attack (blood decals on players are still clientside, funnily enough)

speedvoltage:
Any chance you have code to propose to better fulfill the overall objective?

wgetJane:
just tested this pr and it does just straight up remove the blood impact particle effect, not ideal

wgetJane:
> Any chance you have code to propose to better fulfill the overall objective?

like i said, tf2 already does it, the code is right here in the same repo

`CBaseEntity::TraceBleed` is what's used to do the blood effects (blood impact particle effects, blood decals sprayed on the walls, etc)

https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/game/shared/baseentity_shared.cpp#L2270

inside `CBaseEntity::TraceAttack` is where it gets called for bullet hits, both on the server and client:

https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/game/shared/baseentity_shared.cpp#L2145

to disable the clientside blood effects in hl2mp, then just comment out this line from the `C_HL2MP_Player::TraceAttack` method:

https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/game/client/hl2mp/c_hl2mp_player.cpp#L191

and here's how tf2 networks the serverside blood effects to the attacker (who normally will be filtered from receiving such effects because they can replicate those effects clientside)

by using `CDisablePredictionFiltering` before calling `TraceBleed` in the `CTFPlayer::TraceAttack` method:

https://github.com/ValveSoftware/source-sdk-2013/blob/0759e2e8e179d5352d81d0d4aaded72c1704b7a9/src/game/server/tf/tf_player.cpp#L8424-L8428

wgetJane:
you can network the accurate serverside blood effects with a modded server

you cannot get rid of the predicted clientside blood effects without a modded client

speedvoltage:
> you can network the accurate serverside blood effects with a modded server
Where? How? Apologies for sounding dumb; this crap has been bugging me for a while.

PR #833, opened 2025-02-26
gdvar: Do not overflow when copying kv buffer[512] to gdvar buffer[129]

### LINKED ISSUE #832: [bug] Potential arbitrary code execution when using malformed FGD files
```GDinputvariable::FromKeyValue``` does unconstrained copying of ```MDkeyvalue.szValue``` (512) to ```GDinputvariable.m_szValue``` (129):
https://github.com/ValveSoftware/source-sdk-2013/blob/238a15a79dd2c3c151e98eea8b39e5f161a93703/src/fgdlib/gdvar.cpp#L588-L594

If ```GDinputvariable::FromKeyValue``` is used to read input variable values from FGD file, malicious actor can craft FGD file with variable value of 129+ length and override stack after  ```GDinputvariable.m_szValue```. Next is a matter of technique to craft executable payload and store it in FGD file. As always, such FGD files may be distributed over network to infect communities.

dimhotepus:
Closes #832

### DIFF
--- a/src/fgdlib/gdvar.cpp
+++ b/src/fgdlib/gdvar.cpp
@@ -591,7 +591,7 @@ void GDinputvariable::FromKeyValue(MDkeyvalue *pkv)
 
 	if (eStoreAs == STRING)
 	{
-		strcpy(m_szValue, pkv->szValue);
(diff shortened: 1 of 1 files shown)


dimhotepus:

PR #836, opened 2025-02-26
Prevent identifying cloaked enemy spies through inspection

M60TM:
# Description
There is an exploit where a disguised enemy spy can still be identified through inspection, even while cloaked. This PR fixes the exploit by adding a check for the cloak state when the target is a disguised enemy spy.


wgetJane:
fixes https://github.com/ValveSoftware/Source-1-Games/issues/6222

hamnTF2:
Tested and LGTM. There is no problem on all cloak pda; stock invisibility watch, cloak and dagger, dead ringer.

mastercoms:
The one problem with this is that it will still technically identify spies due to the blocked trace, still better than before.

M60TM:
Damn, I didn't think about that. Thank you.

wgetJane:
that's a pretty minor issue, could maybe use a trace filter similar to the medigun

wgetJane:
just curious why this was closed, are you reworking it to use the trace filter approach?

PR #837, opened 2025-02-26
Build scripts for dedicated server with use of SteamGameServerHTTP

Myknakryu:
# Description
Modified buildscript allows for building dedicated server linux64 binaries for hl2mp and tf2 against server versions of libtier0 and libvstdlib.
Script requires user to provide both libraries from user from TF2 Dedicated Server install.
This also includes modification to `tf_gc_server.cpp` allowing for use of SteamGameServerHTTP instead of SteamHTTP in dedicated server use case.

### DIFF
--- a/game/mod_tf/gameinfo_server.txt
+++ b/game/mod_tf/gameinfo_server.txt
@@ -1,0 +1,89 @@
+"GameInfo"
+{
+	game					"Frog Fortress 2"
+	type 					multiplayer_only
+	nomodels				1
+	nohimodel 				1
+	nocrosshair				0
+	hidden_maps
+	{
+		"test_speakers"		1
+		"test_hardware"		1
+	}
+	nodegraph 				0
+	GameData				"tf.fgd"
+	AdditionalLocalization	"tf"
+	InstancePath			"maps/instances/"
+	advcrosshair 			1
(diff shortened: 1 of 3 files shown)


misyltoad:
We can probably always use SteamGameServerHTTP on the server, no?

Myknakryu:
Currently testing it, when changed to SteamGameServerHTTP only, 
game starts properly, client/ selfhosted server also works fine with proper connection to inventory server and handling any change in loadout.
Should I commit changed version of `tf_gc_server.cpp` with only `SteamGameServerHTTP`?

rwf93:
On Windows x86-64 with a valid ``SteamAPI_Init`` call during startup, and running from the fork provided. This does not work. GC still is unable to connect due the server being forced into an anonymous user. 

tc2_win64_a4oMpQ2bv1

Linux servers, may connect over IPC. I have not tested. In my testing SteamHTTP was valid, and the branch for checking if it was invalid was never hit.

Myknakryu:
Yeah, probably just using only SteamGameServerHTTP would make code cleaner and not requiring windows script for adding preprocessor define into vcxproj.
Not sure if there could be a case where having client selfhosted server using SteamHTTP would be undesirable.

PR #838, opened 2025-02-26
[TF2] tf_bots work up to 100 players

megascatterbomb:
Modifies tf_bots so they work all the way up to 100 players without failing to spawn in or causing significant class imbalances.

## Implementation Details:
- `CTFBot::m_didReselectClass`
  - Added a Getter so that `CTFBotMainAction::Update` can avoid unnecessarily triggering class reevaluations.
  - Moved `false` assignment to `CTFBot::Event_Killed` to prevent an infinite loop of class reevalulation.
- CVars:
  -  `tf_bot_reevaluate_class_in_spawnroom`:  Now `FCVAR_NONE` (was `FCVAR_CHEATS`).
  - `tf_bot_spawn_use_preset_roster`: Now `FCVAR_NONE` (was `FCVAR_CHEATS`). Now defaults to 0 (was 1).
- Preset Roster (`tf_bot_spawn_use_preset_roster 1`):
  - If team size > 12, the preset roster will be reused for bots 13-24, 25-36, etc.
  - Prevents bots from failing to spawn, but may still result in imbalanced class counts. Still only intended for 12v12.
- Dynamic roster (`tf_bot_spawn_use_preset_roster 0`):
  - Now the default, as it is more dynamic and responsive to gamestate.
  - Class limits specified in rosters scale according to team size. Roster definitions assume team size of 12. Team sizes under 12 unaffected.
  - Team composition levels out with larger team sizes (prevents massive class stacks).
- ~~Added more bot names so there's a pool of at least 100 to choose from.~~ Now in a separate PR: https://github.com/ValveSoftware/source-sdk-2013/pull/892

### DIFF
--- a/src/game/server/tf/bot/behavior/tf_bot_behavior.cpp
+++ b/src/game/server/tf/bot/behavior/tf_bot_behavior.cpp
@@ -51,7 +51,7 @@ ConVar tf_bot_hitscan_range_limit( "tf_bot_hitscan_range_limit", "1800", FCVAR_C
 ConVar tf_bot_always_full_reload( "tf_bot_always_full_reload", "0", FCVAR_CHEAT );
 
 ConVar tf_bot_fire_weapon_allowed( "tf_bot_fire_weapon_allowed", "1", FCVAR_CHEAT, "If zero, TFBots will not pull the trigger of their weapons (but will act like they did)" );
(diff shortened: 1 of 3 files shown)


goodusername123:
I really like this pull request as it fixes and changes things that I have had issues with for years. _However_ the extra bot names here kinda feel really ¿forced? They seem like something better fit for a different PR.

0x5066:
added to that, it's a bad idea to reference the infamous coconut myth as it might just re-ignite the whole discussion again

PR #846, opened 2025-02-26
fix sydney sleeper being able to headshot on crit and sniper crit no scope attribute being useless

TheShermanTanker:
# Description

The sniper rifle has a couple of issues, firstly the sydney sleeper is able to headshot if it is crit boosted, despite the weapon stat sheet saying otherwise. This is due to the CanFireCriticalShot crit check returning true always if crit boosted. CanFireCriticalShot has 2 different meanings depending on whether bIsHeadshot is true or not; If it is, then it actually is not checking for a random crit, but is checking whether this weapon is allowed to crit on headshot. If false, then this is properly being used by the crit chance calculator. To fix this issue, simply check for !bIsHeadshot alongside the crit boost to ensure that the sydney sleeper never headshots when crit boosted. Contrary to what intuition may suggest, this does not make the sydney sleeper unable to crit if it happens to hit the head, as bIsHeadshot does not actually inform CanFireCriticalShot whether the shot was a headshot, but instead it tells CanFireCriticalShot what purpose it is supposed to fulfill. Next, the sniper crit no scope attribute is entirely useless since its check runs after CTFSniperRifle::GetDamageType deletes the hitlocations bit. This is unnecessary as the CanFireCriticalShot check is enough to prevent unscoped snipers from headshotting. To allow this attribute to function properly, delete the logic inside GetDamageType and check whether the player is zoomed in CanFireCriticalShot properly

This fixes https://github.com/ValveSoftware/Source-1-Games/issues/4493 and https://github.com/ValveSoftware/Source-1-Games/issues/4570

Quick question with the above aside: Is it possible for this fix to make it into the game proper? It would be nice to have this fix in base TF2

### DIFF
--- a/src/game/shared/tf/tf_weapon_sniperrifle.cpp
+++ b/src/game/shared/tf/tf_weapon_sniperrifle.cpp
@@ -906,14 +906,7 @@ float CTFSniperRifle::GetProjectileDamage( void )
 //-----------------------------------------------------------------------------
 int	CTFSniperRifle::GetDamageType( void ) const
 {
-	// Only do hit location damage if we're zoomed
-	CTFPlayer *pPlayer = ToTFPlayer( GetPlayerOwner() );
-	if ( pPlayer && pPlayer->m_Shared.InCond( TF_COND_ZOOMED ) )
-		return BaseClass::GetDamageType();
-
-	int nDamageType = BaseClass::GetDamageType() & ~DMG_USE_HITLOCATIONS;
-
-	return nDamageType;
+	return BaseClass::GetDamageType();
 }
 
 //-----------------------------------------------------------------------------
@@ -1010,9 +1003,9 @@ bool CTFSniperRifle::CanFireCriticalShot( bool bIsHeadshot, CBaseEntity *pTarget
 		return false;
 
 	CTFPlayer *pPlayer = GetTFPlayerOwner();
-	if ( pPlayer && pPlayer->m_Shared.IsCritBoosted() )
+	if ( pPlayer && pPlayer->m_Shared.IsCritBoosted() && !bIsHeadshot )
 	{
-		m_bCurrentShotIsHeadshot = bIsHeadshot;
+//  	m_bCurrentShotIsHeadshot = bIsHeadshot;
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


TheShermanTanker:
Ping

src/game/shared/tf/tf_weapon_sniperrifle.cpp
```diff
@@ -906,14 +906,7 @@ float CTFSniperRifle::GetProjectileDamage( void )
...
-	// Only do hit location damage if we're zoomed
-	CTFPlayer *pPlayer = ToTFPlayer( GetPlayerOwner() );
-	if ( pPlayer && pPlayer->m_Shared.InCond( TF_COND_ZOOMED ) )
-		return BaseClass::GetDamageType();
-
-	int nDamageType = BaseClass::GetDamageType() & ~DMG_USE_HITLOCATIONS;
-
-	return nDamageType;
+	return BaseClass::GetDamageType();
 }
```
FlaminSarge, line 910: If this is returning the default GetDamageType at all times, you may just be able to remove CTFSniperRifle::GetDamageType entirely from here and from tf_weapon_sniperrifle.h.

FlaminSarge (commented):
> Quick question with the above aside: Is it possible for this fix to make it into the game proper? It would be nice to have this fix in base TF2

If this is a strictly cosmetic change with no effect on stock gameplay there's a decent chance they accept it as long as it doesn't look like it'll introduce any bugs. I'd also recommend splitting this up into separate fixes for the Sleeper and the no-scope, rather than combining them here.

Out of curiosity, what combination of attributes/end result are you going for here? Removing "sniper no headshot without full charge" from The Classic allows it to crit on any hipfire shot without these changes. Are you looking to make the base tf_weapon_sniperrifle also be able to gain that behavior when hipfiring when "sniper crit no scope" is given, despite it not having the hold-to-charge functionality implemented?

TheShermanTanker:
> > Quick question with the above aside: Is it possible for this fix to make it into the game proper? It would be nice to have this fix in base TF2
> 
> If this is a strictly cosmetic change with no effect on stock gameplay there's a decent chance they accept it as long as it doesn't look like it'll introduce any bugs. I'd also recommend splitting this up into separate fixes for the Sleeper and the no-scope, rather than combining them here.
> 
> Out of curiosity, what combination of attributes/end result are you going for here? Removing "sniper no headshot without full charge" from The Classic allows it to crit on any hipfire shot without these changes. Are you looking to make the base tf_weapon_sniperrifle also be able to gain that behavior when hipfiring when "sniper crit no scope" is given, despite it not having the hold-to-charge functionality implemented?

Noted, will split the fix when I have the time. The intention was to fix the attribute, as right now it doesn't work. This way, the Classic can use the attribute rather than the hardcoded GetDamageType override that currently allows it to crit without being scoped in. The base sniper rifles getting the ability to headshot without being scoped in is a bonus that can help mapmakers and custom game types where that behaviour may be desired.

Tl;Dr, it's to fix the attribute.

TheShermanTanker:
The Sydney Sleeper fix has been reverted for now

PR #849, opened 2025-02-26
Allow impulse 101 on caber

allvei:
Allow WeaponRegenerate() to work on all weapons not just EnergyWeapons(). This allows to use impulse 101 on caber.

# Description


JoriKos:
Does this also affect the Spy-Cicle?

PR #861, opened 2025-02-26
DRAFT: [Tier1] Add move operators

Mooshua:
# Description
This PR aims to add support for move-constructors in the tier1 template library. This PR also adds move operators for some of the tier1 classes themselves so that they can have their resources transferred without invoking expensive copies.

*Ideally* every class should also be given a move operator, using the new CUtlSwap, but that's another big change that I don't have the heart for.

## Build Status

**Currently does not build**, will update as changes are made

## Todo
- [ ] Add required operators to every server/client class
- [ ] Add required operators to vgui
- [ ] Convince someone with engine licensing to patch up the whooole engine

## Why?
The tier1 template library was made before things like move semantics existed, and so they were designed around avoiding copies as much as possible. This was a fantastic idea up until move semantics came along, which promised some pretty neat performance benefits in exchange for always keeping memory initialized. 

It's 2025. It's *way* past time time to rip the band-aid off and allow the codebase to benefit from C++11 features.

### DIFF
--- a/src/public/tier0/platform.h
+++ b/src/public/tier0/platform.h
@@ -1597,6 +1597,12 @@ inline T* CopyConstruct( T* pMemory, T const& src )
 }
 
 template <class T>
+inline T* MoveConstruct( T* pMemory, T&& src )
+{
+	return reinterpret_cast<T*>(::new( pMemory ) T(src));
+}
+
+template <class T>
 inline void Destruct( T* pMemory )
 {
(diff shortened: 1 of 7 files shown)


Mooshua:
I don't like that CUtlMemory is uninitializing but there's too many of them to patch it up properly, i'll just work on move constructor support for the individual containers instead.

PR #862, opened 2025-02-26
all MP games: Fix PLAYER_FLAG_BITS truncating CBasePlayer::m_fFlags sent to clients

copperpixel:
This PR fixes an oversight in all multiplayer Source 2013 games (TF, CS:S, DoD:S, HL1MP, HL2MP) caused by `PLAYER_FLAG_BITS` being set to a lower value than needed, which resulted in some player flags not being sent to clients.

For instance, it created a desync error when a player was pushed by a `trigger_push` entity, causing the client's view to be offset from the real position and breaking some community maps (example below).

The solution is to raise the value of the `PLAYER_FLAG_BITS` macro in `src/public/const.h` to 32 bits so no flags will be truncated.

https://github-production-user-asset-6210df.s3.amazonaws.com/48454166/417359472-d2e2ad0b-bd05-413c-b1d8-0d6da6898866.mp4
##### ^ before the fix was applied

https://github-production-user-asset-6210df.s3.amazonaws.com/48454166/417361827-4f7b8ef0-a68a-43fa-a7ae-b6063949272f.mp4
##### ^ after the fix was applied

Also closes issues:
- https://github.com/ValveSoftware/Source-1-Games/issues/7076
- https://github.com/ValveSoftware/Source-1-Games/issues/7085

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -8101,14 +8101,14 @@ void CMovementSpeedMod::InputSpeedMod(inputdata_t &data)
 	}
 }
 
-
-void SendProxy_CropFlagsToPlayerFlagBitsLength( const SendProp *pProp, const void *pStruct, const void *pVarData, DVariant *pOut, int iElement, int objectID)
+void SendProxy_CropFlagsToPlayerFlagBitsLength( const SendProp* pProp, const void* pStruct, const void* pVarData, DVariant* pOut, int iElement, int objectID )
 {
-	int mask = (1<<PLAYER_FLAG_BITS) - 1;
-	int data = *(int *)pVarData;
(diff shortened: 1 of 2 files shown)


misyltoad:

copperpixel:
Yes, I have read the terms.

Adrianilloo:
So, it turns out official HL2DM still uses the previously reduced `m_fFlags` sent bits (10), resulting in incompatible Network Class Tables (preventing clients to connect from SSDK2013 based servers). Is the game updating to sync with the 32-bit restoration?

misyltoad:
There is no guaranteed compatibility with SDK2013 and the shipping titles on Steam.

It'll get updated at some point, but could break at any time.

PR #864, opened 2025-02-27
[TF2] Fix for PD/RD score edge case

rowedahelicon:
# Description
An alternative approach by FlaminSarge has been provided here: https://github.com/ValveSoftware/source-sdk-2013/pull/947

Designed to fix an edge case as described here: https://github.com/ValveSoftware/Source-1-Games/issues/5338

Video demonstration: https://youtu.be/S2UTHr6Zbuc?si=-AhtOK_ibZ4u6ONG

- Teams with scores equal to or above a updated score target lower than it was previously are granted victory upon player disconnect.
- In an event both teams are above the new score target, the team with a higher score is the winner
- In an event where both teams are tied with a score above the new score target, the team who reached that score first is the winner

### DIFF
--- a/src/game/shared/tf/tf_logic_robot_destruction.cpp
+++ b/src/game/shared/tf/tf_logic_robot_destruction.cpp
@@ -754,30 +754,7 @@ int CTFRobotDestructionLogic::ApproachTeamTargetScore( int nTeam, int nApproachS
 
 		if ( nNewScore == m_nMaxPoints )
 		{
-			if ( nTeam == TF_TEAM_RED )
-			{
-				m_OnRedHitMaxPoints.FireOutput( this, this );
-				m_flRedFinaleEndTime = gpGlobals->curtime + m_flFinaleLength;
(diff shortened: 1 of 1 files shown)


Fraeven (approved):

FlaminSarge:
Would this not be better as a requirement that updated score limits (on player disconnect) be at least 1 higher than the current max team score? Simplifies the impl by a lot.

rowedahelicon:
That would probably be a more optimal way to telegraph what's happening as a result of players leaving, yeah, my other idea was to suggest that the target score never go backwards in the first place

FlaminSarge:
> Though it does admittedly use more code to handle a fringe event.

I'd say open a PR with the alternative and see how much less code it uses. Less changes also means less chances for unforseen consequences.

rowedahelicon:
> > Though it does admittedly use more code to handle a fringe event.
> 
> I'd say open a PR with the alternative and see how much less code it uses. Less changes also means less chances for unforseen consequences.

Done: https://github.com/ValveSoftware/source-sdk-2013/pull/947 :3

PR #868, opened 2025-02-27
tf2: Add convar to allow inspecting MvM upgrades

Mikusch:
Currently you can only inspect the upgrades of Steam friends or matchmaking party members. This behavior is undesirable for community MvM servers and there is currently no way to change it.

This PR adds a new convar `tf_mvm_allow_upgrade_inspect` to allow server operators to modify the default behavior.

`-1` (default) = Current behavior (show only for friends and party members)
`0` = Never allow
`1` = Always allow

### DIFF
--- a/src/game/client/tf/tf_hud_inspectpanel.cpp
+++ b/src/game/client/tf/tf_hud_inspectpanel.cpp
@@ -29,6 +29,8 @@
 #include "tier0/memdbgon.h"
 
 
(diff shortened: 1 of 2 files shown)


Ashe-tf:
why close?

Mikusch:
> why close?

Deleted my original fork on accident. Resubmitted as #1364

PR #885, opened 2025-02-27
[HL2MP] Fix compile error in HL2MP when using GLOWS_ENABLE

Blixibon:
# Description
This pull request fixes a compile error in HL2MP when using the `GLOWS_ENABLE` preprocessor.

`GLOWS_ENABLE` enables code that can add glow outline effects on entities. It is used in TF2 by default, but it can be enabled for any SDK project. However, while it can be compiled as-is under base HL2, there is a dependency within an `#ifdef TF_CLIENT_DLL` in `c_team_train_watcher.cpp` which prevents it from compiling under HL2MP.

As this dependency (`teamplayroundbased_gamerules.h`) is not necessarily TF2-specific, this pull request moves the dependency outside of the `#ifdef`.

### DIFF
--- a/src/game/client/c_team_train_watcher.cpp
+++ b/src/game/client/c_team_train_watcher.cpp
@@ -8,10 +8,10 @@
 #include "c_team_train_watcher.h"
 #include "igameevents.h"
 #include "c_team_objectiveresource.h"
+#include "teamplayroundbased_gamerules.h"
 
 #ifdef TF_CLIENT_DLL
 #include "tf_shareddefs.h"
-#include "teamplayroundbased_gamerules.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


iojuedioe9rd (approved):

Bitl:
This causes an error in C_TeamTrainWatcher::OnDataChanged in non TF projects. I feel like #ifdef TF_CLIENT_DLL should be surrounded around the if ( ObjectiveResource() ) block, as shown below.

image

Blixibon:
> This causes an error in C_TeamTrainWatcher::OnDataChanged in non TF projects.

I'm not experiencing an error when compiling for HL2MP on Windows, with or without `GLOWS_ENABLE`. What is the error you're encountering?

PR #887, opened 2025-02-27
Fix inefficient lookup for item definitions by name

ficool2:
When `items_game.txt` is parsed by the client and server, they do a LOT of item definition lookups by the name. Unfortunately, `CEconItemSchema::GetItemDefinitionByName` searches the list of items manually which is terribly inefficient. In fact, it's so inefficient that it added 1-2 seconds of loading time on startup! This PR resolves this by using a dictionary mapping from name to item definition, reducing the time complexity from O(N) -> O(1). This reduces loading time on startup by at least 1 second.

Profiler capture before (in milliseconds)

before

Profiler capture after (in milliseconds)

after

### DIFF
--- a/src/game/shared/econ/econ_item_schema.cpp
+++ b/src/game/shared/econ/econ_item_schema.cpp
@@ -4293,6 +4293,7 @@ void CEconItemSchema::Reset( void )
 	m_vecAttributeTypes.Purge();
 	m_mapItems.PurgeAndDeleteElements();
 	m_mapItems.Purge();
+	m_mapItemsName.Purge();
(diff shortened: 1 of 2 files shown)


HyperionCoding:
I finally tested this in MvM and here are the results

Without patch:
```
tf_logic_mann_vs_machine(.PEAVCBaseEntity@@) thinking for 46.81 ms!!!
```
With patch:
```
tf_logic_mann_vs_machine(.PEAVCBaseEntity@@) thinking for 34.33 ms!!!
```

This is without cosmetics on robots so the gain will be larger the more items they have.

PR #888, opened 2025-02-27
Fix debug build crash and CUtlBuffer assertion spam

TorutheRedFox:
# Description
Fixes CUtlBuffer assertion spam (https://developer.valvesoftware.com/wiki/SDK_Known_Issues_List#Assertion_Failed_in_utlbuffer.cpp) and makes failing to retrieve CEcomItem fail gracefully, allowing debug builds to run without crashing until the underlying issue with the inventory is fixed.

### DIFF
--- a/src/game/shared/tf/tf_item_inventory.cpp
+++ b/src/game/shared/tf/tf_item_inventory.cpp
@@ -965,7 +965,9 @@ void CTFPlayerInventory::LoadLocalLoadout()
 
 					CEconItemView *pItem = GetInventoryItemByItemID(uItemId);
 					if (pItem) {
-						pItem->GetSOCData()->Equip(iClass, iSlot);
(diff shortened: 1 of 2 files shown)


zxz41:
`Assert( nSize != 0 );` is technically not a bug. If you're allocating a buffer with a zero size, something have either gone wrong, or you're likely doing something unintentionally. I haven't checked if `-1` have any intended effect.

TorutheRedFox:
> `Assert( nSize != 0 );` is technically not a bug. If you're allocating a buffer with a zero size, something have either gone wrong, or you're likely doing something unintentionally. I haven't checked if `-1` have any intended effect.
> 
> It's why this assert is being hit on Debug builds currently for the GCSDK interacting with CUtlBuffer.

the underlying issue couldn't be fixed (by us, the public) without access to the GCSDK source as it's the one creating the size 0 buffers

PR #893, opened 2025-02-28
[TF2] Fix for Casual late-joiners seeing Competitive logo on Match Status HUD doors

rabscootle:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/3516

Code has been updated to refresh the Match Status HUD doors' skin and submodel at the conclusion of a match. Check is similar to what is performed at start of match.

Due to Casual Mode's original behavior (no late joining, no abandoning) skin and submodel are currently only set at the beginning of a match for all gamemodes that support it. Late-joiners in Casual mode will see the Competitive logo at the end of the match if the model was not updated at the start of a previous match.

image

### DIFF
--- a/src/game/client/tf/tf_hud_match_status.cpp
+++ b/src/game/client/tf/tf_hud_match_status.cpp
@@ -575,7 +575,7 @@ void CTFHudMatchStatus::FireGameEvent( IGameEvent * event )
 
 		const IMatchGroupDescription* pMatchDesc = GetMatchGroupDescription( TFGameRules()->GetCurrentMatchGroup() );
 
-		//FIX: Refresh versus doors so Casual late-joiners do not see the wrong skin
+		// FIX: Refresh versus doors so late-joiners do not see the wrong skin
 		int nSkin = 0;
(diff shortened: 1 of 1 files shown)


src/game/client/tf/tf_hud_match_status.cpp
```diff
@@ -574,6 +574,17 @@ void CTFHudMatchStatus::FireGameEvent( IGameEvent * event )
 		}
 
 		const IMatchGroupDescription* pMatchDesc = GetMatchGroupDescription( TFGameRules()->GetCurrentMatchGroup() );
+
+		// FIX: Refresh versus doors so late-joiners do not see the wrong skin
+		int nSkin = 0;
+		int nSubModel = 0;
+		if (pMatchDesc->BGetRoundDoorParameters(nSkin, nSubModel))
```
mastercoms, line 581: Missing NULL check for pMatchDesc. See below if statement on 589. It checks if pMatchDesc is non-null, thus we should preserve null safety in the new check as well.

PR #899, opened 2025-02-28
Fix most medigun prediction errors

treacherousfiend:
# Description
Currently, healing people with any medigun will cause constant prediction errors, because the code which calculates how much ubercharge you get on any given tick is server only, unless you are lucky and are in a situation where you are getting exactly the default ubercharge per tick (stock medigun, damaged patient, no critheals, etc.)

This PR fixes that by making all possible server only code for this calculation also be on the client. The only exception to this is an unused weapon attribute which would modify ubercharge rate as long as you weren't in a respawn room. The functions it calls are server only, and would require changes to how respawn room entities are handled on the client to fix, so I decided that it would be out of scope for this PR.
Due to the one last part, I've simply moved the `#ifdef GAME_DLL` to only encompass that code rather than removing it entirely.

In my testing this change causes no issues.

I'm not entirely sure why it was server only in the first place? Maybe the idea was that the client doesn't really need to know the exact ubercharge percentage except in specific scenarios, but I don't think prediction errors every tick is a good tradeoff for the client running slightly less code for ubercharge calculations, especially for the medic. This bug is also known to cause heavy view jittering on listen servers, although it doesn't seem to be an issue on dedicated servers.

### DIFF
--- a/src/game/shared/tf/tf_weapon_medigun.cpp
+++ b/src/game/shared/tf/tf_weapon_medigun.cpp
@@ -1252,7 +1252,6 @@ bool CWeaponMedigun::FindAndHealTargets( void )
 
 			if ( pTFPlayer && weapon_medigun_charge_rate.GetFloat() )
 			{
-#ifdef GAME_DLL
 				int iBoostMax = floor( pTFPlayer->m_Shared.GetMaxBuffedHealth() * 0.95);
 				float flChargeModifier = 1.f;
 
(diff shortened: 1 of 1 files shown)


wgetJane:
does this work with different ubercharge rates like with the quick-fix?

treacherousfiend:
Likely yes, I did not test all of the mediguns, but did do all of my testing with the Vaccinator since I was fixing another issue at the time, and it removed all prediction errors outside of normal ones due to networking instability.

PR #905, opened 2025-02-28
https: Prevent man-in-the-middle attacks due to using http proto and server-side redirection

dimhotepus:
See https://github.com/ValveSoftware/Source-1-Games/issues/7120

### DIFF
--- a/src/game/client/econ/econ_trading.cpp
+++ b/src/game/client/econ/econ_trading.cpp
@@ -544,7 +544,7 @@ const char* UniverseToCommunityURL( EUniverse universe )
 	default: // return public if we don't have a better guess.
 	case k_EUniversePublic: return "https://steamcommunity.com";
 	case k_EUniverseBeta: return "https://beta.steamcommunity.com";
-	case k_EUniverseDev: return "https://localhost/community";
+	case k_EUniverseDev: return "http://localhost/community";
 	}
 
 	// Should never get here.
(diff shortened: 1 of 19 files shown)


halamix2:
- I'd suggest skip changing youtube.com links, since the Youtube support is broken anyway (see https://github.com/ValveSoftware/Source-1-Games/issues/2277)

src/game/client/econ/econ_trading.cpp
```diff
@@ -544,7 +544,7 @@ const char* UniverseToCommunityURL( EUniverse universe )
 	default: // return public if we don't have a better guess.
 	case k_EUniversePublic: return "https://steamcommunity.com";
 	case k_EUniverseBeta: return "https://beta.steamcommunity.com";
-	case k_EUniverseDev: return "https://localhost/community";
+	case k_EUniverseDev: return "http://localhost/community";
```
halamix2, line 547: question is why it was https used here in the first place. If the local server has a self-signed cert/ custom root CA(s) then this change might even break support for local tests and development

`EUniverse` at https://partner.steamgames.com/doc/api/steam_api

dimhotepus: Good point.

I've looked for dev universe usage over the code base and found other places use ```http://``` for localhost. So I guessed as there are much more places with ```http://``` for localhost, it is the valid one. 

I may be wrong and all places should be changed to ```https://``` thought.

ktwrd: MITM attacks also do not matter at all for localhost, and you probably shouldn't mess with stuff in relation to internal valve stuff :p

PR #906, opened 2025-02-28
[TF2] Fix ProgressBar.FgColor being overwritten

CriticalFlaw:
# Description
On every tick, instead of using the `ProgressBar.FgColor` set in the clientscheme, the game will default to a pure white color. With this change the set FgColor will be used.

The red flash that happens when the item meter is full can be set with the `ProgressBar.FlashColor` property in the clientscheme. In case its absent the red color used today will be used instead.

This would address https://github.com/ValveSoftware/Source-1-Games/issues/2498

Animation


gidi30:
This breaks the Cow mangler and thermal thruster turning the bar red under certain circumstances due to them doing it in `GetProgressBarColor` which is no longer called here.
A better solution would be modify each version of `GetProgressBarColor` to return the foreground color where it currently returns white instead of replacing it completely.

CriticalFlaw:
> This breaks the Cow mangler and thermal thruster turning the bar red under certain circumstances due to them doing it in `GetProgressBarColor` which is no longer called here. A better solution would be modify each version of `GetProgressBarColor` to return the foreground color where it currently returns white instead of replacing it completely.

It is intentional to have every weapon use the same color when the meter is full, but I think you meant for when they are charging. Made a correction to have `GetProgressBarColor` still be called and just invoke the set FgColor when appropriate. That should restore the original behaviour

gidi30:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/906#issuecomment-3508853522

Now it will only use the foreground colour for the Cow mangler and Thermal thruster because the base versions of `GetProgressBarColor` which always return white were not modified.

CriticalFlaw:
> > Replying to [#906 (comment)](https://github.com/ValveSoftware/source-sdk-2013/pull/906#issuecomment-3508853522)
> 
> Now it will only use the foreground colour for the Cow mangler and Thermal thruster because the base versions of `GetProgressBarColor` which always return white were not modified.

You're talking about these ones?

image

CriticalFlaw:
Made the final correction so that the Cow Mangler and Thermal Thruster use the intended colors.

PR #910, opened 2025-02-28
Make info_null a serverside-only entity

SirYodaJedi:
# Description
Inheriting CServerOnlyEntity instead of CBaseEntity reduces potential for entity limit crashes, as CNullEntity still exists for the first tick.

### DIFF
--- a/src/game/server/subs.cpp
+++ b/src/game/server/subs.cpp
@@ -22,10 +22,10 @@ void CPointEntity::Spawn( void )
 }
 
 
-class CNullEntity : public CBaseEntity
+class CNullEntity : public CServerOnlyEntity
 {
 public:
-	DECLARE_CLASS( CNullEntity, CBaseEntity );
+	DECLARE_CLASS( CNullEntity, CServerOnlyEntity );
 
 	void Spawn( void );
 };


A1exKH (approved):
LGTM.

Bitl:
wouldn't this still add to the entity limit count? this entity is defined in the server dlls already

maybe removing info_null form the code entirely and adding an instance in the compiler could work?

A1exKH (approved):
LGTM.

SirYodaJedi:
> wouldn't this still add to the entity limit count? this entity is defined in the server dlls already

It would count toward the *separate* serverside entity limit, instead of the networked entity limit (edict limit). There's a lot more room in the serverside entity space, which mostly only used by logic entities. The edict limit is the cause of entity limit crashes 99.9% of the time.

> maybe removing info_null form the code entirely and adding code that detects it in the compiler would work better (ala how it detects info_ladder)?

info_ladder *isn't* detected. It's just treated as an unknown entity, causing console spam (it's better to use func_detail with a %compileladder material outside of L4D because of this). The entity that info_ladder is created from, func_ladder, is removed by VBSP, which is undesirable for info_null, which needs to be read by VRAD. Could a function be added to VRAD that removes info_null and func_null? Perhaps, but VRAD doesn't editing the entity lump to begin with (never was designed to), so significantly more modification would be necessary.

In my own fork of VBSP, I rename info_null to _null, which does prevent it from counting towards either entity limit, but results in a bunch of the aforementioned console spam. I determined that simply making info_null not an edict was more clean and practical, and prevents it from causing crashes as long as you don't have a couple thousand of them (for what it's worth, I've also pushed the same commit to Mapbase mapbase-source/source-sdk-2013#324 ).

> LGTM.

You said this already.

SirYodaJedi:
As per #1046, It seems like it may be possible to do a similar change for infodecal and info_projecteddecal; those seem to use temp ents to network their location to the client instead of edicts. If I can successfully verify this, I'll add that to this PR.

SirYodaJedi:
Added the decal de-edicting, after some testing by @BreakinBenny. I figure this doesn't count as a duplicate of #1046, because of that PR being closed by it's OP.

PR #912, opened 2025-03-01
[TF2] Prevent overflow of CBaseTeamObjectiveResource.m_iNumTeamMember

megascatterbomb:
# Description

CBaseTeamObjectiveResource.m_iNumTeamMember has only 4 bits to communicate how many cappers are standing on a control point. When >15 cappers are present this triggers an overflow, causing an incorrect number of cappers to be displayed and client/server desync on the capture progress indicators.

This PR adds the shared defines `MAX_TRANSMIT_CAPPERS_BITS` and `MAX_TRANSMIT_CAPPERS` which are used by the relevant functions to protect against overflows.

To stop desyncs, the capture speed no longer increases above x15. This change will barely affect gameplay as the [formula](https://wiki.teamfortress.com/wiki/Control_point_timing#Base_capture_times) used to determine capture time has a negligible impact above x15.

This approach was taken to avoid breaking demos. In the future, if the number of bits is increased, only the `MAX_TRANSMIT_CAPPERS_BITS` needs to be changed.

### DIFF
--- a/src/game/server/team_objectiveresource.h
+++ b/src/game/server/team_objectiveresource.h
@@ -12,6 +12,7 @@
 
 #include "shareddefs.h"
 
+// Number of bits used to transmit the number of point cappers to the client.
 #define MAX_TRANSMIT_CAPPERS_BITS		4
 #define MAX_TRANSMIT_CAPPERS			(1 << MAX_TRANSMIT_CAPPERS_BITS) - 1
 
--- a/src/game/server/trigger_area_capture.cpp
+++ b/src/game/server/trigger_area_capture.cpp
@@ -469,12 +469,20 @@ void CTriggerAreaCapture::CaptureThink( void )
 
 		float flReduction = flTimeDelta;
 		if ( CaptureModeScalesWithPlayers() )
-		{
-			// Diminishing returns for successive players.
-			for ( int i = 1; i < m_TeamData[m_nTeamInZone].iNumTouching && i < MAX_TRANSMIT_CAPPERS; i++ )
+		{	
+#ifdef TF_DLL
+			// Diminishing returns for successive players. MAX_TRANSMIT_CAPPERS prevents overflow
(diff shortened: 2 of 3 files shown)


FlaminSarge:
Given that 100player exists it is probably permissible to just increase it to 8 bits rather than capping it.

wgetJane:
@megascatterbomb how bad is the desync with like ~18 players at the current behaviour vs clamping the value that's sent to 15 without clamping the cap speed?

wgetJane:
> Simply clamping what's sent to the client at 15 without changing the cap time to match doesn't cause nearly as bad of a desync, however I don't see the harm in making the progress indicator 100% reliable.

imo that's preferable over clamping the cap speed, though i guess ppl that play on increased maxplayers servers have more of a say on this issue

FlaminSarge:
> That would require breaking demos, which isn't worth it for a single visual bug.

Wait, under what conditions do demos break? Fields widening seems to be one; what other conditions will do so?

FlaminSarge:
> > > That would require breaking demos, which isn't worth it for a single visual bug.
> > 
> > 
> > Wait, under what conditions do demos break? Fields widening seems to be one; what other conditions will do so?
> 
> Aside from directly modifying the bit count, I think _most_ changes to how those bits are interpreted by the recipient will break demos to some extent. If this PR were implemented and an old demo was played back, it would simply result in the same kind of desync that we already deal with. In other cases there may be more serious problems which could prevent playback.
> 
> TL:DR Changing the bit format always breaks; altering interpretation of data has to be validated on a case by case basis.

Alright, makes sense.
However, as more and more fixes come in, trying to maintain demo compatibility is going to get tougher and tougher, and if demo compat is enforced for PR's it's going to severely limit what changes can go in.

Realistically, Valve needs to branch TF2 on Steam right now before any of these PRs go to live, for the sake of preserving demos made before SDK was made public. Then fixes that come in wouldn't affect the ability to play back those demos, as long as the branch exists (on Steam). They've done this before for Jungle Inferno's release.

PR #914, opened 2025-03-01
Fix floating health bars not rendering for giant robots in MvM under certain conditions + add better support for see_enemy_health on all classes

Marxvee:
This introduces a QoL fix where if tf_hud_target_id_disable_floating_health is 1, it will now still display the health bars for entities that do not have a visible target ID to fall back to, such as enemy mini bosses (aka giant robots in MvM) and robot destruction bots. Before this fix, having floating health bars disabled would result in you being unable to view the health of giant robots whereas you normally should be able to.

EDIT: Now added a feature to allow the Solemn Vow's "allows you to see enemy health" attribute to work when classes other than the Medic (and Heavy?) have the attribute applied to them. Enemies are also now differentiated with the underused #TF_playerid_diffteam localization string. 
This also removes 3 redundant checks involving checking if the local player is Medic or Heavy, and removes the second check if the target is a enemy disguised Spy.


Whurrhurr:
there's another issue with floating health bars, if the place that the healthbar would normally be at is outside the user's screenspace, it does not show up at all, even when the spot where the healthbar would normally show up is pulled into view. here's an example of this:

https://github.com/user-attachments/assets/e57490bc-a0e4-4f5b-9f68-e7744d38a9db

in cases like these, i think the floating healthbar should show up on the edge of your screen instead of just not showing up at all.

Marxvee:
i saved the file under a different editor and now the file diff is messed up... I CRY....

> there's another issue with floating health bars, if the place that the healthbar would normally be at is outside the user's screenspace, it does not show up at all, even when the spot where the healthbar would normally show up is pulled into view. here's an example of this:
> 
>  2025-02-28.21-50-11.mp4 
> in cases like these, i think the floating healthbar should show up on the edge of your screen instead of just not showing up at all.

i'm currently unsure as to why the health icon unloads like that, but i'll try to look into this at some point!

Marxvee:
Ah, whoops! I definitely pushed personal changes to the wrong branch, let me try to fix it...

PR #915, opened 2025-03-01
[TF2] Make engineer bots care about teleporters

LordofCreepers:
This is a slight rework that now makes engineer bots remove sappers/repair/upgrade their teleporters like any other buildings. In order to not have these changes deviate their behavior from the original too much, teleporters are prioritized the least (in particular, they only repair teleporters when every other building is already at full health, upgrade them when every other is already upgraded, and remove sappers when all other buildings don't have sappers on them). Also comes with reordering and splitting original building maintenance code into separate functions for ease of further modification

https://github.com/user-attachments/assets/c1cc5693-b57f-4ebf-81ab-c271da76eee0

https://github.com/user-attachments/assets/8fc45326-4ecd-41c4-8b65-021ec8eb2767

### DIFF
--- a/src/game/server/tf/bot/behavior/engineer/tf_bot_engineer_building.cpp
+++ b/src/game/server/tf/bot/behavior/engineer/tf_bot_engineer_building.cpp
@@ -231,6 +231,41 @@ bool CTFBotEngineerBuilding::IsInPositionToWork(
 }
 
 //---------------------------------------------------------------------------------------------
+// Proccesses navigation towards ideal working spot
+void CTFBotEngineerBuilding::NavigateToWorkingSpot( CTFBot* me, const Vector& spot )
+{
+	if ( m_repathTimer.IsElapsed() )
+	{
+		// Periodically recalculate the path in case if our building... teleports... I guess...
+		m_repathTimer.Start( RandomFloat( 1.0f, 2.0f ) );
+
+		CTFBotPathCost cost( me, FASTEST_ROUTE );
+		m_path.Compute( me, spot, cost );
+	}
+
+	// Go to building
+	m_path.Update( me );
+}
+
+//---------------------------------------------------------------------------------------------
+// Proccesses maintaining buildings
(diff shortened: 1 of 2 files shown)


LordofCreepers:
> Couple issues:
> 
> * Engineers seem to prioritize upgrading their entrance before building their exit.
> * Once they start upgrading their entrance they tend to not stop, even after running out of metal.
> 
> Even after engaging in a fight he went straight back to physically abusing his tele.
> 
>  tele_glitch.mp4

Should be fixed now

LordofCreepers:
Ok, I lied. NOW fixed

Mickeon:
The PR is riddled with "_Dispenser_" being misspelled as "_Dispencer_". I would heavily suggest addressing that.

LordofCreepers:
> Issues I called out appear to be fixed. Good work!
> 
> Everything seems to work as intended _if_ the Engineer places their entrance teleporter, however if they don't (for whatever reason), then they won't place any teles until they die.
> 
> Not sure what if any change should be made here. Any thoughts?

Frankly speaking, hard to say. Building placement is done through separate tasks (individual tasks for each building) for engineers. They're in a separate file, which is why I didn't look at them this thorough. Changes obviously can be done (and I would've loved to do them), but, it is likely that any placement of teleporter entrance is extreme braindead and will require some more work than just adding a check to achieve what you want (my speculation on this part though). Days off are over, so, despite the fact that I could probably study the current placement code, I can probably only do something about it at the end of the week (although maybe I could ditch some work a little earlier in exchange for this, but, I'll have to see then). Besides, I still have some tweaks I want to do to this, and I'm not even sure if what's you're suggesting is more of a material for an entirely new PR

TL;DR: Sounds more than fine by me, but Idk how much work it'll be and I don't have that much free time now

LordofCreepers:
So, for now this is it. You guys can check if my new tweaks cause any problems, but, for now, I'm done with this PR unless it gets merged

PR #916, opened 2025-03-01
show the actual path of steam_api.dll instead of printf formatting

flashintv:
normally if launcher fails to find steam_api.dll, an error message with the printf formatting gets shown instead of the actual path to where the dll is supposed to be

### DIFF
--- a/src/launcher_main/main.cpp
+++ b/src/launcher_main/main.cpp
@@ -157,7 +157,7 @@ static bool LoadSteam( const char *pRootDir )
 	s_SteamModule = Launcher_LoadModule(szPathBuffer);
 	if ( !s_SteamModule )
 	{
-		char szErrorBuffer[4096];
+		char szErrorBuffer[4140];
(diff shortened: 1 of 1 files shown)


flashintv:
it seems i forgot to increase the buffer size if somehow path is longer than 4052

src/launcher_main/main.cpp
```diff
@@ -149,15 +149,19 @@ static bool LoadSteam( const char *pRootDir )
...
-	char szBuffer[4096];
+	char szPathBuffer[4096];
 	// Assemble the full path to our "steam_api.dll"
-	_snprintf( szBuffer, sizeof( szBuffer ), STEAM_API_DLL_PATH, pRootDir );
-	szBuffer[sizeof( szBuffer ) - 1] = '\0';
-
-	s_SteamModule = Launcher_LoadModule( szBuffer );
+	_snprintf(szPathBuffer, sizeof(szPathBuffer), STEAM_API_DLL_PATH, pRootDir );
+	szPathBuffer[sizeof(szPathBuffer) - 1] = '\0';
+
```
Margen67, line 156:
```suggestion

```

PR #917, opened 2025-03-01
Fix bots never using their Seek and Destroy (Deathmatch AI)

ShowinGrowin:
# Description

Bots are intended to deathmatch when they don't understand the current gamemode. This feature was eventually broken when the cvar tf_bot_offense_must_push_time was added to the game. This check is only relevant in specific gamemodes.

Since only specific gamemodes account for this logic, it causes them to not go into their "deathmatch" mode when they're supposed to. I added a check for said gamemodes so they will still use "seek and destroy" logic if there is no other supported gamemode logic to go by.

### DIFF
--- a/src/game/server/tf/bot/behavior/tf_bot_seek_and_destroy.cpp
+++ b/src/game/server/tf/bot/behavior/tf_bot_seek_and_destroy.cpp
@@ -87,6 +87,17 @@ ActionResult< CTFBot >	CTFBotSeekAndDestroy::Update( CTFBot *me, float interval 
 		{
 			return Done( "Time to push for the objective" );
 		}
+
+
+		// Added proper fix to the bot offense push time bug that causes seek and destroy to never happen
+		// Basically just check for these gamemodes and otherwise allow seek and destroy at all times
+		if (TFGameRules()->GetGameType() == TF_GAMETYPE_ESCORT || TFGameRules()->GetGameType() == TF_GAMETYPE_CP)
+		{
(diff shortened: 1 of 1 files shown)


ShowinGrowin:
> I think you forgot to delete the original code just before the part you added. As currently written there'll be no change in behaviour as the condition on L86 will pass before the identical condition on L96 (that you wrapped with additional checks) is ever evaluated.
> 
> image

Yeah my bad. I originally had it in the master fork but decided to move it to a branch and ended up accidentally keeping a copy of the original code. Now it's fixed.

ShowinGrowin:
Tested it out and it worked. I've added that change alongside the original one as well. Thanks!

Ashetf2:
How sure how it is supposed to work but it doesn't in my end

https://github.com/user-attachments/assets/2d6bd31c-d870-4188-b0b4-f7c1fa4f8e05

ShowinGrowin:
The problem with Payload Race is bots think the gamemode is payload. They aren't really set up to properly understand and support two payload carts though.

The seek and destroy AI will occur when the bots have literally no game logic to use. Which should work on most unsupported maps/gamemodes. There are a few exceptions like CTF and PLR though. In these they somewhat understand the gamemode. I think these would preferably be fixed with actual real support for these gamemodes (especially for PLR since they already understand payload) instead of using their deathmath AI.

mastercoms:
Wouldn't it be cleaner to just return FLT_MAX for conditions you're checking? And perhaps even better to implement timers for these modes?

Issue #691, opened 2025-02-20
KeyValues::LoadFromFile seems to ignore overridden files


TorutheRedFox:
KeyValues::LoadFromFile doesn't seem to be loading a custom version of GameMenu.res that I put in to the files for the TF2 SDK, both from the custom folder and as a loose file

Modifying the code to make it read a differently named file works just fine, but for some reason, leaving it as-is makes the original file take priority.

PR #920, opened 2025-03-01
Fix missing vgui entry in gameinfo.txt

TorutheRedFox:
# Description

Addresses https://github.com/ValveSoftware/source-sdk-2013/issues/691 by adding a missing vgui path entry into gameinfo.txt

Note: I am aware that vgui_write isn't actually a thing but I added it anyway in case someone would like to make it a thing in the future (plus it doesn't look out of place).


TorutheRedFox:
created from wrong base branch

PR #927, opened 2025-03-02
Fix players clipping through displacements by uncrouching

chrb22:
# Description
On servers hosted on Linux it is possible that the origin calculated in ducking-related functions gets slightly horizontally shifted because of floating point precision errors. This only happens around powers of 2 and is caused by the compiler being allowed to reorder the instructions to something mathematically equivalent (`-ffast-math`). The expressions as they are written in the code would not lead to this kind of error happening, so the fix is to simply tell the compiler not to reorder the instructions in the relevant functions (`CGameMovement::CanUnduck`, `CGameMovement::FinishUnDuck`, `CGameMovement::FinishDuck`).

This fixes an exploit that let players uncrouch into displacements because `CanUnduck` incorrectly reported that there was nothing above the player whereafter `FinishUnDuck` moved the player into the displacement. `FinishUnDuck` (after `CanUnduck` fix) and `FinishDuck` could also let players clip into geometry by themselves, but it was a lot harder to find and perform as the player needed to get really close to the surface before uncrouching or crouching for anything to happen (usually done by wiggling in the air).

Videos of `CanUnduck`+`FinishUnDuck` and `FinishDuck` causing the player to clip into the geometry. Notice the stuck messages in the top right. The player automatically gets teleported out if they don't move by the regularly occurring unstuck routine.

https://github.com/user-attachments/assets/436e9adc-4725-4945-ae0f-f5152feefdb6

https://github.com/user-attachments/assets/d4f95f1b-06c0-4c58-b747-b954aabca3f6

### DIFF
--- a/src/game/shared/gamemovement.cpp
+++ b/src/game/shared/gamemovement.cpp
@@ -4059,6 +4059,7 @@ void CGameMovement::FixPlayerCrouchStuck( bool upward )
 	mv->SetAbsOrigin( test ); // Failed
 }
 
+[[gnu::optimize("no-associative-math")]]
 bool CGameMovement::CanUnduck()
 {
 	int i;
@@ -4098,6 +4099,7 @@ bool CGameMovement::CanUnduck()
 //-----------------------------------------------------------------------------
 // Purpose: Stop ducking
 //-----------------------------------------------------------------------------
+[[gnu::optimize("no-associative-math")]]
 void CGameMovement::FinishUnDuck( void )
 {
 	int i;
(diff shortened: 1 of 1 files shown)


misyltoad:
I wonder if we want to instead do this for all of gamemovement related code. There's likely more subtleties.

chrb22:
Oh I'm guessing you meant expanding the fix to all functions in the movement code. In that case I'm not really sure either. I initially understood it as you being careful about a change that touches movement in all games.

chrb22:
I went through `gamemovement.cpp` and `tf_gamemovement.cpp` and it was only two extra ducking-related functions that could have this kind of problem.

groovyvs:
apart from OOB, this bug is harmless

PR #928, opened 2025-03-02
[TF2] Make the Playermodel in the HUD use the players custom model if applicable

gidi30:
# Description
Currently the Playermodel Panel will always display the classes default model (or the Yeti model if the corresponding taunt is used), Which causes a discrepancy if the Map or Server changes the players model.

This PR changes the behavior for the Playerstatus Panel and the scoreboard to use the players custom model if they have one set.

Gamemodes like VSH and ZI which were added officially would benefit from this change as they use custom models for the players.

playerstatus before:
grafik

playerstatus after:
grafik

scoreboard before:
grafik

scoreboard after:
grafik

### DIFF
--- a/src/game/client/tf/tf_hud_playerstatus.cpp
+++ b/src/game/client/tf/tf_hud_playerstatus.cpp
@@ -446,6 +446,7 @@ void CTFHudPlayerClass::UpdateModelPanel()
 		int nClass;
 		int nTeam;
 		int nItemSlot = m_nLoadoutPosition;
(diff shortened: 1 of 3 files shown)


gidi30:
I was not entirely happy with this pr as the model on the hud didn't instantly refresh when the players model changed unless another event that triggers a hud refresh happened (things like post_inventory_application or switching weapons).
I also went back and fixed some issues like the game crashing if the players model is set to an error model which it didn't before.

PR #930, opened 2025-03-02
[TF2] Prevent players with a lot of speed into a ramp from getting grounded

chrb22:
# Description
Sometimes when a player flies into a ramp with a lot of speed they get grounded instead of sliding up it. This is caused by the 2 unit check for ground under the player that happens every movement update. If the player's final position ends up within 2 units of the ramp they are trying to slide up, without actually colliding with it, they will get grounded and get brought down to their walking speed.

This fix predicts how much vertical velocity the player would have gotten from colliding with the ramp on the next movement update had they not been grounded. In the case their vertical velocity would have exceeded the required 250u/s, which unconditionally makes players airborne, they will not get grounded. This allows the player to collide with the ramp on the next movement update and properly slide up it, or get grounded if they for some reason lost all their speed in the mean time from e.g. knockback.

https://github.com/user-attachments/assets/8294053a-9584-419a-bfb9-595fc8ed641d

The fix first applies one full movement update of gravity to the player's current velocity (`FinishGravity` + `StartGravity`), then it gets a slide coefficient from `AirMove` (used for airblast debuff according to TF2 wiki), whereafter the velocity is finally projected onto the ramp with `ClipVelocity` to get a prediction.

There are already SourceMod addons that use the same trick to fix this and they have been in use on jumping and competitive community servers for quite a while without issues. I decided to use a ConVar because in gamemodes like bhop and surf it is a good thing if you get grounded on a ramp because a speed-preserving bhop gives you a smoother entry onto the ramp when you collide with it right after. I enabled the fix by default though, since that's what makes sense in normal gameplay.

Feel free to chime in if you have a better name/description for the ConVar.

### DIFF
--- a/src/game/shared/gamemovement.h
+++ b/src/game/shared/gamemovement.h
@@ -180,15 +180,6 @@ protected:
 	// Does not change the entities velocity at all
 	void			PushEntity( Vector& push, trace_t *pTrace );
 
-	// Slide off of the impacting object
-	// returns the blocked flags:
-	// 0x01 == floor
-	// 0x02 == step / wall
-	//
-	// redirectCoeff - multiplier on the clipped velocity to apply along their new direction. 0 -> stomp into-normal
-	//                 velocity, 1 -> redirect it all along new vector.
-	int				ClipVelocity( Vector& in, Vector& normal, Vector& out, float overbounce, float redirectCoeff = 0.f );
-
 	// If pmove.origin is in a solid position,
 	// try nudging slightly on all axis to
 	// allow for the cut precision of the net coordinates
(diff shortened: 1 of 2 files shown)


wgetJane:
looks good, though is this a tf2-only issue? would it make sense to have this be fixed in general?

chrb22:
> looks good, though is this a tf2-only issue? would it make sense to have this be fixed in general?

From my understanding this only really benefits TF2 as in other games you gain more from the control of being grounded or preserving/gaining speed by bhopping/ahopping.

tom-seewald:
Now that it has been ~8 months, what are the main blockers on this PR? Does it need additional testing or review?

a584-droid:
> Now that it has been ~8 months, what are the main blockers on this PR? Does it need additional testing or review?

Why would other players want this? Soldier legs are like gelatin or what? To collide with the surface instead of fall down? Or kind of…

PR #934, opened 2025-03-03
Allow gestures to loop backwards if the playback rate is negative

FlaminSarge:
# Description
If the gesture playback rate is negative, the logic here gets stuck reduce flCycle here below 0 indefinitely. Instead of that, reset it to 1.0 once it drops below 0, which in TF2's case allows looping taunts to also loop backwards.

### DIFF
--- a/src/game/shared/Multiplayer/multiplayer_animstate.cpp
+++ b/src/game/shared/Multiplayer/multiplayer_animstate.cpp
@@ -1287,6 +1287,11 @@ void CMultiPlayerAnimState::UpdateGestureLayer( CStudioHdr *pStudioHdr, GestureS
 	// Get the current cycle.
 	float flCycle = pGesture->m_pAnimLayer->m_flCycle;
 	flCycle += pPlayer->GetSequenceCycleRate( pStudioHdr, pGesture->m_pAnimLayer->m_nSequence ) * gpGlobals->frametime * GetGesturePlaybackRate() * pGesture->m_pAnimLayer->m_flPlaybackRate;
+	if ( flCycle < 0.0f )
+	{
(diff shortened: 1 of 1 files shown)


src/game/shared/Multiplayer/multiplayer_animstate.cpp
```diff
@@ -1287,6 +1287,11 @@ void CMultiPlayerAnimState::UpdateGestureLayer( CStudioHdr *pStudioHdr, GestureS
 	// Get the current cycle.
 	float flCycle = pGesture->m_pAnimLayer->m_flCycle;
 	flCycle += pPlayer->GetSequenceCycleRate( pStudioHdr, pGesture->m_pAnimLayer->m_nSequence ) * gpGlobals->frametime * GetGesturePlaybackRate() * pGesture->m_pAnimLayer->m_flPlaybackRate;
+	if ( flCycle < 0.0f )
```
FlaminSarge: If blanket-applying the negative cycle handling may cause issues, this can be further guarded by also checking that playbackRate specifically is negative before resetting cycle to 1.0.
```suggestion
	float flCycle = pGesture->m_pAnimLayer->m_flCycle;
	float flPlaybackRate = GetGesturePlaybackRate();
	flCycle += pPlayer->GetSequenceCycleRate( pStudioHdr, pGesture->m_pAnimLayer->m_nSequence ) * gpGlobals->frametime * flPlaybackRate * pGesture->m_pAnimLayer->m_flPlaybackRate;
	if ( flCycle < 0.0f && flPlaybackRate < 0.0f )
```

FlaminSarge:
I am going to attempt TF2-specific handling for non-looping taunts as well before I move this out of Draft.

PR #935, opened 2025-03-03
[TF2] Let player shoot on the same tick they canceled their reload

chrb22:
# Description
When a player tries to shoot a weapon with a non-empty clip while reloading in TF2, it cancels the reload on the first tick and only fires the weapon on the tick after. This happens because `ItemBusyFrame` (idle logic) and `ItemPostFrame` (shooting logic) are separated with an if/else statement controlled by `gpGlobals->curtime < m_flNextAttack`. Canceling a reload is effectively done by setting `m_flNextAttack = gpGlobals->curtime` inside `ItemBusyFrame`, so `ItemPostFrame` is skipped over even though `m_flNextAttack` marks the weapon as ready to shoot.

Changing the if/else to an if/if solves this. I'm pretty sure it shouldn't cause any problems as it only applies when `m_flNextAttack` is intentionally set to `gpGlobals->curtime` by weapon logic. However this is mostly a change directed at TF2, so I could specifically target TF2's reload cancel if need be.

### DIFF
--- a/src/game/shared/baseplayer_shared.cpp
+++ b/src/game/shared/baseplayer_shared.cpp
@@ -272,7 +272,8 @@ void CBasePlayer::ItemPostFrame()
 			GetActiveWeapon()->ItemBusyFrame();
 		}
 	}
-	else
+
+	if ( gpGlobals->curtime >= m_flNextAttack )
 	{
 		if ( GetActiveWeapon() && (!IsInAVehicle() || UsingStandardWeaponsInVehicle()) )
 		{


AzureWoof:
This definitely should be implemented. Most of the community uses auto reload, so I could imagine there being plenty of instances where players die the tick before they deal damage because they fired while reloading.

wgetJane:
> This definitely should be implemented. Most of the community uses auto reload, so I could imagine there being plenty of instances where players die the tick before they deal damage because they fired while reloading.

this is also very relevant for rocketjumping

misyltoad:
Thanks, this makes sense. I always wondered why it felt delayed, but never looked into it and thought it was just in my mind. :frog:

TorutheRedFox:
i didn't even realize that it was delayed but it definitely felt off and this explains why

PR #938, opened 2025-03-04
[TF2] Fix a few control point-related response criteria oversights

AndrewBetson:
# Description

Fixes oversights that cause players to call for help capping/defending control points that aren't actually cappable by either themselves or the enemy team, whether due to map progression or the match being in the "waiting for players" stage.

The latter case is fixed by ensuring the capture point the player is stood on is actually cappable by the enemy team, and the former by checking whether the point is locked and if it is set to allow our team to cap it.


henke37:
Does this affect the HUD? I seem to recall seeing weird hud status when standing on locked points.

AndrewBetson:
@henke37 No, these changes (should) only effect which response rules are used when players use the "Help!" voice command or fire their weapon while standing on a control point.

AndrewBetson:
Closing this while I revise with a better solution, will open new PR when that is done.

AndrewBetson:
Closing this while I revise with a better solution, will open new PR when that is done.

PR #942, opened 2025-03-04
Add ConVar to disable MvM-specific diguise on backstab nerfs.

Brain-dawg:
Also removes FVCAR_HIDDEN from the old buyback behavior cvars.

The disguise on backstab mechanic found on the Your Eternal Reward/Wanga Prick was severely kneecapped into being effectively useless very early into MvM's development, developers were concerned that it would be overpowered based on the provided comment.

Discussion of weapon/item balance is generally very thorny, however I will say: 

- This behavior has been removed with an external plugin for several years for Potato.TF's events and archive servers.
- Most people who have used it seem to agree that it is not overpowered, while also enabling a more unique playstyle where spy focuses on clearing up common bots/uber medics rather than taking down giants.  
- The downside of not being able to freely disguise (especially after the dead ringer nerf) is seemingly significant enough to keep it out of "overpowered" territory.  Especially when compared to other weapons the MvM community believes are genuinely overpowered.
- There is also the more subtle downside that bumping bots while stabbing them will still occasionally cause them to be alerted by your presence anyway, meaning it requires fairly careful positioning to pull off long chain-stabs.

Adding a cvar to disable this nerf would allow community servers to experiment with this weapon more, without relying on external plugins/extensions.


A1exKH (approved):
LGTM.

A1exKH (approved):
@Brain-dawg changes LGTM.

PR #944, opened 2025-03-04
fix being able to use firstperson when tiny

Ashetf2:
# Description

Fixes ValveSoftware/Source-1-Games#4425, adding a simple additional check to TF_COND_HALLOWEEN_TINY


HyperionCoding:
Why would that be a bug that needs a fix?

PR #945, opened 2025-03-05
Enhance and expose model glow effects to other games

xen-000:
Right now the model glow effect is exclusive to TF2 due to `GLOWS_ENABLE` being defined for that game only. On top of that, the glow effect was solely governed by the `tf_glow` entity, which is rather wasteful as one must created for each glowing model entity.

This PR adds glow properties and setters to `CBaseAnimating`, allowing any model entity to glow should the mapper choose. The `tf_glow` entity was also moved out of TF2 to be called `env_entity_glow` (in TF2 the `tf_glow` alias will still exist).

The originally intended glow halo effect was also added. The bloom strength was meant to be controlled by `glow_outline_effect_width` but the code was left in an incomplete state, making the cvar pointless. The old style, outline, is retained and is the default to maintain backwards compatibility. It can be controlled in entities and both styles can be used by different entities at once.

Outline
image

Halo
image

In addition, the alternative glow modes were fixed, now glows can be configured to only show when the model is occluded or visble.

https://github.com/user-attachments/assets/179d32d0-a97a-45ef-bb97-d77091dbef20

If this PR is to be merged, the added materials should be excluded and instead shipped with the games that will use them. Other games such as CS:S will need `GLOWS_ENABLE` added to their client/server vpc scripts. The alternative is to get rid of the define altogether but that can be up for discussion. Also, the `CBaseAnimating` and glow style additions are up for discussion as well but ultimately the commits can be excluded if it comes down to that.

The fgd entries for the aforementioned `CBaseAnimating` properties can be something like this under `@BaseClass = Studiomodel`:
```
	GlowEnabled(choices) : "Glow model" : 0 : "Whether this model should glow." =
	[
		0 : "No"
		1 : "Yes"
	]
	
	GlowColor(color255) : "FX Color (R G B)" : "255 255 255 255" : "The color of the glow effect (<r> <g> <b> <a> 0-255)"
	
	GlowMode(choices) : "Glow Visibility Mode" : 0 : "When the glow should be visible" =
	[
		0 : "Always"
		1 : "Only when model is occluded"
		2 : "Only when model is visible"
	]
	
	GlowStyle(choices) : "Glow Style" : 0 : "The glow style to use" =
	[
		0 : "Outline"
		1 : "Halo"
	]
	
	input EnableGlow(void) : "Enable glow on this model"
	input DisableGlow(void) : "Disable glow on this model"
	input SetGlowColor(color255) : "Set glow color (<r> <g> <b> <a> 0-255)"
	input SetGlowMode(integer) : "Set glow mode (0 = Always, 1 = Only when occluded, 2 = Only when visible)"
	input SetGlowStyle(integer) : "Set glow style (0 = Outline, 1 = Halo)"
```

`tf_glow` will need this in the FGD:
```
	Style(choices) : "Glow Style" : 0 : "The glow style to use" =
	[
		0 : "Outline"
		1 : "Halo"
	]
```


FlaminSarge:
Given that it's controlled by the VMTs, not sure how viable this is but: Would be useful to have a param (or maybe a client cvar) to control the glow's blur factor to determine how blurry/sharp it is.

Glow-within-radius would also be useful but potentially out of scope.

FlaminSarge:
I'd say preserving the old appearance/behavior is definitely still in scope, but I meant "only show glow while within certain radius of the object, or outside of certain radius of object" would be out of scope.

PKPenguin321:
+1 to retaining old behavior in addition to these changes

xen-000:
> +1 to retaining old behavior in addition to these changes

I added it back and made it the default for existing maps.

dreadcraftteam:
hl2_win64_y3VKZ9ngRg

How to fix it?

Bitl:
For whatever reason, this PR seems to not have the outline_add_to_screen.vmt included. I assume that the contents are

```
"screenspace_general"
{
   	"$PIXSHADER" "haloaddoutline_ps20"

    	"$ALPHA_BLEND_COLOR_OVERLAY" 1
    	"$basetexture" "_rt_FullFrameFB"
    	"$ignorez" 1
    	"$linearread_basetexture" 1
    	"$linearwrite" 1
}
```

based off reading the commit history.

EDIT: it also seems like halo_add_to_screen.vmt doesn't work, changing the PIXSHADER line to

```"$PIXSHADER" "haloaddoutline_ps20"```

seems to make it work properly on my end

PR #946, opened 2025-03-05
[TF2] Add Pyro kill animations

rowedahelicon:
# Description

Video demonstration: https://youtu.be/uVbb11_q6kw

This PR enables the [unused burning death animations found in every class except the Pyro](https://tcrf.net/Team_Fortress_2#Unused_Taunts_and_Animations). The process for doing so has been adjusted to work like other custom death animations, only being activated by specific fire based kills (_**TF_DMG_CUSTOM_BURNING,  TF_DMG_CUSTOM_BURNING_ARROW, TF_DMG_CUSTOM_BURNING_FLARE, TF_DMG_CUSTOM_FLARE_PELLET, TF_DMG_CUSTOM_PLASMA_CHARGED**_) rather than players simply being on fire at the time of death. 

It also modifies `C_TFRagdoll::FrameAdvance` to add the "turn to ash" particle effect at the end of the animation sequence to clean up the corpse in an appropriate way.

~~I know this is probably not going to get accepted, I like them and I wanted to at least try to get them back in~~

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -511,6 +511,7 @@ ConVar cl_ragdoll_physics_enable( "cl_ragdoll_physics_enable", "1", 0, "Enable/d
 ConVar cl_ragdoll_fade_time( "cl_ragdoll_fade_time", "15", FCVAR_CLIENTDLL );
 ConVar cl_ragdoll_forcefade( "cl_ragdoll_forcefade", "0", FCVAR_CLIENTDLL );
 ConVar cl_ragdoll_pronecheck_distance( "cl_ragdoll_pronecheck_distance", "64", FCVAR_GAMEDLL );
+ConVar cl_ragdoll_burn_anim_enable( "cl_ragdoll_burn_anim_enable", "0", 0, "Enable/disable burning death animations." );
 
 IMPLEMENT_CLIENTCLASS_DT_NOBASE( C_TFRagdoll, DT_TFRagdoll, CTFRagdoll )
 	RecvPropVector( RECVINFO(m_vecRagdollOrigin) ),
@@ -828,10 +829,18 @@ void C_TFRagdoll::CreateTFRagdoll()
 				iDeathSeq = -1;
 			}
 
-			//We want to end this particular animation with us becoming ash
+			// we want to end this particular animation with us becoming ash
 			if ( iDeathSeq > -1 && iDeathSeq == LookupSequence( "primary_death_burning" ) )
 			{
-				m_bBecomeAsh = 1;
+				// offer the option to disable these because they are a little silly
+				if ( !cl_ragdoll_burn_anim_enable.GetBool() )
+				{
+					iDeathSeq = -1;
+				}
+				else
+				{
+					m_bBecomeAsh = 1;
+				}
 			}
 		}
 	}
@@ -943,7 +952,7 @@ void C_TFRagdoll::CreateTFRagdoll()
 
 	if ( m_bBecomeAsh && !m_bDissolving && !m_bGib )
 	{		
-		//The death animation has a varying sequence at which it plays
+		// the death animation has a varying sequence at which it plays
 		if (iDeathSeq > -1 && iDeathSeq == LookupSequence( "primary_death_burning" ) )
 		{
 			m_flTimeToDissolve = SequenceDuration( LookupSequence( "primary_death_burning" ) );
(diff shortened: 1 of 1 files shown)


ficool2:
I suspect these were intentionally scrapped because they can be mistaken for alive players taunting (I believe TF2C had this problem in playtests and then disabled it by default)

Ashkon-JstarC:
Community contributions should purely be bug fixes/security fixes and nothing else. Easily a thumbs down.

rowedahelicon:
> I suspect these were intentionally scrapped because they can be mistaken for alive players taunting (I believe TF2C had this problem in playtests and then disabled it by default)

Potentially, I've had a plugin on my server that utilizes them and most people don't even seem to notice them at all, so I figured why not try and see, plus it was fun

timeimp:
Making this a server-config item that admins could manage, you'd get the best of both worlds.

rowedahelicon:
> Making this a server-config item that admins could manage, you'd get the best of both worlds.

Better yet, I've added a client side cvar!

CombineSlayer24:
> I suspect these were intentionally scrapped because they can be mistaken for alive players taunting (I believe TF2C had this problem in playtests and then disabled it by default)

I don't see how animations can confuse players. There's backstab animations. Just shoot them to make sure they are actually alive.

CombineSlayer24:
> > Making this a server-config item that admins could manage, you'd get the best of both worlds.
> 
> Better yet, I've added a client side cvar!

There you go, best of both worlds.

FlaminSarge:
> I know this is probably not going to get accepted, I like them and I wanted to at least try to get them back in

Useful for community forks

bakedhorse:
> > I suspect these were intentionally scrapped because they can be mistaken for alive players taunting (I believe TF2C had this problem in playtests and then disabled it by default)
> 
> I don't see how animations can confuse players. There's backstab animations. Just shoot them to make sure they are actually dead.

The problem is that spy can only stab so many people at once, while pyro can burn a whole group of people and even kill at once. Creating a cluster of dead players doing animations.
The backstab animations are really short and are a good indicator of "Hey a spy just stabbed that sniper, theres a spy around here".
These burned ones are inconsistent in both their length and quality, some of them are long and others short. They're missing face flexes, Heavy shows a perfect example of this with him looking up with a blank face when he supposedly should be screaming.
They don't really indicate anything either, they just show "Hey theres a pyro" which you should already know from seeing people on fire.

However if this was to ever be added, having a cvar toggle or even toggle in the settings menu itself would be ideal. As a PR that forks/mods can use, that's perfectly fine for those creators to use it for their needs.

Bitl:
Made a version of this with the ability to disable the ash effect for community mods

https://github.com/BitlDevelopmentStudios/source-sdk-2013-bds-base/commit/b03b90bae80c1dcf45ea226c21b53f863504da06

x8c8r:
> Community contributions should purely be bug fixes/security fixes and nothing else. Easily a thumbs down.

While I see your point, I believe that the mindset of "community should only fix" is pretty bad.

I agree that this PR isn't something that should be added, but not all additions are equal. For example someone might want to add a toggle for a thing that bothers people (no example in mind, so just imagine), would that be bad? 

I believe that additions should be limited to Quality of Life features rather than none at all, although I wouls bet in the future we may even see exceptions to this. At the end of the day it is really just not up to us to decide whether things like this should be merged. (At least I doubt we have a big enough say)

0x5066:
at the end of the day it's up to valve whether or not to merge something like this, if e.g. people here like the idea enough then it probably gives them reason enough to consider pushing this in whatever next update they have in store, and if it doesnt make it in, other mods can benefit from these changes

condor00fr:
I had always had the assuption that theses animations where meant for flare crits only, seeing as headshots and stabbed animations soley plays when crit killed it makes sense.

ewillerup:
It should be off by default but toggleable with a client-side command like `tf_fire_death_animations 1`, similar to how `tf_taunt_first_person 1` is available for players that want to try the experience.

src/game/client/tf/c_tf_player.cpp
```diff
@@ -936,9 +951,17 @@ void C_TFRagdoll::CreateTFRagdoll()
...
 
 	if ( m_bBecomeAsh && !m_bDissolving && !m_bGib )
-	{
-		ParticleProp()->Create( "drg_fiery_death", PATTACH_ABSORIGIN_FOLLOW );
-		m_flTimeToDissolve = 0.5f;
+	{		
+		// the death animation has a varying sequence at which it plays
+		if (iDeathSeq > -1 && iDeathSeq == LookupSequence( "primary_death_burning" ) )
+		{
+			m_flTimeToDissolve = SequenceDuration( LookupSequence( "primary_death_burning" ) );
```
ZzZombo, line 958: Avoid repeated calls like these.

destroyer087:
Closes https://github.com/ValveSoftware/Source-1-Games/issues/6236

PR #947, opened 2025-03-05
Adjusted `CTFPlayerDestructionLogic::EvaluatePlayerCount` to add a po…

rowedahelicon:
…int above the top scoring team in the event that players disconnecting adjust scores below what the new winning threshold would be.

# Description
This is an alternative solution to the problem brought up by https://github.com/ValveSoftware/source-sdk-2013/pull/864 using a simpler approach suggested by FlaminSarge. 

It adjusts the `m_nMaxPoints` value to be 1 point higher than the winning / tied team in the event that an adjusted max point score is below or equal to the scores of the current round.


A1exKH (approved):
LGTM.

PR #953, opened 2025-03-06
[HL2DM] Fix physcannon and physics issues

speedvoltage:
This PR fixes some issues with the physcannon and physics objects.

Fixes in this PR:

- Moved `phys_swap` client-side, while still letting the server receive proper weapon switch information. I fail to understand why this was set exclusively server-side.
- Fixed being able to switch from the physcannon to another weapon while holding an object. 
- Fixed NULL pointers in relation to held physics props. 
- Fixed an issue where dying while holding an object in the physcannon can cause that object to desynchronize between the client and server. 
- Prevent objects from breaking/exploding after yo-yoing them. 
- Fixed props losing collisions for the player holding said object if switching to Spectators. 
- Fixed an issue where an orb would levitate in the air if a player disconnected while holding an orb in their physcannon. 
- Update `ForceDropOfCarriedPhysObjects` filter.
- Fixed an issue where players could punt invisible items at their spawning location even after being picked up. 
- Fixed some physics NULL pointers.

### DIFF
--- a/src/game/client/weapon_selection.cpp
+++ b/src/game/client/weapon_selection.cpp
@@ -400,6 +400,38 @@ void CBaseHudWeaponSelection::UserCmd_Slot10(void)
 	SelectSlot( 10 );
 }
 
+void ClientInstantPhysSwap()
+{
+	C_BasePlayer *pPlayer = C_BasePlayer::GetLocalPlayer();
+	if ( !pPlayer )
+		return;
+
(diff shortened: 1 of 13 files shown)


misyltoad:
This is a lot of changes that need to be split up to be properly reviewed.

speedvoltage:
> This is a lot of changes that need to be split up to be properly reviewed.

I completely understand. A lot of those affect the physcannon, but I can definitely split this into multiple PRs for easier review. Thanks for the comment.

PR #954, opened 2025-03-06
[HL2DM] Move .357 snap after firing to client-side

speedvoltage:
This PR addresses an issue where firing the .357 causes the player’s view angles to "snap" back to the firing angles due to the server processing this. This becomes particularly problematic at higher pings, where firing the weapon while moving your view results in your view snapping back to the firing angles after this many milliseconds corresponding to the player's latency.

Some server ops resolve this by disabling the view angle snapping entirely on the server, which does prevent the issue. However, this also removes the intended randomization in where the crosshair ends up after the weapon’s disorienting recoil effect, reducing the intended gameplay feedback.

The fix here moves the view angle snapping logic to the client-side, preventing the annoying snapping caused by the server processing the shots.

### DIFF
--- a/src/game/shared/hl2mp/weapon_357.cpp
+++ b/src/game/shared/hl2mp/weapon_357.cpp
@@ -11,6 +11,7 @@
 
 #ifdef CLIENT_DLL
 	#include "c_hl2mp_player.h"
+	#include <prediction.h>
 #else
 	#include "hl2mp_player.h"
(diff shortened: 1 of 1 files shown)


misyltoad:
Yah. Makes sense to just make this clientside and predict once given how viewpunch and the SetViewAngles differ here. Thanks.

PR #957, opened 2025-03-06
[HL2DM] Fix shotgun issues, and improve trace logic

speedvoltage:
This PR aims at fixing long standing shotgun issues.

- Fixed shotgun trace logic, using all 3 trace methods for the most accurate detection. 
    The default trace logic uses half trace lines, half trace hulls. The problem with this is that you end up with a shotgun that is either way too OP or too weak. With the proposed changes here, all three traces are used for maximum precision.

- Fixed secondary shotgun prediction bug.
    Looks like secondary attack was missing for prediction.

- Fixed shotgun infinite reload loop bug.
    The reload variable did not seem networked, so you could end up with a shotgun being stuck in an infinite reload loop animation.

- Fixed shotgun locking up when holding reload and attack keys simultaneously. 
    If you held the attack and reload keys simultaneously, your weapon ends up in an in-between state trying to do both, but is stuck.

- Set reload animation for the player model.
    If the player reloaded, the player model would not show any reload animation, making it impossible for other players to see a player reloading a shotgun.

### DIFF
--- a/src/game/server/hl2mp/hl2mp_player.cpp
+++ b/src/game/server/hl2mp/hl2mp_player.cpp
@@ -684,7 +684,7 @@ bool CHL2MP_Player::WantsLagCompensationOnEntity( const CBasePlayer *pPlayer, co
 {
 	// No need to lag compensate at all if we're not attacking in this command and
 	// we haven't attacked recently.
-	if ( !( pCmd->buttons & IN_ATTACK ) && (pCmd->command_number - m_iLastWeaponFireUsercmd > 5) )
+	if ( !( ( pCmd->buttons & IN_ATTACK ) || ( pCmd->buttons & IN_ATTACK2 ) ) && ( pCmd->command_number - m_iLastWeaponFireUsercmd > 5 ) )
 		return false;
 
 	// If this entity hasn't been transmitted to us and acked, then don't bother lag compensating it.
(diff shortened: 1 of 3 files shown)


speedvoltage:
This is absolutely right. When we tested this with other players in-game, a lot of players felt the shotgun was a lot better than before. Some thought it was a little "too much better." In my pre-20th anniversary repo, I had the cone adjusted to 15 degrees to compensate, but I did not think of applying it again. So I updated the shotgun to have a 15-degree cone radius to compensate the hitreg changes, which should balance things out.

I also fixed the `#ifdef` guards a bit, because I might have affected other games inadvertently. The changes should only apply to HL2DM.

Thanks for the feedback!

Sirbucket:
This is probably an ok solution since it greatly improves the actual registration for players with poor connection or when the server takes a fat dump. Only thing you should be wary about is overloading the trace engine if someone makes a weapon that fires stupid fast or something like that. I don't think its a realistic problem but could happen. Also Im not sure what the third type of trace does but traceline and tracehull should be more than enough. If Im not understanding something then that would be nice to know.

speedvoltage:
> Also Im not sure what the third type of trace does but traceline and tracehull should be more than enough. If Im not understanding something then that would be nice to know.

Of course. You are correct on the fact that a trace line and trace hull would be sufficient most of the times and that, realistically, the trace ray usage would be redundant. That said, the trace ray adds some robustness and handles some edge cases, especially when handling more complex surfaces, props, and player interactions. The trace ray is going to shine best when you want to ensure pellets register correctly across all surface types, such as world geometry, props, displacements, and handle scenarios like narrow gaps between props or irregular collision geometry. Is this an absolute necessity? I guess this is open to discussion. Since I come from the competitive world in HL2DM, every shot counts and needs to be accurate.

In short, you can absolutely do fine with just trace line and hull.

PR #959, opened 2025-03-06
[HL2DM] Fixed ladder bugs

speedvoltage:
This PR is to fix ladder issues.

- Fixed a bug where players could respawn on ladders. 
    Players could spawn or respawn on ladders during mp_restartgame or suicide for example.

- Fixed an issue where players disconnecting while on a ladder would cause it to become inoperable. 
    Disconnecting when mounting a ladder would cause it to break.

- Movement fixes.
    Properly move in the direction you are looking.

- Fixed an exploit where some players could use climb ladders at insane speeds. 
    Notorious speedrun trick among speedrunners, you could, through manual inputs or scripts/macros, climb ladders at a very fast pace.

- Fixed an issue where players could dismount ladders and go through walls.
    Minor issue since most maps do not have this setup, but if two ladders are on two sides of the same wall, then players could clip through walls to the dismount node.

### DIFF
--- a/src/game/server/hl2mp/hl2mp_player.cpp
+++ b/src/game/server/hl2mp/hl2mp_player.cpp
@@ -166,6 +166,12 @@ CHL2MP_Player::~CHL2MP_Player( void )
 
 void CHL2MP_Player::UpdateOnRemove( void )
 {
+	if ( auto p = GetLadderMove() )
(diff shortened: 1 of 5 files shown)


A1exKH (approved):
LGTM.

PR #961, opened 2025-03-07
[HL2DM] Disable client-side blood impacts and blood decals

speedvoltage:
This PR is really super simple. Discussed in #829, this moves all the client-side blood to the server's side to provide accurate hit information. The main problem with it being client-side in the first place is that a blood impacts may not guarantee a hit. By having be server-sided, if blood is shown, a hit is confirmed.

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -918,6 +918,8 @@ void CBasePlayer::DrawDebugGeometryOverlays(void)
 //=========================================================
 void CBasePlayer::TraceAttack( const CTakeDamageInfo &inputInfo, const Vector &vecDir, trace_t *ptr, CDmgAccumulator *pAccumulator )
 {
+	CDisablePredictionFiltering disabler;
+
 	if ( m_takedamage )
 	{
 		CTakeDamageInfo info = inputInfo;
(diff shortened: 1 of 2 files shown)


src/game/server/player.cpp
```diff
@@ -918,6 +918,8 @@ void CBasePlayer::DrawDebugGeometryOverlays(void)
 //=========================================================
 void CBasePlayer::TraceAttack( const CTakeDamageInfo &inputInfo, const Vector &vecDir, trace_t *ptr, CDmgAccumulator *pAccumulator )
 {
+	CDisablePredictionFiltering disabler;
```
wgetJane: i think this should be moved down below to where the TraceBleed call is and also should only be for hl2mp:

```cpp
		{
#ifdef HL2MP
			CDisablePredictionFiltering disabler;
#endif
			SpawnBlood(ptr->endpos, vecDir, BloodColor(), info.GetDamage());// a little surface blood.
			TraceBleed( info.GetDamage(), vecDir, ptr, info.GetDamageType() );
		}

speedvoltage: Yes, that makes more sense indeed. Made the changes.

PR #969, opened 2025-03-08
[TF2] Fix thirdperson weapon shell ejection

NachosChipeados:
Fixes weapon shell ejection coming out of the wrong spot when in thirdperson, by using the fix already used by the minigun's brass particle.

Before:

https://github.com/user-attachments/assets/ffed78d4-3b4a-476f-8784-e3005bc26eb8

After:

https://github.com/user-attachments/assets/eac09fac-6e0d-49b7-aa8d-5c7e0699e38c

### DIFF
--- a/src/game/shared/tf/tf_weaponbase.cpp
+++ b/src/game/shared/tf/tf_weaponbase.cpp
@@ -331,6 +331,7 @@ CTFWeaponBase::CTFWeaponBase()
 	
 #ifdef CLIENT_DLL
 	m_iCachedModelIndex = 0;
+	m_hEjectBrassWeapon = NULL;
 	m_iEjectBrassAttachpoint = -2;
 
 	m_bInitViewmodelOffset = false;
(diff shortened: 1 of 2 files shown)


src/game/shared/tf/tf_weaponbase.cpp
```diff
@@ -4836,6 +4837,10 @@ bool CTFWeaponBase::OnFireEvent( C_BaseViewModel *pViewModel, const Vector& orig
 {
 	if ( event == 6002 && ShouldEjectBrass() )
 	{
+		m_hEjectBrassWeapon = GetWeaponForEffect();
```
misyltoad: is there a reason to add a member for this? It might as well just be a local var with a fallback to pViewModel

NachosChipeados: I did it this way because thats [what the minigun also does](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/game/shared/tf/tf_weapon_minigun.cpp#L1205), but i can definitely change it if needed

TheNotoriousFrogposter:
When you spectate other players in third person, their guns don't eject any shotgun shells at all. Just my two cents, but shell ejection shouldn't be happening in third person at all.

PR #970, opened 2025-03-08
VScript: VectorAngles and VectorAnglesReference

BlockBuilder57:
# Description
Exposes the commonly-used VectorAngles function to VScript.

This allows for a QAngle to be created from a direction vector. This can either be relative to the global up direction, or to a reference up (exposed by VectorAnglesReference through the C++ `VectorAngles( Vector&, Vector&, QAngle& )` overload.)

### DIFF
--- a/src/game/server/vscript_server.cpp
+++ b/src/game/server/vscript_server.cpp
@@ -1208,6 +1208,20 @@ QAngle RotateOrientation( QAngle posAngles, QAngle entAngles )
 	return outAngles;
 }
 
+QAngle ScriptVectorAngles( const Vector forward )
+{
+	QAngle outAngles;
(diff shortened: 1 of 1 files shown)


ficool2:
This should be added to the Vector class, not as a global method

ficool2:
> I would agree (if that code was public) were it not for RotatePosition/RotateOrientation already being present in vscript_server.cpp. There isn't a relevant instance for math functions like there is for Convars and NetProps. I could close this PR and create another to add many of the Vector, QAngle, and Quaternion mathlib functions to such an instance, if you believe that to be a better course of action?

I would loosely follow existing conventions because there's plenty of historical oversights/poor design decisions in the API, having it directly part of the instances would be better

PR #973, opened 2025-03-09
[TF2/HL2MP] Fix broken sprays for client indices greater than 64

AndrewBetson:
# Description
Caused partially by `m_nPlayer` network var being converted to bit stream using only 6 bits (`log2(101) = 6(.658)`),
and partially by 36 missing playerlogo client effect materials.

Fixed by manually setting the number of bits needed to convert `m_nPlayer` (`PLAYER_INDEX_BITS`, defined in `shareddefs.h` in case there are other parts of the codebase that could benefit from it) and by defining the missing 36 playerlogo client effect materials.

If merged into live TF2/HL2:DM, the missing playerlogo VMTs also need to be added to the `hl2` mod folder; conveniently provided in the following tarball:

[PlayerLogo65Through100.tar.gz](https://github.com/user-attachments/files/19149344/PlayerLogo65Through100.tar.gz)

Fixes https://github.com/ValveSoftware/Source-1-Games/issues/5266


src/game/client/c_te_playerdecal.cpp
```diff
@@ -105,6 +105,44 @@ CLIENTEFFECT_MATERIAL( "decals/playerlogo61" )
 CLIENTEFFECT_MATERIAL( "decals/playerlogo62" )
 CLIENTEFFECT_MATERIAL( "decals/playerlogo63" )
 CLIENTEFFECT_MATERIAL( "decals/playerlogo64" )
+#if defined( TF_CLIENT_DLL ) || defined( HL2MP )
```
ficool2: This should be changed to `#if MAX_PLAYERS > 64`

src/game/server/te_playerdecal.cpp
```diff
@@ -94,7 +94,7 @@ void CTEPlayerDecal::Test( const Vector& current_origin, const QAngle& current_a
 IMPLEMENT_SERVERCLASS_ST(CTEPlayerDecal, DT_TEPlayerDecal)
 	SendPropVector( SENDINFO(m_vecOrigin), -1, SPROP_COORD),
 	SendPropInt( SENDINFO(m_nEntity), MAX_EDICT_BITS, SPROP_UNSIGNED ),
-	SendPropInt( SENDINFO(m_nPlayer), Q_log2( MAX_PLAYERS ), SPROP_UNSIGNED ),
+	SendPropInt( SENDINFO(m_nPlayer), PLAYER_INDEX_BITS, SPROP_UNSIGNED ),
```
ficool2: You don't need `PLAYER_INDEX_BITS`, you can change it to `Q_log2( MAX_PLAYERS ) + 1` (this matches how other places handle this problem), for example see `m_iWeaponID` in `CTEFireBullets`. 

Alternatively, `static_cast<int>( ceil( log2( MAX_PLAYERS ) ) )`

ficool2 (changes requested):

src/game/client/c_te_playerdecal.cpp
AndrewBetson: Changed this to `#if MAX_PLAYERS > 65` so CStrike doesn't get caught by it.

src/game/server/te_playerdecal.cpp
AndrewBetson: This is something I hadn't actually thought to do, and even though it *technically* means the 33 and 65 `MAX_PLAYERS` cases will use 1 bit more than is necessary, it *is* more consistent with existing code. Thank you for the suggestion!

goodusername123:
At what point would it make sense to rewrite the playerlogo VMTs with the [patch shader](https://developer.valvesoftware.com/wiki/Patch)? all of them have identical parameters and there's so many of them. If this where to be done then making a new "playerlogo_shared" VMT would probably make the most sense although just pointing at "playerlogo1.vmt" would work as well I guess.

src/game/server/te_playerdecal.cpp
```diff
@@ -94,7 +94,7 @@ void CTEPlayerDecal::Test( const Vector& current_origin, const QAngle& current_a
 IMPLEMENT_SERVERCLASS_ST(CTEPlayerDecal, DT_TEPlayerDecal)
 	SendPropVector( SENDINFO(m_vecOrigin), -1, SPROP_COORD),
 	SendPropInt( SENDINFO(m_nEntity), MAX_EDICT_BITS, SPROP_UNSIGNED ),
-	SendPropInt( SENDINFO(m_nPlayer), Q_log2( MAX_PLAYERS ), SPROP_UNSIGNED ),
+	SendPropInt( SENDINFO(m_nPlayer), Q_log2( MAX_PLAYERS )+1, SPROP_UNSIGNED),
```
Margen67, line 97:
```suggestion
	SendPropInt( SENDINFO(m_nPlayer), Q_log2( MAX_PLAYERS )+1, SPROP_UNSIGNED ),
```

PR #974, opened 2025-03-09
[TF2] Custom kill icon support

Lizard-Of-Oz:
### Description
This PR allows mappers to define custom kill icons by packing a new file called `scripts/map_textures.txt` that's structured identical to the game's `scripts/mod_textures.txt` where the stock kill icons are defined.
This PR also reloads kill icons on LevelInit.

Custom materials/textures are fully supported (unless different maps happened to store non-identical textures under the same name, which is a broader issue with asset caching anyway).

There's several ways for mappers to make use of this feature:
- This PR adds a keyvalue `killicon` to `trigger_hurt`, `func_croc` and `point_hurt`. When not empty, its value will be used as the kill icon name.
- By default, the game uses the inflictor's `classname` keyvalue for the kill icon name. Using VScript, you can use OnTakeDamage hook to assign a dummy entity as the inflictor:
```js
myCustomInflictor <- SpawnEntityFromTable("info_target", { classname = "my_custom_kill_icon" });
<...>
OnScriptHook_OnTakeDamage = function(params)
{
  if (/* insert condition */)
    params.inflictor = myCustomInflictor;
}
```
- It's possible to change `classname` of an existing entity with a map output `MyEntityName | AddOutput | classname my_custom_kill_icon` or with a VScript call `myEntity.KeyValueFromString("classname", "my_custom_kill_icon")`. However, the side-effects of this change are unclear and I recommend the other 2 methods.

### Considerations

- All changes to the kill icon loading procedure are contained within hud.cpp, in case if you want to merge only the icon loading functionality without adding the `killicon` keyvalue.
- While I've tested compatibility with `sv_pure` to the best of my ability, the sv_pure system works in mysterious ways.
- I've benchmarked the time it takes to reload the kill icons - it's negligible.
- An alternative path for the file could be `maps/%bspname%_map_textures.txt`, but keeping the file name in sync with the bsp name is very annoying to deal with during map development, and there is no need for this naming convention here, as the file appears to get reloaded between maps correctly.
- This PR unintentionally makes it possible to pack `scripts/mod_textures.txt` that will override the stock file until the map change. In fact, it's _already_ possible in Live TF2, but it requires the user to use `hud_reloadscheme` command for changes to take effect.
I see this as non-issue, as if a bad actor wants to mess with players with a poisoned map, there's many other opportunities to do so.
If you disagree, this can be fixed with the following change to hud.cpp that will force the game to load `hud_textures` and `mod_textures` from the _game's_ files.
```diff
-	LoadHudTextures( textureList, "scripts/hud_textures", NULL );
-	LoadHudTextures( textureList, "scripts/mod_textures", NULL );
+	const char* dudKey = "123";
+	LoadHudTextures( textureList, "scripts/hud_textures", (const unsigned char*) dudKey );
+	LoadHudTextures( textureList, "scripts/mod_textures", (const unsigned char*) dudKey );
```

### DIFF
--- a/src/game/client/hud.cpp
+++ b/src/game/client/hud.cpp
@@ -533,6 +533,8 @@ void CHud::LevelInit( void )
 		group->bHidden = false;
 		group->m_pLockingElements.Purge();
 	}
+
+	RefreshHudTextures();
 }
 
 //-----------------------------------------------------------------------------
@@ -811,6 +813,14 @@ void CHud::RefreshHudTextures()
 
 	CUtlDict< CHudTexture *, int >	textureList;
 
+	// loading custom kill icons, presumably packed into the map
+	LoadHudTextures( textureList, "scripts/map_textures", NULL );
+	int co = textureList.Count();
+	for ( int index = 0; index < co; index++ )
+	{
+		CHudTexture* tex = textureList[ index ];
+		AddSearchableHudIconToList( *tex );
+	}
 	// check to see if we have sprites for this res; if not, step down
 	LoadHudTextures( textureList, "scripts/hud_textures", NULL );
 	LoadHudTextures( textureList, "scripts/mod_textures", NULL );
--- a/src/game/server/pointhurt.cpp
+++ b/src/game/server/pointhurt.cpp
@@ -22,6 +22,9 @@ BEGIN_DATADESC( CPointHurt )
 	DEFINE_KEYFIELD( m_flDelay, FIELD_FLOAT, "DamageDelay" ),
 	DEFINE_KEYFIELD( m_bitsDamageType, FIELD_INTEGER, "DamageType" ),
 	DEFINE_KEYFIELD( m_strTarget, FIELD_STRING, "DamageTarget" ),
+#ifdef TF_DLL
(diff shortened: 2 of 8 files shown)


Lizard-Of-Oz:
> An alternative path for the file could be
> scripts/map_textures.txt

I've been given a better direction on how the game should choose a file from which to load the kill icon sheet. Will update the PR shortly.

Lizard-Of-Oz:
The "better direction" just committed adds `SetCustomKillIconsFile` input to `tf_gamerules` entity, built identical to already existing `SetCustomUpgradesFile` input.

The reason behind this is that if you're developing several maps with custom assets in parallel, having several files under the same name would make development more difficult.

The mappers are now expected to give their kill icon sheet file a unique name and make an input `tf_gamerules | SetCustomKillIconsFile | pathToMyKillIconSheet`. No extra steps are needed to load the materials of the freshly loaded kill icons. The game will do so automatically.

Once the input is fired, the clients are instructed to reload their kill icons (reload time is negligible) and to load the custom kill icons sheet provided by the input.

The gamerules entity also keeps the file path as a NetProp, so that the newly joined clients will be informed of the custom kill icons file upon loading.

PR #975, opened 2025-03-09
x64 and VS2022 support for the Singleplayer branch.

SrcEngGuy:
# Description
The is self-explainable. I don't have to write anything here, except there are a few broken includes. This is currently a WiP pull request.


Blixibon:
Wouldn't this require 64-bit binaries in the singleplayer SDK base?

Bitl:
> Wouldn't the x64 support require the singleplayer SDK base to have x64 engine binaries as well?
no, as the multiplayer base itself supports singleplayer code. iirc before the tf2 sdk update, some singleplayer mods used the multiplayer branch as it was a newer branch at the time.

Blixibon:
> no, as the multiplayer base itself supports singleplayer code. iirc before the tf2 sdk update, some singleplayer mods used the multiplayer branch as it was a newer branch at the time.

If HL2 mods can already be made with the MP branch, then what is this pull request for?

SrcEngGuy:
> Wouldn't the x64 support require the singleplayer SDK base to have x64 engine binaries as well?

SinglePlayer branch also supports be played on the MP branch.

SrcEngGuy:
> > no, as the multiplayer base itself supports singleplayer code. iirc before the tf2 sdk update, some singleplayer mods used the multiplayer branch as it was a newer branch at the time.
> 
> If x64 HL2 mods can already be made with the MP branch, and the only option to support them is with MP binaries, then why should these changes be ported to the SP branch in the first place?

I believe VALVe will update Half-Life 2 and its episodic releases with support for x64, and if they do not, this pull request was also made for VS2022 support.

PR #977, opened 2025-03-09
Botvision

Brain-dawg:
The VScript API has a few functions for getting various bot interfaces, however only GetLocomotionInterface was actually correctly exposed, and ILocomotion is the only class that exists in the root table with any existing functions.  This means GetVisionInterface, GetIntentionInterface, and GetBodyInterface are currently useless.

While it would be nice to have all of the other interfaces exposed, IVision is, in my opinion, the most valuable one to have.

I've also attached an example vscript file to test all of the vision interface functions:

- save file to tf/scripts/vscripts and change the extension from .txt to .nut
- load any mvm map (mvm_decoy has no delay before bots start spawning)
- uncomment a bot.GetScriptScope().ThreatThink function, change values as necessary to test
- `script_execute visiontest` in console
- start the wave (can skip the 10s timer by doing ```ent_fire "tf_gamerules" "runscriptcode" "NetProps.SetPropFloat(self, `m_flRestartRoundTime`, 1)"``` )
- `tf_mvm_jump_to_wave 1; ent_fire "player" "TerminateScriptScope"` to reset and kill the think function on the bots
- comment out the confirmed working think function and uncomment another one
- repeat

[visiontest.txt](https://github.com/user-attachments/files/19152122/visiontest.txt)

### DIFF
--- a/src/game/server/NextBot/NextBotVisionInterface.h
+++ b/src/game/server/NextBot/NextBotVisionInterface.h
@@ -11,7 +11,21 @@
 
 class IBody;
 class INextBotEntityFilter;
+class IKnownEntity;
 
+inline HSCRIPT ToHScript(const CKnownEntity *pKnownEntity)
+{
+	if (!pKnownEntity)
+		return NULL;
+	
+	// Return the underlying entity instead of the knowledge wrapper
+	return ToHScript(pKnownEntity->GetEntity());
+}
+
+inline CKnownEntity *ToKnownEntity( HSCRIPT hScript )
+{
+	return NULL;
+}
 
 //----------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/game/server/NextBot/NextBotVisionInterface.h
```diff
@@ -103,11 +136,16 @@ class IVision : public INextBotComponent
 	 * A visible spot on the subject is returned in 'visibleSpot'.
 	 */
 	virtual bool IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *visibleSpot = NULL ) const;
-
+
```
FlaminSarge: nit: remove this whitespace char

PR #979, opened 2025-03-09
All the headers of .h and .cpp are in size

Unusuario1:
# Description
This changes the header of all the .h and .cpp files to be in size with the bottom of the header.
before:
image
now:
image


kisak-valve:
Hello @Unusuario1, this isn't worth the comment churn.

PR #980, opened 2025-03-10
[VRAD] Added support "-extrapasses"

Unusuario2:
# Description
Lets the user scale how many extra passes you want your map to go through, makes the lightmaps less sharp
Default (-extrapasses 4):
image

Custom (-extrapasses 16):
image


SirYodaJedi:
Does this actually control supersampling? VRAD refers to supersampling as "extra", because that's what QRAD called it for some reason.

Unusuario2:
> Does this actually control supersampling? VRAD refers to supersampling as "extra", because that's what QRAD called it for some reason.

Yes

PR #982, opened 2025-03-10
Fix players entering enemy spawnrooms with partner taunts

Ashetf2:
# Description
ValveSoftware/Source-1-Games#3881

This fix adds a check if the player that initiates the taunt is inside their respawnroom, and if the round hasn't ended yet, so that players can still taunt with enemies when they lose.

(Ignore the tc2 logo)

https://github.com/user-attachments/assets/9ec27dbc-bf1a-4e20-aee0-c595077e60fb

https://github.com/user-attachments/assets/9bb3ec44-50df-4bc9-85ae-693bcfc00945

Of course, the other way doesn't make sense, and thus is allowed

https://github.com/user-attachments/assets/8cd9b7a5-1665-4bfb-bf13-acefe825fa73

Also, i think this would make the current fix, prevent if there is a respawnroomvisualizare between, useless, but we better keep it just in case

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -21567,6 +21567,18 @@ CTFPlayer *CTFPlayer::FindPartnerTauntInitiator( void )
 			continue;
 		}
 
+		// Check if: 1. the initiator is inside a respawnroom and 2. the round has not ended, 
+		// to not execute when players can already enter the enemy respawnroom.
+		// This should be an easy way to fix players entering spawnrooms with partner taunts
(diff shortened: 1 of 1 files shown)


wgetJane:
does this only work on enemies or also teammates?

Ashe-tf:
> does this only work on enemies or also teammates?

Thanks, that's fixed now

https://github.com/user-attachments/assets/aff42e72-c284-4e8a-959c-b7b71a46c723

https://github.com/user-attachments/assets/b0f8ddb1-7239-49f1-bbde-e87ad65c2b42

PR #983, opened 2025-03-10
[All] Add bicubic lightmap sampling support to VertexLitGeneric

AndrewBetson:
# Description
This PR adds support for bicubic lightmap sampling to VertexLitGeneric's lightmapped (non-bump/non-Phong) path.

Below are some demonstration screenshots taken on the TF2 map `rd_asteroid`.
Admittedly this is the only official map I could find with substantial usage of lightmapped props -- fitting given that the feature was developed specifically *for* this map.

Disabled             |  Enabled
  | 
 | 
 |


Unusuario2:
Have you tested if this new feature works on lightmapped props?
https://developer.valvesoftware.com/wiki/$lightmap

WhiteRedDragons:
Lets hope people actually have margins on their props..
Or they get the same issue we had before the SDK Update, where they hacked in a 1 pixel outline for Brush Lightmaps..

SirYodaJedi:
> Lets hope people actually have margins on their props.. Or they get the same issue we had before the SDK Update, where they hacked in a 1 pixel outline for Brush Lightmaps..

Wonder if it would make sense to have bicubic prop lightmaps use a separate cvar from bicubic brush lightmaps, in that case. Honestly, I wish it was a VMT boolean.

PR #988, opened 2025-03-11
Implement IsNextBot()

Bitl:
# Description
This implements a boolean in baseentity.h that functions similarly to IsNPC or IsPlayer. When referencing a NextBot NPC or Player, the IsNextBot() value will return true.

This was originally made for my HL2 NPC awareness of NextBot feature to my personal fork and my FIREFIGHT RELOADED mod (https://github.com/BitlDevelopmentStudios/source-sdk-2013-bds-base/commit/14091c39328fd5ee02ec8967881c026157727573), but I decided to send a PR of IsNextBot() and not the senses implementation as I felt that the feature would only suit specific mods that have working NPCs in multiplayer or incorporating NextBot NPCs into singleplayer. IsNextBot() is general enough that I see it getting implemented in an SDK branch.

### DIFF
--- a/src/game/server/NextBot/NextBot.h
+++ b/src/game/server/NextBot/NextBot.h
@@ -40,6 +40,7 @@ public:
 	virtual Vector EyePosition( void );
 
 	virtual INextBot *MyNextBotPointer( void ) { return this; }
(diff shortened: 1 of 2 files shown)


A1exKH (approved):
LGTM.

PR #991, opened 2025-03-11
[TF2] Add extra killbinds

RealGamerX:
This PR allows players to killbind in a lot more ways than just the standard 'kill' and 'explode'. 

https://github.com/user-attachments/assets/faf70684-837f-4514-8287-e1864247607a

The new variants and their console commands are:
1. tf_burn
2. tf_freeze
3. tf_incinerate
4. tf_disintegrate
5. tf_disintegrate_and_gib
6. tf_mistify
7. tf_midas_touch


RealGamerX:
> I like these but i feel like the tf_ prefix should be removed in favor of making these sound more generic, much like the "kill" or "explode" commands just do they fit in with those
> 
> just a personal thing though

These are tf-specific commands, so I felt that they needed a tf-specific prefix, to stay in line with most other console commands.

Bitl:
> > I like these but i feel like the tf_ prefix should be removed in favor of making these sound more generic, much like the "kill" or "explode" commands just do they fit in with those
> > just a personal thing though
> 
> These are tf-specific commands, so I felt that they needed a tf-specific prefix, to stay in line with most other console commands.

fair

Hedwig7s:
Would love, although I feel like these could get combined into a single command
Like tf_suicide [variant] (defaults to kill)

Bitl:
> Worth adding a cvar to allow server operators to disable this feature (or leave it disabled by default and allow server operators to enable it).
> 
> Also agreed that these should all be one command with a parameter instead of 7 commands.

here's one i just drafted up.

`//------------------------------------------------------------------------------
// kill commands with custom ragdolls
//------------------------------------------------------------------------------
CON_COMMAND(tf_kill, "Kills the player in multiple entertaining ways. If none is selected, the normal kill behavior will be used.\nList:\n0 = Normal\n1 = Burn\n2 = Freeze\n3 = Incinerate\n4 = Disintegrate\n5 = Disintegrate and Gib\n6 = Mistify\n7 = Turn to Gold")
{
	int iRagdollType = atoi(args[1]);
	const char* cPlayerName = args[2];

	if (args.ArgC() > 2 && sv_cheats->GetBool())
	{
		// Find the matching netname
		for (int i = 1; i <= gpGlobals->maxClients; i++)
		{
			CTFPlayer* pPlayer = ToTFPlayer(UTIL_PlayerByIndex(i));
			if (pPlayer)
			{
				if (Q_strstr(pPlayer->GetPlayerName(), cPlayerName))
				{
					pPlayer->CommitSuicideWithCustomRagdoll(iRagdollType);
				}
			}
		}
	}
	else if (args.ArgC() > 1)
	{
		CTFPlayer* pPlayer = ToTFPlayer(UTIL_GetCommandClient());
		if (pPlayer)
		{
			pPlayer->CommitSuicideWithCustomRagdoll(iRagdollType);
		}
	}
	else
	{
		CTFPlayer* pPlayer = ToTFPlayer(UTIL_GetCommandClient());
		if (pPlayer)
		{
			pPlayer->CommitSuicideWithCustomRagdoll(TF_KILLBIND_CUSTOM_RAGDOLL_TYPE_NONE);
		}
	}
}`

Hedwig7s:
I could see this being problematic considering when you see certain ragdolls you expect certain classes
Might not be a bad idea having a client side toggle although I'm not sure how much extra complexity that would add

FlaminSarge:
> I mean the same is true of `explode`, if someone explodes in the game I'd generally expect their death to come from either a demo or soldier

I think they're mainly talking about the lingering ragdolls for ice, specifically (other ones to a lesser extent).

drHyperion451:
I really don't think it's a good idea to name the command "suicide" if kill is already an option. Maybe adding some arguments like "kill ice", or "kill smite" will be okay (and also keep "explode" while having "kill explode").

Hedwig7s:
> I really don't think it's a good idea to name the command "suicide" if kill is already an option. Maybe adding some arguments like "kill ice", or "kill smite" will be okay (and also keep "explode" while having "kill explode").

I don't think changing base source commands is the best way
Maybe tf_kill [variant] [target] where target only works on servers

JoriKos:
I agree that a ConVar to toggle them on/off for servers should exist. The statues and the mistify one should be disabled on Casual servers due to them being indicators of a specific weapon, for example. Coming across a frozen player can indicate a Spy nearby, or the mistify indicating a Classic Sniper ~~(though they're not much of a threat)~~ is nearby.

The rest, imo, are not a gameplay issue because 99% of the time you'd see the player die that way and not see any enemies nearby capable of causing any of these deaths. The Classic death specifically can be an issue because Snipers can't necessarily be seen all the time, but I do think there is an argument to be made that it will largely be used in meme interactions that will be obvious it's not a Classic death.

RealGamerX:
- Added a server cvar to toggle the custom ragdolls (thanks @FlaminSarge ).
- Combined all the commands into one (thanks @Bitl ).
- Added 2 extra killbind variants:
1. Decapitate (Hitman's Heatmaker headshot);
2. Cloak (Your Eternal Reward backstab).

https://github.com/user-attachments/assets/6a3cd049-3c32-468b-ada7-541093ada5b0

Hedwig7s:
The availability would be rather poor with a sourcemod plugin
The impact of ragdolls on casual play is minimal so it'd probably be fine to have it in casual, although it can definitely be a bit confusing
Then again most pros play with ragdolls off so they can't be that relevant

KaelaSavia:
should be disabled by default in casual/comp and togglable by convar instead

last thing we need is even more confusing gameplay

blackletum:
RealGamerX, I do not have the skill to do so, but I was thinking you could add another effect to this.

This PR references the unused burning animations that are included in the models:

https://github.com/ValveSoftware/source-sdk-2013/pull/946

This could be added as another variant!

Hedwig7s:
> should be disabled by default in casual/comp and togglable by convar instead
> 
> last thing we need is even more confusing gameplay
> 

That would kind of defeat the entire point of having it in the base game
This definitely could be recreated easily with a sourcemod plugin so community only really defeats the point and even then there's the debate of on or off by default on community

TrenchcoatCraig:
I don't particularly like tf_freeze and tf_midas_touch, I feel they reduce the clarity of the game state. They both accomplish a similar goal of SDing while turning the merc's into a solid material. Disintegrate and gib is also redundant with disintegrate. 

If tf_freeze and tf_midas_touch were replaced with a single kill bind that turned the merc into a unique material like stone, it would be more clear at a glance that this is a player who kill binded, and not a player who was killed by a Spycicle or Golden Pan/Saxxy. 

Also, the cloak one is especially bad for game state clarity as Your Eternal Reward backstabs don't show up in the kill feed anyway, raising false alarms for the presence of a spy if a player who is unaware of the kill binds witnesses it. A regular backstab kill bind I think would be more fitting.

Hedwig7s:
To be fair, midas touch just indicates australium (no more informative than a normal ragdoll) although freeze is a bit riskier

TrenchcoatCraig:
> To be fair, midas touch just indicates australium (no more informative than a normal ragdoll) although freeze is a bit riskier

That's not quite accurate. midas touch indicated 'Golden' weapons, not Australium. Golden weapons would be the Golden Frying Pan, the Saxxy, and the Golden Wrench. While Australium weapons have unique kill icons in the kill feed, the regdolls/gibus's produced match those of their non Australium counterparts.

Hedwig7s:
@RealGamerX Any explanation?

RealGamerX:
> @RealGamerX Any explanation?

accidentally renamed the branch this pull request was on, which closed this PR. Made a new PR with the now renamed branch.

loltestyas-stack:
how install this killbind?

PR #992, opened 2025-03-11
[FIX/ALL] Ambiguous operator invocation compile-time error

Mooshua:
# Description

<details>
  <summary>Original Description</summary>
  
> On more recent versions of C++, this line throws an ambiguous operator error since it can't determine if it should be using the CUtlSymbol equality operator or the built-in int equality operator.
> 
> This patch removes the ambiguity by explicitly stating that the CUtlSymbol equality operator should be used. This doesn't change the underlying behavior, it just fixes the ambiguous resolution error.
> 
> This may also be done in another open PR, I didn't check. Sorry in advance if that's the case :)
</details>

On more recent versions of C++, the compiler isn't able to resolve some of the ambiguities left in the original codebase. This PR aims to fix up some of the more glaring ones without actually changing the behavior of the code.

Many of these could probably be better fixed with new operators, but again to avoid changes to the code and *just* make the darn thing compile, I'm just adding clarity where it's needed.

Patches:
- Ambiguous Operator Resolution in `CUtlSymbolTable::HasElement`
- Ambiguous Symbol Resolution in `CUtlSortVector::QuickSort`
- Ambiguous Operator Resolution in `CParticleCollection::IsValid`

### DIFF
--- a/src/public/particles/particles.h
+++ b/src/public/particles/particles.h
@@ -2434,7 +2434,7 @@ FORCEINLINE Vector CParticleCollection::GetControlPointVelocityAtCurrentTime( in
 
 FORCEINLINE bool CParticleCollection::IsValid( void ) const 
 { 
-	return ( m_pDef != NULL && m_pDef->GetMaterial() );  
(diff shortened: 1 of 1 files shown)


eepycats (approved):
lgtm!!

PR #995, opened 2025-03-12
Fix constant redefinition errors

Deathreus:
# Description
When trying to compile a mod I ran into this error, I wasn't sure on the best approach to get rid of it as idk what does and doesn't include t0constants and fileio


Bitl:
reverted it from my fork because although this seems good i'm not sure if this could potentially break anything

Mooshua:
This is caused by t0constants being included after fileio. There's a lot of include ordering that's required but not documented.

Deathreus:
Nothing in the SDK includes fileio, but that doesn't mean mods didn't as I've shown, and it didn't even need to be included really, but someone else could run into this error, and really I think the constants can be deleted because t0constants *should* always be included, but the ifdef is a "just-in-case"

PR #996, opened 2025-03-12
[TF2] Fix the Third Degree's inconsistent behavior with Amputator healing

Tkain:
# Description

The Third Degree's current behavior with regard to players being healed by the Amputator taunt is inconsistent. Currently, as the TF2 wiki notes, when a Medic performs the Amputator taunt around a group of players:

- Attacking the taunting Medic with the Third Degree will damage the Medic only.
- Attacking a healing player with the Third Degree will damage the player and the taunting Medic only.

This PR fixes this inconsistency by not damaging the Amputator-taunting Medic when attacking a healing player with the Third Degree. This is done by not marking a healer as 'connected' to another player the healer's healing target is not also the player, which is only the case when they are being AoE healed.

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -10048,19 +10048,11 @@ void CTFPlayer::AddConnectedPlayers( CUtlVector<CTFPlayer*> &vecPlayers, CTFPlay
 
 	for ( int i = 0 ; i < pPlayerToConsider->m_Shared.GetNumHealers() ; i++ )
 	{
+		// Make sure the Medic is healing us specifically - don't count AoE healing
 		CTFPlayer *pMedic = ToTFPlayer( pPlayerToConsider->m_Shared.GetHealerByIndex( i ) );
-		if ( pMedic )
+		if ( pMedic && pMedic->MedicGetHealTarget() == pPlayerToConsider )
 		{
-			// Make sure the healer is healing us with a Medigun - don't count AoE healing
-			CTFWeaponBase *pWpn = pMedic->GetActiveTFWeapon();
-			if ( pWpn && pWpn->GetWeaponID() == TF_WEAPON_MEDIGUN )
-			{
-				CWeaponMedigun *pMedigun = assert_cast< CWeaponMedigun* >( pWpn );
-				if ( pMedigun->GetHealTarget() == pPlayerToConsider )
-				{
-					AddConnectedPlayers( vecPlayers, pMedic );
-				}
-			}
+			AddConnectedPlayers( vecPlayers, pMedic );
 		}
 	}
 }


FlaminSarge:
> `flOverhealBonus <= 1.0f` (i.e. that the healer is not providing overheal)

While this works out for most cases, if a mod makes a medigun that does not provide any overheal, will the Third Degree no longer function for that medigun?

Sirbucket:
Can probably be done with a specific check for the amputator instead (or we create a custom healing type for aoe heals to not count?)

HyperionCoding:
The check should be against `TF_COND_RADIUSHEAL_ON_DAMAGE` instead as that is specifically the amputator taunt effect.

wgetJane:
what's the inconsistent behaviour? isn't the third degree supposed to damage the medic through the patient?

the third degree not damaging patients through the medic i would assume is probably intentional to avoid accidentally griefing your own team since it's a big aoe

Tkain:
> The check should be against `TF_COND_RADIUSHEAL_ON_DAMAGE` instead as that is specifically the amputator taunt effect.
> 
> > Automatically adds condition 20 and 21 and causes the player and every nearby teammate to begin gaining health as if being healed by the [Amputator](https://wiki.teamfortress.com/wiki/Amputator). The player gets the credit for any healing that occurs. Conditions 20 and 21 are automatically removed when the player ends any taunt, but condition 55 remains active (though it does nothing after this point).

Good to know. I'll see if I can check that condition instead. If I can, I'll probably replace the other `flOverhealBonus <= 1.0f` check with it since there's probably no harm in doing so (correct me if I'm wrong!).

> what's the inconsistent behaviour? isn't the third degree supposed to damage the medic through the patient?
> 
> the third degree not damaging patients through the medic i would assume is probably intentional to avoid accidentally griefing your own team since it's a big aoe

The inconsistent behavior is specifically that the Third Degree will damage the Medic if any of their patients are hit but will not damage any patients if the Medic is hit. I assumed that this behavior was unintentional for three reasons:

- The `AddConnectedPlayers` function doesn't make any reference to AoE healing or the player list you'd need to check to get the players someone is currently AoE healing (`pPlayerToConsider->m_Shared.m_iRadiusHealTargets`), so I assumed AoE healing was not considered when the function was written.
- The Third Degree's description says that players connected by Medigun _beams_ are also hit. The Amputator taunt's healing has no beam.
- It's listed as a bug on the TF2 Wiki, although admittedly that could have been added by anybody.

Tkain:
Very quick update on `TF_COND_RADIUSHEAL_ON_DAMAGE`: if that's a player condition, then that only tells us that a player is being healed by AoE healing - not whether a specific healing source is AoE-healing them, which is what we need to know whether to count the healer as 'connected' to the player or not.

I think the only way to check for AoE healing more robustly is to add a new boolean field to `healers_t` denoting whether a healing source is AoE healing or not. That might be better suited for a different PR.

FlaminSarge:
... Would looking through mediguns for m_hHealingTarget not work for the Third Degree check?

FlaminSarge:
Nice, that medigun check works out really well for this.

Tkain:
I implemented the check as you recommended before realizing we actually only needed to do something much simpler - to check if a player is being AoE healed, we only need to check that their healer's "healing target" isn't the player, which only occurs under AoE healing. (Under direct healing, the Medic's healing target is set to the player.)

I hope this covers everything because I don't want to keep poking at this really tiny bug fix.

FlaminSarge:
> we only need to check that their healer's "healing target" isn't the player, which only occurs under AoE healing. (Under direct healing, the Medic's healing target is set to the player.)

One more consideration for the sake of community mods: MedicGetHealTarget() only functions if the healer is specifically a medic. Some mods grant mediguns to non-medics; under the existing code, those players also get hit by the Third Degree, but with your change, they would not. To that end, the GetActiveWeapon -> medigun.m_bHealing check may preserve behavior for those mods.

Tkain:
I guess, but there are some other usages of `MedicGetHealTarget()` that cause similar issues when giving Mediguns to non-Medics. Whoever wants to make that change in their mod will need to edit `MedicGetHealTarget()` anyway, so I don't think it's worth adding the extra complexity to make this one specific feature work beforehand.

FlaminSarge:
> I guess, but there are some other usages of `MedicGetHealTarget()` that cause similar issues when giving Mediguns to non-Medics. Whoever wants to make that change in their mod will need to edit `MedicGetHealTarget()` anyway, so I don't think it's worth adding the extra complexity to make this one specific feature work beforehand.

I meant more for stock TF2 + SourceMod, but fair.

PR #1000, opened 2025-03-13
[TF2] Update Default FOV from 75 -> 90 for freshly generated configurations.

Sirbucket:
# Description
Self explanatory but most new player guides almost immediately suggest changing the fov from 75 -> 90 as it is a direct benefit to do so.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -502,11 +502,3 @@ FodyWeavers.xsd
 
 # JetBrains Rider
 *.sln.iml
-
-#Linux files from build outputs
-*.a
-
-game/mod_hl2mp_linux64
-game/mod_tf_linux64
-src/.ninja_deps
-src/.ninja_log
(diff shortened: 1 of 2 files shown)


HyperionCoding:
I think this should be a part of larger default setting updates if one were to happen.

Sirbucket:
Its just added as a bonus because if I dont update the gitignore I'd have to remove the excess files before I commit.

blackletum:
> Its just added as a bonus because if I dont update the gitignore I'd have to remove the excess files before I commit.

When you commit, you can choose which files are contained within the push. You do not need to change .gitignore for this.

Bloxxel64:
Nice, though i do agree that a new default update is probably overdue, this is a good start tho:

Sirbucket:
I think I'll make a larger default setting update at some point in the future, for now Im going to close this as I forgot to make a new branch and it would create a messy git history with any other prs I do.

PR #1011, opened 2025-03-14
[HL2MP] Spectator observer fixes

speedvoltage:
This PR is to tackle a few observer issues when being a spectator:

- Fixed an issue where spectators could spectate a player who joined team Spectator. 
    -> If you spectate a player in IN_EYE mode mainly, then you end up spectating someone who is now a spectator. It is a funny and harmless bug, though. Now, any time a player switches to team Spectator and you are in IN_EYE mode, you will be switched back to free roam observer mode.
- Switch to free roaming when joining team spectator.
    -> For some reason, you are stuck in place when switching to spectators. It looks like you should be in free roam mode, but you are not or is not set properly.
- Switch to free roaming when player leaves the game.
    -> Another harmless bug. If you spectate a player in IN_EYE mode amd then leaves, you are still stuck in that last observer mode. Now, any time a player leaves, it switches you back to free roam mode.
- Removed a useless second third-person view.
    -> No idea why this needed to be a thing. I guess it is for SourceTV mainly, but one of the observer modes is identical to third-person mode, which means you have to press space bar twice to get back to free roam mode from first-person view. The fix is to simply remove that extra observer mode outside of SourceTV.

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -2287,7 +2287,7 @@ bool CBasePlayer::StartObserverMode(int mode)
 	
     AddSolidFlags( FSOLID_NOT_SOLID );
 
-	SetObserverMode( mode );
+	SetObserverMode( OBS_MODE_ROAMING );
 
 	if ( gpGlobals->eLoadType != MapLoad_Background )
 	{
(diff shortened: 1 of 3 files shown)


Bitl:
One recent issue I noticed is that the change to bool CBasePlayer::StartObserverMode(int mode) breaks the TF2 freezecam mode. 

The code there should be the following as a result:

`#if HL2MP
	SetObserverMode(OBS_MODE_ROAMING);
#else
	SetObserverMode(mode);
#endif`

PR #1012, opened 2025-03-14
[HL2MP] Fix spectator exploits

speedvoltage:
**Issue**: 
By default, it is possible to unfairly kill people when attacking them and switching to spectators quickly. 

**Example**: If TDM is enabled and you throw a grenade to a team mate and switch to spectators before that grenade blows, then that grenade will hurt or kill what used to be your team mate. This is because the game thinks you are no longer in the same team, and logically, you are not anymore. 

**Fix**: 
The fix is to treat all damage done by anyone in spectators as no damage. This fix also applies if TDM is disabled.

### DIFF
--- a/src/game/server/hl2/func_tank.cpp
+++ b/src/game/server/hl2/func_tank.cpp
@@ -1450,6 +1450,9 @@ void CFuncTank::Use( CBaseEntity *pActivator, CBaseEntity *pCaller, USE_TYPE use
 	if ( !pPlayer )
 		return;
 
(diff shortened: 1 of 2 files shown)


wgetJane:
why not check the projectile's team instead of the attacker's team?

Bitl:
> why not check the projectile's team instead of the attacker's team?

if i recall, in HL2MP projectiles aren't assigned to a team, so its owner's team has to be checked.

PR #1014, opened 2025-03-14
[HL2MP] Fix stuck sticking props collision

speedvoltage:
**Issue**: 
Stuck props like saw blades have no collisions anymore.

**Fix**: 
Simply set collisions to `COLLISION_GROUP_INTERACTIVE`.


wgetJane:
this is for shooting the sawblade/harpoon/etc props with the gravity gun into a wall right? isn't that intentional?

teufortressIndustries:
> this is for shooting the sawblade/harpoon/etc props with the gravity gun into a wall right? isn't that intentional?

this toggles the collision of the stuck prop after the prop hit the wall

PR #1018, opened 2025-03-14
[HL2MP] Turn off flashlight on ConVar update

speedvoltage:
**Issue**: 
When `mp_flashlight` is set to 0, players who already have their flashlights on can keep them on until they manually turn them off.

**Fix**: 
Automatically disable all active flashlights when `mp_flashlight` is set to 0.

### DIFF
--- a/src/game/server/game.cpp
+++ b/src/game/server/game.cpp
@@ -24,6 +24,28 @@ void MapCycleFileChangedCallback( IConVar *var, const char *pOldString, float fl
 	}
 }
 
+void flashlight_changed( IConVar *pConVar, const char *pOldString, float flOldValue )
+{
+	ConVarRef var( pConVar );
+	if ( !var.IsValid() )
+	{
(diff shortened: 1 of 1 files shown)


src/game/server/game.cpp
```diff
@@ -24,6 +24,28 @@ void MapCycleFileChangedCallback( IConVar *var, const char *pOldString, float fl
...
 
+void flashlight_changed( IConVar *pConVar, const char *pOldString, float flOldValue )
+{
+	ConVarRef var( pConVar );
+	if ( !var.IsValid() )
+	{
+		return;
+	}
+
+	if ( var.GetInt() < 1 )
```
ficool2: Should be == 0 to match `CMultiplayRules::FAllowFlashlight` behavior, -1 will enable flashlights but your code will disable them

ficool2 (changes requested):

PR #1026, opened 2025-03-14
[HL2MP] Fix RPG Rocket Damage

speedvoltage:
**Issue**: 
Rockets blast radius can affect other players located above a ceiling or on the other side of a wall when they should not be damaged.

**Fix**: 
Adjust the origin on the Z axis to prevent unfair damage caused by rockets.


wgetJane:
why not use WorldSpaceCenter() as the centre of the explosion?

Bitl:
explosionOrigin doesn't seem to be used anywhere in the function. shouldn't it replace GetAbsOrigin() in the explosion function?

PR #1035, opened 2025-03-14
[HL2MP] Fix trigger_push not lifting players off the ground

speedvoltage:
**Issue**: 
When a `trigger_push` is set to push players upwards, they often remain stuck to the ground and are not properly lifted when in motion (moving about). The original 1-unit Z-axis push is insufficient to "unglue" moving players from the floor.

**Fix**: 
Change the 1-unit Z push to 16. It might be overkill, but it at least ensures that players get pushed upwards when inside an active volume.


ficool2:
This will breaks maps that expect the old behavior. I can remember maps which had a corridor where you could proceed ahead or optionally choose to jump to use the trigger push instead

Bitl:
I'd advise maybe adding a convar (that's off by default) that enables or disables this behavior

PR #1046, opened 2025-03-14
[HL2MP] Make info_null and infodecal server-side

speedvoltage:
**Issue**: 
`info_null` and `infodecal` are not server-sided.

**Fix**: 
Make them server-sided.

### DIFF
--- a/src/game/server/subs.cpp
+++ b/src/game/server/subs.cpp
@@ -22,10 +22,10 @@ void CPointEntity::Spawn( void )
 }
 
 
-class CNullEntity : public CBaseEntity
+class CNullEntity : public CServerOnlyEntity
 {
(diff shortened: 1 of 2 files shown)


henke37:
That is not an issue. That is stating a fact without explaining anything.

Bitl:
Partial duplicate of https://github.com/ValveSoftware/source-sdk-2013/pull/910?

SirYodaJedi:
How does this affect named decals? How does the client know where to apply the decals?

SirYodaJedi:
> How does this affect named decals? How does the client know where to apply the decals?

It seems decals are networked using a different system from the edict system, so yes, it is safe to de-edict them.

PR #1047, opened 2025-03-14
[HL2MP] Fix game_ui interacting with other entities

speedvoltage:
**Issue**: 
When a player activates a `game_ui`, they are still able to interact with other interactive objects, such as buttons, doors, etc., which can be abused in some cases.

**Fix**: 
Whenever a player activates a `game_ui`, +USE will exclusively deactivate the `game_ui` if the flag is set in Hammer.

### DIFF
--- a/src/game/server/game_ui.cpp
+++ b/src/game/server/game_ui.cpp
@@ -43,6 +43,8 @@ public:
 	void Deactivate( CBaseEntity *pActivator );
 
 	float				m_flFieldOfView;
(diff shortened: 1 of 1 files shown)


ficool2:
This breaks compatibility with maps that expect players to +use a button while using game_ui, including when "use deactivates" flag is on

Bitl:
To keep the original behavior, maybe you could use a keyvalue that toggles this behavior?

PR #1049, opened 2025-03-14
[HL2MP] Fix a crash with game_ui

speedvoltage:
**Issue**: 
When a player disconnects when controlling a `game_ui`, the server or game crashes.

**Fix**: 
Prevent the warning from showing up and causing crashes.


ficool2:
Did you actually test this? This crash was fixed when the `pActivator` nullptr check was introduced recently

PR #1052, opened 2025-03-14
[HL2MP] Fix "Start locked" on func_rot_button

speedvoltage:
**Issue**: `
Start locked` on `func_rot_button` doesn't work because it is missing a flag check.

**Fix**: 
Make it work xD


Bitl:
Duplicate of https://github.com/ValveSoftware/source-sdk-2013/pull/401?

PR #1055, opened 2025-03-14
[HL2MP] Fix an exploit with env_info

speedvoltage:
**Issue**: 
If a player attempts to use `ent_info worldspawn`, their game will crash.

**Fix**: 
Prevent the usage of `ent_info worldspawn`.


ficool2:
The cause of this crash should be fixed rather than disabling it..

Bitl:
looks good to merge now

PR #1057, opened 2025-03-14
[HL2MP] Allow god mode

speedvoltage:
**Issue**: 
Why is the `god` command not available in MP?

**Fix**: 
Allow Jesus to take over the game.


src/game/server/client.cpp
```diff
@@ -1237,9 +1237,6 @@ void CC_God_f (void)
 	   if ( gpGlobals->deathmatch )
```
ficool2: Remove the check for TF2 too (I'm not sure why these checks exist there too... since `buddha` is allowed)

PR #1059, opened 2025-03-14
[HL2MP] Add missing physbox kill death notice icon

speedvoltage:
**Issue**: 
A physbox kill shows the suicide skull in the death notice.

**Fix**: 
Use the physics icon when a kill with physbox is made.


Bitl:
Alternatively this could be done in mod_textures.txt

PR #1063, opened 2025-03-14
[HL2MP] Prevent deleting vital entities

speedvoltage:
**Issue**: 
Deleting certain entities crashes the game. Obviously, deleting `worldspawn` is a bad idea for example.

**Fix**: 
Prevent deleting vital entities responsible for running the game properly.


ficool2:
This is a poor way of going about it. You can bypass it by changing the classname of the entity, for example `ent_fire player addoutput "classname abc";ent_remove abc"` (you can also this with ent_create)

PR #1065, opened 2025-03-14
[HL2MP] Death fall black screen removal

speedvoltage:
**Issue**: 
When a player die by fall damage, the screen becomes black.

**Fix**: 
Remove the black fade, because it is unnecessary.

### DIFF
--- a/src/game/server/movehelper_server.cpp
+++ b/src/game/server/movehelper_server.cpp
@@ -390,16 +390,6 @@ bool CMoveHelperServer::PlayerFallingDamage( void )
 
     }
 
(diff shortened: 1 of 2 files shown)


Bitl:
Why not move

```
virtual bool  FlPlayerFallDeathDoesScreenFade( CBasePlayer *pl ) { return false; }
```

to hl2mp_gamerules.h rather than do it in the base gamerules class? The removal of code in movehelper_server.cpp is probably not needed either as it already checks if FlPlayerFallDeathDoesScreenFade is enabled.

PR #1072, opened 2025-03-14
"Ignoring unreasonable position" console spam fix

speedvoltage:
**Issue**: 
The log was often cluttered with the message "Ignoring unreasonable position" because the engine was picking up on unrealistic entity positions. This happened due to floating-point precision issues or incorrect calculations that led to coordinates being way too large or invalid. These repetitive log entries were spamming the console.

**Fix**: 
Implemented a stronger validation check to make sure that position values stay within a sensible range before they trigger an error log. This cuts down on the number of irrelevant warnings and makes sure that only genuinely problematic entity positions get flagged, which enhances the clarity of debugging.

### DIFF
--- a/src/game/server/baseentity.cpp
+++ b/src/game/server/baseentity.cpp
@@ -6749,7 +6749,21 @@ void CBaseEntity::SetLocalOrigin( const Vector& origin )
 	}
 }
 
-void CBaseEntity::SetLocalAngles( const QAngle& angles )
+inline float FloatNormalize_floorf_inline( float angle )
+{
+	angle = angle - 360.0f * floorf( angle / 360.0f + 0.5f );
+	if ( angle > 180.0f )
+	{
+		angle -= 360.0f;
+	}
+	else if ( angle < -180.0f )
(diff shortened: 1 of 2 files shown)


src/game/server/baseentity.cpp
```diff
@@ -6759,21 +6773,27 @@ void CBaseEntity::SetLocalAngles( const QAngle& angles )
 	//        handling things like +/-180 degrees properly. This should be revisited.
 	//QAngle angleNormalize( AngleNormalize( angles.x ), AngleNormalize( angles.y ), AngleNormalize( angles.z ) );
 
+
+#ifdef HL2MP
+	QAngle normalizedAngles( FloatNormalize_floorf_inline( angles.x ), FloatNormalize_floorf_inline( angles.y ), FloatNormalize_floorf_inline( angles.z ) );
```
ficool2, line 6777: Why make this exclusive to HL2DM?

ficool2 (changes requested):

PR #1088, opened 2025-03-14
[HL2MP] Add view kick for melee weapons

speedvoltage:
**Issue**:
None of the melee has view kick. The screen therefore remains static while the attack is performed.

**Fix**:
Apply the `AddViewKick()` function to the crowbar and the stunstick.


ficool2:
Given this was commented out originally, this was probably intentionally disabled as I can imagine it could be disorienting in the fast pacing of multiplayer

speedvoltage:
Fair enough. We can add a cvar for this and disable the view kick by default.

PR #1093, opened 2025-03-14
[HL2MP] Clamp max player position and angle values

speedvoltage:
**Issue**:
Players may end up being placed out of bounds or having invalid rotations, oftentimes caused by teleportation glitches, map exploits, or physics errors.

**Fix**:
Clamp player position and angle values to ensure they remain within a valid and reasonable range.

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -592,6 +592,7 @@ CBasePlayer::CBasePlayer( )
 
 	m_bForceOrigin = false;
 	m_hVehicle = NULL;
+	m_hUseEntity = NULL;
 	m_pCurrentCommand = NULL;
 	m_iLockViewanglesTickNumber = 0;
 	m_qangLockViewangles.Init();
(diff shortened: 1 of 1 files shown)


src/game/server/player.cpp
```diff
@@ -2631,6 +2632,17 @@ void CBasePlayer::ObserverUse( bool bIsPressed )
 
 void CBasePlayer::JumptoPosition(const Vector &origin, const QAngle &angles)
 {
+	Vector neworigin;
+	QAngle newangles;
+
+	// Clamp the position and angles to prevent crashes
+	neworigin.x = clamp( origin.x, MIN_COORD_FLOAT, MAX_COORD_FLOAT );
```
Mooshua, line 2639: These lines don't do anything, the only case where this validation is run is when switching spectator targets

src/game/server/player.cpp
```diff
@@ -2631,6 +2632,17 @@ void CBasePlayer::ObserverUse( bool bIsPressed )
...
+	neworigin.y = clamp( origin.y, MIN_COORD_FLOAT, MAX_COORD_FLOAT );
+	neworigin.z = clamp( origin.z, MIN_COORD_FLOAT, MAX_COORD_FLOAT );
+	newangles.x = clamp( angles.x, MIN_COORD_FLOAT, MAX_COORD_FLOAT );
```
Mooshua, line 2642: COORD_FLOAT? Not sure that's right

src/game/server/player.cpp
```diff
@@ -2631,6 +2632,17 @@ void CBasePlayer::ObserverUse( bool bIsPressed )
...
+	newangles.y = clamp( angles.y, MIN_COORD_FLOAT, MAX_COORD_FLOAT );
+	newangles.z = clamp( angles.z, MIN_COORD_FLOAT, MAX_COORD_FLOAT );
+
 	SetAbsOrigin( origin );
 	SetAbsVelocity( vec3_origin );	// stop movement
 	SetLocalAngles( angles );
```
Mooshua, line 2648: Checks against flMaxEntityEulerAngle, should change that instead

Mooshua (changes requested):
Not only does this PR not change anything or fix any bugs, it doesn't even address the actual problematic parts of the codebase.

PR #1095, opened 2025-03-14
[HL2MP] Fix ammo crate exploit

speedvoltage:
**Issue**:  
If a player tosses a grenade or any explosive at an `item_ammo_crate` and then switch to a crowbar before the explosion happens, the crate will open up, giving ammo to the player who triggered it, thus creating an unfair advantage. Plus, only the crowbar can open these crates, which means the stunstick can’t do anything with them.

**Fix**:  
Implemented a check to ensure that damage is only registered from a crowbar and have also made the stunstick capable of opening this type of crate.

### DIFF
--- a/src/game/server/hl2/item_ammo.cpp
+++ b/src/game/server/hl2/item_ammo.cpp
@@ -877,12 +877,15 @@ void CItem_AmmoCrate::Use( CBaseEntity *pActivator, CBaseEntity *pCaller, USE_TY
 	if ( pPlayer == NULL )
 		return;
 
+	if ( !pPlayer->IsAlive() )
+		return;
+
 	m_OnUsed.FireOutput( pActivator, this );
 
 	int iSequence = LookupSequence( "Open" );
 
 	// See if we're not opening already
-	if ( GetSequence() != iSequence )
+	if ( GetSequence() != iSequence && pPlayer->IsAlive() )
 	{
 		Vector mins, maxs;
 		trace_t tr;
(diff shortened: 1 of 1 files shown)


src/game/server/hl2/item_ammo.cpp
```diff
@@ -877,12 +877,15 @@ void CItem_AmmoCrate::Use( CBaseEntity *pActivator, CBaseEntity *pCaller, USE_TY
...
+	if ( !pPlayer->IsAlive() )
+		return;
+
 	m_OnUsed.FireOutput( pActivator, this );
 
 	int iSequence = LookupSequence( "Open" );
 
 	// See if we're not opening already
-	if ( GetSequence() != iSequence )
+	if ( GetSequence() != iSequence && pPlayer->IsAlive() )
```
misyltoad: You already checked this.

src/game/server/hl2/item_ammo.cpp
```diff
@@ -929,7 +932,9 @@ int CItem_AmmoCrate::OnTakeDamage( const CTakeDamageInfo &info )
 	{
 		CBaseCombatWeapon *weapon = player->GetActiveWeapon();
 
-		if (weapon && !stricmp(weapon->GetName(), "weapon_crowbar"))
+		if ( weapon && ( ( !stricmp( weapon->GetName(), "weapon_crowbar" ) // Is my weapon a crowbar?
```
misyltoad: Can we just check for DMG_CLUB and not the weapon classname string now?

speedvoltage: Hi! First of all, please accept my apologies for responding months later. I somehow missed this.

In response to the comment, you could in theory just check DMG_CLUB indeed, but I come from the world of HL2DM and plugins may add new weapons that could potentially use DMG_CLUB, so that was why I made it my decision to check for `CBaseHL2MPBludgeonWeapon` instead of just `DMG_CLUB`.

PR #1101, opened 2025-03-14
[HL2MP] Fixed roll grenade sound missing

speedvoltage:
**Issue**:
The frag grenade doesn't use its rolling sound when using the secondary attack.

**Fix**:
Suppress prediction to restore this missing sound.


misyltoad:
Should pribably in braces around the WeaponSound call

misyltoad:
Not sure why this is necessary... This should get called on the client given there's code here...

I'll take a deeper look later.

Bitl:
Can confirm the roll grenade sound doesn't work in the SDK base by default

PR #1109, opened 2025-03-14
[HL2MP] Don't play fall damage if no damage is applied

speedvoltage:
**Issue**:
If no damage is applied during a fall damage, the fall damage sound still plays.

**Fix**:
Edit the check for fall damage to prevent the sound from playing when no damage is applied.


src/game/server/movehelper_server.cpp
dimhotepus, line 1: ```int iDamageTaken = xxxx;```

```iDamageTaken > 0``` and ```iDamageTaken >= 1``` always have same results.

It means such change can be dropped to reduce MR scope.

speedvoltage: Yes, this might be irrelevant now. I remember having to change this, because introducing a new `mp_falldamage` value to disable fall damage would still cause the sound to play in some situations. I can't seem to reproduce this now, so I will just close this.

PR #1111, opened 2025-03-14
[HL2MP] Fix external prop influence

speedvoltage:
**Issue**:
Whenever player 1 tags an explosive prop e.g. uses their pistol to shoot one bullet on an explosive oil drum, which doesn't set it on fire, that barrel has that player 1 as the owner, no matter what happens next. Therefore if that prop is used in the future by player 2 and kills player 3, then player 1 gets the point/frag because the game assumes they are the one who made the kill and shows so in the death notice.

**Fix**:
Reset prop ownership after 2 seconds if it was not manipulated by anyone after the last interaction.

### DIFF
--- a/src/game/server/props.cpp
+++ b/src/game/server/props.cpp
@@ -1050,7 +1050,7 @@ int CBreakableProp::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 		return 1;
 	}
 
-	if( info.GetAttacker() && info.GetAttacker()->MyCombatCharacterPointer() )
+	if ( !IsOnFire() && ToBaseCombatCharacter( info.GetAttacker() ) != NULL )
 	{
 		m_hLastAttacker.Set( info.GetAttacker() );
 	}
@@ -1164,6 +1164,11 @@ int CBreakableProp::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 	float flRatio = clamp( (float)m_iHealth / (float)m_iMaxHealth, 0.f, 1.f );
 	m_OnHealthChanged.Set( flRatio, info.GetAttacker(), this );
 	m_OnTakeDamage.FireOutput( info.GetAttacker(), this );
+
+	if ( !IsOnFire() )
+	{
+		m_hLastAttacker.Term();
+	}
 
 	return ret;
 }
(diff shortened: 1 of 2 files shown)


src/game/server/props.cpp
```diff
@@ -1050,7 +1050,7 @@ int CBreakableProp::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 		return 1;
 	}
 
-	if( info.GetAttacker() && info.GetAttacker()->MyCombatCharacterPointer() )
+	if ( !IsOnFire() && ToBaseCombatCharacter( info.GetAttacker() ) != NULL )
 	{
 		m_hLastAttacker.Set( info.GetAttacker() );
```
Mooshua, line 1055: Doesn't this already have the intended behavior of setting the last attacker whenever a prop takes damage? Now, if the prop is on fire and get insta-detonated by another player, the credit goes to the person who set the prop on fire

src/game/server/props.cpp
```diff
@@ -1675,22 +1680,19 @@ void CBreakableProp::Break( CBaseEntity *pBreaker, const CTakeDamageInfo &info )
...
 
 	CBaseEntity *pAttacker = info.GetAttacker();
-	if ( m_hLastAttacker )
-	{
-		// Pass along the person who made this explosive breakable explode.
-		// This way the player allies can get immunity from barrels exploded by the player.
-		pAttacker = m_hLastAttacker;
-	}
-	else if( m_hPhysicsAttacker )
+	if ( HasPhysicsAttacker() )
```
Mooshua, line 1683: Why are we overriding m_hLastAttacker with `HasPhysicsAttacker()`? If `Break` is called from `Event_Kill` then this will override the person who actually caused the damage with the person holding the prop with the physgun. If someone is holding an exploding barrel, wouldn't this make them the attacker in this case?

src/game/server/props.cpp
```diff
@@ -1801,10 +1803,11 @@ void CBreakableProp::Break( CBaseEntity *pBreaker, const CTakeDamageInfo &info )
...
 		}
 	}
+}
 
-#ifndef HL2MP
-	UTIL_Remove( this );
-#endif
+void CBreakableProp::OnBreak( const Vector &, const AngularImpulse &, CBaseEntity * )
+{
+	Remove();
```
Mooshua, line 1810: Why are we causing the remove to occur so early? We haven't even finished the `Break` invocation and we're already destroying the prop? Would having `EFL_KILLME` be set while the damage handler was still running cause any unexpected behavior?

src/game/server/props.cpp
```diff
@@ -2657,6 +2662,16 @@ void CPhysicsProp::InputWake( inputdata_t &inputdata )
 	}
 }
 
+void CPhysicsProp::Break( CBaseEntity *pBreaker, const CTakeDamageInfo &info )
+{
+	m_bFirstCollisionAfterLaunch = false;
+
+	// Setup the think function to remove the flags
+	RegisterThinkContext( "PROP_CLEARFLAGS" );
+	SetContextThink( &CPhysicsProp::ClearFlagsThink, gpGlobals->curtime, "PROP_CLEARFLAGS" );
```
Mooshua, line 2671: Why are we setting a think routine in the same breath we're calling `Break`?

Mooshua (commented):
I'm not entirely sure what this PR aims to achieve, but I do think there should be some more thought put into this. This seems like a LOT of change  for behavior that should already be default

PR #1117, opened 2025-03-15
[TF2] Fix organ props spawning from non-Vita-Saw damage

Tkain:
# Description

As part of the Vita-Saw's weapon gimmick of collecting organs from players, organ props are spawned whenever a player is hit with the Vita-Saw. However, the check for when to spawn an organ prop is imprecise; currently, TF2 will spawn a prop when a player takes damage if their attacker is a Medic holding the Vita-Saw. This logic means that if a Medic shoots a projectile (ex. a Crusader's Crossbow syringe) and switches to the Vita-Saw before the projectile hits a player, the projectile hit will cause an organ prop to spawn. This is presumably unintended, as such a hit does not increment a player's organ count.

This PR makes it so organ props will only spawn when a player takes damage if the _attacking weapon_ is the Vita-Saw, preventing the prior case from spawning props.

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -9095,44 +9095,40 @@ int CTFPlayer::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 		}
 	}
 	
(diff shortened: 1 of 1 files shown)


wgetJane:
lame!

PR #1118, opened 2025-03-16
[TF2] Prevent inspecting the Flamethrower while firing

Tkain:
# Description

In TF2, you can (sometimes) inspect the flamethrower while firing it. This causes the inspect animation to play while the flamethrower is emitting flames, which looks pretty weird:

https://github.com/user-attachments/assets/f70bc753-e159-4a52-8e7a-e6f3224f428b

This PR disables inspecting the flamethrower while firing it.


HyperionCoding:
Other weapons should be checked too while at it. At least minigun comes to my mind.

PR #1119, opened 2025-03-16
fix (most) shield charge prediction errors

claiwe:
# Description
This PR fixes the prediction errors generated for shield charge meters caused by the client not applying the `TF_COND_SHIELD_CHARGE` condition (and thus the charge meter being off by a few ticks).


wgetJane:
haven't tested this yet, but i think this'd only fix prediction for when you begin the charge

afaik these are the parts of demoknight shield charging that's unpredicted:
\- starting the charge (you will suddenly rubberband forwards)
\- ending a charge by running out of charge meter (you will suddenly rubberband backwards)
\- ending a charge by hitting an obstacle (walls, players, etc)
\- the charge regen itself (constant pred errors with cl_showerror 2 while your charge is regenerating)

 (https://github.com/ValveSoftware/Source-1-Games/issues/6017)

PR #1120, opened 2025-03-16
allow caber to get shield mini-crits

claiwe:
# Description
The caber doesn't get shield mini-crits, while most other demo melees do. Not sure if this is a bug but it seemingly might be a balance change, and #690 was closed without any further input, so I'll open this anyways.

### DIFF
--- a/src/game/shared/tf/tf_gamerules.cpp
+++ b/src/game/shared/tf/tf_gamerules.cpp
@@ -6109,9 +6109,9 @@ bool CTFGameRules::ApplyOnDamageModifyRules( CTakeDamageInfo &info, CBaseEntity 
 					eDamageBonusCond = TF_COND_OFFENSEBUFF;
 				}
 			}
-			else if ( pTFAttacker && (bitsDamage & DMG_RADIUS_MAX) && pWeapon && ( (pWeapon->GetWeaponID() == TF_WEAPON_SWORD) || (pWeapon->GetWeaponID() == TF_WEAPON_BOTTLE)|| (pWeapon->GetWeaponID() == TF_WEAPON_WRENCH) ) )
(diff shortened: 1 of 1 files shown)


Marxvee:
This would not allow the explosion hit to minicrit via charging by the way-- unless that's the intention. I'd personally argue that the explosion should also minicrit, since it can already fullcrit via charging.

If you were to allow the explosion to minicrit, you'd have to first add the damagebit in tf_weapon_bottle.cpp under `CTFStickBomb::Smack()`. Change the following block of code under `int dmgType = DMG_BLAST | DMG_NOCLOSEDISTANCEMOD;` from this,
```
			if ( IsCurrentAttackACrit() )
				dmgType |= DMG_CRITICAL;
```	
to this:
```
			if ( IsCurrentAttackACrit() )
				dmgType |= DMG_CRITICAL;
			else if  (m_bMiniCrit )
				dmgType |= DMG_RADIUS_MAX;
```	
then make an exception in tf_gamerules.cpp under `CTFRadiusDamageInfo::CalculateFalloff()` and change the first statement below from this:
```
	if ( dmgInfo->GetDamageType() & DMG_RADIUS_MAX )
```
to this:
```
	if ( dmgInfo->GetDamageCustom() != TF_DMG_CUSTOM_STICKBOMB_EXPLOSION && dmgInfo->GetDamageType() & DMG_RADIUS_MAX )
```

This should make the caber explosion also minicrit from charging, and not suffer from dealing less damage than intended due to the damage bit.

claiwe:
> This would not allow the explosion hit to minicrit via charging by the way-- unless that's the intention. I'd personally argue that the explosion should also minicrit, since it can already fullcrit via charging.

I do think the caber explosion should mini-crit if the caber itself does. In my time of sleepy I completely forgot to add that to this PR, oopsies

src/game/shared/tf/tf_weapon_bottle.cpp
```diff
@@ -257,6 +257,8 @@ void CTFStickBomb::Smack( void )
 			int dmgType = DMG_BLAST | DMG_USEDISTANCEMOD;
 			if ( IsCurrentAttackACrit() )
 				dmgType |= DMG_CRITICAL;
+            else if ( m_bMiniCrit )
+                dmgType |= DMG_RADIUS_MAX;
```
FlaminSarge: Check the tabs on these, most of the SDK uses \t instead of spaces

PR #1121, opened 2025-03-16
[TF2] Make the respawn timer display "Respawning is not available" for extremely long respawn time

Lizard-Of-Oz:
On certain featured maps (Halloween Arena and VS Saxton Hale) and many community maps respawn is "disabled" by setting the respawn wave time to a very high number (e.g. 99999), but the game still displays it as "Respawning in N seconds...".

This PR makes the game display "Respawning is not available" in such a case, which will add a touch of "legitimacy".
It also hides the respawn countdown from the team composition panel.

This PR requires the addition of a new locale
`"game_respawntime_disabled"		"Respawning is not available"`

Because this PR checks the time _left_ until respawn, mappers shall set the respawn wave time way above 1000 seconds used in this code commit.

image

### DIFF
--- a/src/game/client/tf/tf_teamstatus.cpp
+++ b/src/game/client/tf/tf_teamstatus.cpp
@@ -298,7 +298,7 @@ bool CTFTeamStatusPlayerPanel::Update( void )
 		if ( iRespawnWait != m_iPrevRespawnWait )
 		{
 			m_iPrevRespawnWait = iRespawnWait;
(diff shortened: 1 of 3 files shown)


HyperionCoding:
IMO, using negative values to indicate no respawning would make more sense.

ficool2:
> IMO, using negative values to indicate no respawning would make more sense.

This wouldn't accommodate backwards compatibility for maps (which is most of them) that use super high respawn times. Should check both <0 and >1000

Issue #663, opened 2025-02-19
TF2: CUtlBuffer crash when loading local loadout


ClayHanson:
When running Frog Fortress 2 from the launcher (mod_tf_win64.exe), the game opens fine, but then asserts when retrieving my local loadout from the item servers. If I ignore the assert, it immediately re-appears. Ignoring all assertions crashes the game.

The callstack references functions I do not have access to.

The assertion:
![Image](https://github.com/user-attachments/assets/64d7ca82-7de0-49c1-afc9-09158f56750e)

The callstack:
![Image](https://github.com/user-attachments/assets/763e3db2-c006-46ab-bac8-5dcd4ece6ac1)

The last function on the callstack before running into functions i don't have the sourcecode for:
![Image](https://github.com/user-attachments/assets/0e9b64c4-19ba-496b-b566-44bd44071a0c)

Where it crashes specifically:
![Image](https://github.com/user-attachments/assets/74d4f841-75bd-4d83-90d4-b2dba48ca359)

Marioiscool246:
Compile under Release, Debug enables assertions.

ClayHanson:
Can confirm that the release build does not crash, so why does the debug build crash? Debug should not crash

995qa:
I read this while the game was compiling under Debug mode! Ty for letting me know, ig I'll restart the build

TorutheRedFox:
I got it to fail gracefully by replacing line 968 of tf_item_inventory.cpp with
```C++
CEconItem* socData = pItem->GetSOCData();
if (socData)
	socData->Equip(iClass, iSlot);
```

granted, this makes your inventory inaccessible in a debug build still as it doesn't solve the root cause of the issue, but at least it makes a debug build usable at all

P.S. the assert can be fixed by changing it to check for it being >= 0 instead, as per https://developer.valvesoftware.com/wiki/SDK_Known_Issues_List#Assertion_Failed_in_utlbuffer.cpp

zxz41:
```
GCSDK::CGCClientSharedObjectTypeCache::BParseCacheSubscribedMsg(
		const CMsgSOCacheSubscribed_SubscribedType &msg,
		CUtlVector<GCSDK::CSharedObjectt*,CUtlMemory<GCSDK::CSharedObject *,int>> &vecCreatedObjects,
		CUtlVector<GCSDK::CSharedObject *,CUtlMemory<GCSDK::CSharedObject *,int>> &vecUpdatedObjects,
[...]

PR #1123, opened 2025-03-17
[TF2] Fix being able to taunt while loading the Beggar's Bazooka

Tkain:
# Description

In TF2, if a player taunts while loading rockets into the Beggar's Bazooka, the loaded rockets will only fire after the taunt finishes playing. Using this with a looping taunt effectively allows a Bazooka user to store their loaded rockets for as long as they want and fire them whenever they want. (This doesn't significantly impact gameplay, but it has been used to troll players in friendly scenarios.)

This PR fixes this by not allowing a Bazooka user to taunt while any rockets are loaded into the Bazooka.

This resolves https://github.com/ValveSoftware/Source-1-Games/issues/2703.

### DIFF
--- a/src/game/shared/tf/tf_weapon_rocketlauncher.cpp
+++ b/src/game/shared/tf/tf_weapon_rocketlauncher.cpp
@@ -225,6 +225,14 @@ void CTFRocketLauncher::ModifyEmitSoundParams( EmitSound_t &params )
 	}
 }
 
+bool CTFRocketLauncher::OwnerCanTaunt( void )
+{
+	if ( AutoFiresFullClip() && ( m_iClip1 > 0 ) )
+		return false;
+
+	return BaseClass::OwnerCanTaunt();
+}
+
 void CTFRocketLauncher::Misfire( void )
 {
 	BaseClass::Misfire();
--- a/src/game/shared/tf/tf_weapon_rocketlauncher.h
+++ b/src/game/shared/tf/tf_weapon_rocketlauncher.h
@@ -55,6 +55,8 @@ public:
 
 	virtual int		GetWeaponID( void ) const			{ return TF_WEAPON_ROCKETLAUNCHER; }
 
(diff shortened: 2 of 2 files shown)


Trico-Everfire:
If I can be honest. This has kinda become an iconic part of TF2. I don't think it particularly hurts anything in casual. it's part of a lot of funny meme strats and is funny if you stack multiple beggars soldiers. Feels kinda weird to fix it tbh.

BringBackQuickPlay:
https://www.youtube.com/watch?v=gvdf5n-zI14

Brain-dawg:
You can also do it while taunting with mannrobics and likely other taunts that do special actions on attack, don't need to load before taunting.  Blocking taunting while your clip is non-zero wouldn't fix this.

Tkain:
Well that doesn't mean this fix doesn't work as advertised; the special action bug you mentioned is just a different way of triggering it that needs its own bug fix. (And unfortunately I can't provide that fix because I own no such taunts to test it with.)

Trico-Everfire:
> Well that doesn't mean this fix doesn't work as advertised; the special action bug you mentioned is just a different way of triggering it that needs its own bug fix. (And unfortunately I can't provide that fix because I own no such taunts to test it with.)

Then... Code a trigger for that taunt... You have source code access. Pressing that specific taunt in the menu just triggers code, just call the trigger from elsewhere.

Tkain:
> Pressing that specific taunt in the menu just triggers code, just call the trigger from elsewhere.

My goodness, what an idea. Why didn't I think of that?

But seriously, come on. If you'd looked into how it actually works, you'd know that TF2's item system is strongly data-driven to the point where it isn't as simple as just "calling the trigger from elsewhere". It doesn't help that the system for getting the player's inventory currently [completely breaks on debug builds](https://github.com/ValveSoftware/source-sdk-2013/issues/663). But hey, prove me wrong - you have source code access too.

PR #1125, opened 2025-03-17
[TF2] Fix flipped viewmodels for Scout's baseball/bauble and Botkiller heads

Marxvee:
# Description
This pull request introduces two visual fixes related to flipped viewmodels. Currently, the held baseball/bauble while the Sandman/Wrap Assassin is active and all botkiller weapon attachment models in first person have inverted normals while the client's viewmodels are flipped, 

Below is a before and after of these changes (all screenshots are taken with cl_flipviewmodels set to 1):
## Before:
image
image

## After:
image
image


Marxvee:
This has been updated to use the new TeamFortress_ShouldFlipClientViewModel() function that was added in 20877e4.

PR #1126, opened 2025-03-17
Fix crash on VGUI menu bar button cursor entrance

RGBACatlord:
# Description
This is done in Menu as well but missed here, crashes engine -tools.


ficool2:
Why the #ifdefs? The 64-bit code will work on both architectures

RGBACatlord:
I just copied it over from vgui::Menu 🤷‍♂️

PR #1131, opened 2025-03-20
Fix unreachable BSP lumpsize warning messages

treacherousfiend:
# Description
The map compilers print out a nice table of stats showing how much space of the various BSP lumps is filled, and there is supposed to be 3 messages that display if the filled space ranges from >80% full ("VERY FULL!"), >95% full ("SIZE DANGER!"), and >99.9% full ("SIZE OVERFLOW!!!"). However the ordering of the if/else if statements only allows the first message to be displayed, since 80% is the first if statement that is checked, and anything above that, including 95% and 99.9%, are of course greater than 80%, and only show the message associated with that.

This PR fixes that by simply swapping the order of the if/else if statements around, checking against 100% first, then 95%, and then 80%, rather than 80 -> 95 -> 100.

## Additional changes
- Changed the >99.9% check to >100%, since you can have a lump be 100% full and not have any issues, and its not an overflow until it goes over 100%
- changed the percentage constants to floats instead of doubles. Minor thing, just noticed it and figured I'd change it since the percentage variable is a float anyway.

## Notes
- I don't think in most situations the 100% one will be hit? I believe most if not all of the lumps cause the compiler to error out and stop the compiler if overflowed preventing it from even getting to this list. I have heard of some of the percentages not getting calculated correctly and going over 100%, but at that point its inaccurate anyway and doesn't really matter.
- This is a super duper minor thing to change, and honestly the one message at >80% is enough to get the attention of the mapper to let them know how high their usage is, but I noticed these messages were here and figured I might as well just fix them.

the previously unused messages in action:
image
image

### DIFF
--- a/src/utils/common/bsplib.cpp
+++ b/src/utils/common/bsplib.cpp
@@ -2829,16 +2829,16 @@ void	WriteLumpToFile( char *filename, int lump, int nLumpVersion, void *pBuffer,
 
 int ArrayUsage( const char *szItem, int items, int maxitems, int itemsize )
 {
-	float	percentage = maxitems ? items * 100.0 / maxitems : 0.0;
+	float	percentage = maxitems ? items * 100.0f / maxitems : 0.0f;
 
     Msg("%-17.17s %8i/%-8i %8i/%-8i (%4.1f%%) ", 
 		   szItem, items, maxitems, items * itemsize, maxitems * itemsize, percentage );
-	if ( percentage > 80.0 )
-		Msg( "VERY FULL!\n" );
-	else if ( percentage > 95.0 )
-		Msg( "SIZE DANGER!\n" );
-	else if ( percentage > 99.9 )
+	if ( percentage > 100.0f )
 		Msg( "SIZE OVERFLOW!!!\n" );
+	else if ( percentage > 95.0f )
+		Msg( "SIZE DANGER!\n" );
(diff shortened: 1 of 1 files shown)


0x5066:
Wouldn't it then make sense to then additionally print stuff like that? This would cause less confusion on the mapper's end and doesn't necessarily mean that things are bad

ficool2:
These warnings sound too threatening, some limits such as entdata can be safely exceeded too. I think it's better to not include them

treacherousfiend:
> Some of the limits are mere recommendations, such as the entdata lump, that's why they can reach above 100 %.

> These warnings sound too threatening, some limits such as entdata can be safely exceeded too. I think it's better to not include them

Agreed, to be honest I wouldn't be surprised if someone at Valve already noticed this "issue" years ago and intentionally left it since it is such a minor thing that makes it sound significantly more serious than it actually is, since in most cases its not an issue until it actually breaks the size limit, at which point the compile fails anyway, and unless it was the first compile of a map or at least the first compile after adding, say 410 or so overlays to a map that previously had none, the mapper already has plenty of warning that they're at least getting _close_ to a limit.

I think ideally maybe this whole system should be rewritten to better account for the actual limits these files have (and maybe don't stop counting the number of items when the map does break limits in some cases like tjuncs), but that's a project for another day.

I'll leave this PR open though, I don't really see a reason to close it

PR #1134, opened 2025-03-21
[TF2] Fix jar weapon skins sometimes not being kept when thrown

Tkain:
# Description

According to the TF2 Wiki, reskins of the Jarate and the Mad Milk may not use their own models (instead using the model of the regular Jarate and Mad Milk respectively) when thrown if the throwing player switches to a different weapon quickly enough. This is likely because the projectile's model is set based on the player's active weapon when the projectile is thrown, which may not necessarily be the throwing weapon:

https://github.com/ValveSoftware/source-sdk-2013/blob/a62efecf624923d3bacc67b8ee4b7f8a9855abfd/src/game/shared/tf/tf_weapon_jar.cpp#L250-L273

This PR addresses this issue by changing how the throwing weapon is retrieved; instead of using `GetActiveWeapon` (the thrower's active weapon), we use `Weapon_OwnsThisID` to get whatever kind of weapon the thrower has which might throw the current projectile (ex. `TF_WEAPON_JAR` for Jarate).

~~I can't actually reproduce or test the original bug because I don't own any reskins of the Jarate or Mad Milk, so I'd appreciate if someone who owns these could see if they can reproduce the original bug and see if this PR fixes it.~~ (Update 6/9/2026: I have since been able to determine that is bug is reproducible in the SDK and that this PR fixes it at the time of writing.)

### DIFF
--- a/src/game/shared/tf/tf_weapon_jar.cpp
+++ b/src/game/shared/tf/tf_weapon_jar.cpp
@@ -143,9 +143,9 @@ CBaseEntity *CTFJar::FireJar( CTFPlayer *pPlayer )
 // Purpose:
 //-----------------------------------------------------------------------------
 CTFProjectile_Jar *CTFJar::CreateJarProjectile( const Vector &position, const QAngle &angles, const Vector &velocity, 
-												   const AngularImpulse &angVelocity, CBaseCombatCharacter *pOwner, CTFWeaponBase *pWeapon )
+												   const AngularImpulse &angVelocity, CBaseCombatCharacter *pOwner, const CTFWeaponInfo &weaponInfo )
 {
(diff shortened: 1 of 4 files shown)


torzod:
i can reproduce the bug consistently with all 3 jar reskins, but am unable to test the fix at this time

wgetJane:
any update on this? cant test since i dont own any of the reskins either lol

Whurrhurr:
you could try to test it out using [this map](https://steamcommunity.com/sharedfiles/filedetails/?id=3415171198) (souped up version of itemtest using vscript to give you the reskins, if you don't have them, the menus in this map are opened with mouse3.)

Tkain:
Thanks for the tip, @Whurrhurr - I can now verify that this PR works as intended.

PR #1140, opened 2025-03-22
Fix function name in MenuBar for 32-bit platforms

Bitl:
# Description

In the most recent commit, I noticed that the `MenuBar::OnCursorEnteredMenuButton` is erroneously defined as `void Menu::OnCursorEnteredMenuItem(int VPanel)` on non 64-bit platforms. While it might be unnecessary since this repository compiles for 64-bit platforms primarily, I felt like this helps with the consistency of the codebase.


Bitl:
closed because it's been fixed in the latest commit

PR #1091, opened 2025-03-14
[HL2MP] Fix infinite stack loop on "angle" entity KeyValue

speedvoltage:
**Issue**:
When attempting to set the angle keyvalue on an entity, there is a risk of encountering an endless stack loop.

**Fix**:
The update ensures that the angle keyvalue is assigned properly, preventing any unintended recursion. By modifying the method of processing the keyvalue, this fix eliminates the infinite loop and ensures that angles are assigned securely, thus avoiding any stack overflow issues.

PR #1143, opened 2025-03-23
angle keyvalue crash fix baseentity_shared.cpp

Nesciuse:
# Description

"angle" keyvalue causes a crash because of a mistake in CBaseEntity::KeyValue method which causes infinite recursion


kisak-valve:
Hello @Nesciuse, this pull request is a duplicate of #380.

blackletum:
> Hello @Nesciuse, this pull request is a duplicate of #380.

Since #380 was closed because the head repo was deleted, could this PR be reopened?

kisak-valve:
Noting that the same happened on #1091.

Nesciuse:
I am just genuinely wondering how can something this simple and so obviously unable to break anything (as the alternative is straight up infinite recursion crash every time, so not even any backwards compatibility is in the question here) be still not merged.

PR #1147, opened 2025-03-24
Added `level_visits` userinfo convar

Lizard-Of-Oz:
In VScript projects, it would be useful to know if a particular player is new to the map/gamemode, to, for example, display hint messages. Even TF2 itself does this when displaying the map's goal description on round start.

Fortunately, the game already keeps track of this statistic - all we need is to put into a userinfo convar so that VScript could retrieve it using `GetClientConvarValue`. I think it's a good compromise between safety and versatility, giving that cookies are most likely out of the question due to privacy and security reasons.


henke37:
Is a convar the best way to communicate here? Aren't those usually for read/write data?

Lizard-Of-Oz:
It's the most straightforward and the least disruptive approach. Everything is already in place, all that's needed is just 2,5 lines of code.

PR #1150, opened 2025-03-24
Add %alphatexture (arbitrary VTF for texture shadows)

SirYodaJedi:
# Description
If present, texture shadows will be generated from alpha of `%alphatexture` instead of `$basetexture`. This also permits texture shadows from static props without `$translucent` or `$alphatest` being present.

I also moved `$alphatest` check before `$translucent` check, because it's probably more common (not that that really speeds much up).

I refrained from renaming the now inaccurately-named variables because I didn't want to accidentally break anything.

# Additional information
For more information, see [VDC page](https://developer.valvesoftware.com/wiki/alphatexture) and https://github.com/Facepunch/garrysmod-requests/issues/2603

# Legalese
All code is either written by me or is existing SDK code. Permission is granted to use my changes as you see fit; no credit is necessary.

### DIFF
--- a/src/utils/vrad/vradstaticprops.cpp
+++ b/src/utils/vrad/vradstaticprops.cpp
@@ -703,9 +703,17 @@ public:
 			if ( pVMT->LoadFromBuffer( pMaterialName, buf ) )
 			{
 				bFound = true;
-				if ( pVMT->FindKey("$translucent") || pVMT->FindKey("$alphatest") )
+				if ( pVMT->FindKey("$alphatest") || pVMT->FindKey("$translucent") || pVMT->FindKey("%alphatexture") )
 				{
-					KeyValues *pBaseTexture = pVMT->FindKey("$basetexture");
+					KeyValues *pBaseTexture = NULL;
+					if ( pVMT->FindKey("%alphatexture") )
+					{
+						pBaseTexture = pVMT->FindKey("%alphatexture");
+					}
+					else
+					{
+						pBaseTexture = pVMT->FindKey("$basetexture");
+					}
 					if ( pBaseTexture )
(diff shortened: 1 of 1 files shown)


src/utils/vrad/vradstaticprops.cpp
```diff
@@ -703,9 +703,17 @@ class CShadowTextureList
...
 			{
 				bFound = true;
-				if ( pVMT->FindKey("$translucent") || pVMT->FindKey("$alphatest") )
+				if ( pVMT->FindKey("$alphatest") || pVMT->FindKey("$translucent") || pVMT->FindKey("%alphatexture") )
 				{
-					KeyValues *pBaseTexture = pVMT->FindKey("$basetexture");
+					KeyValues *pBaseTexture = NULL;
+					if ( pVMT->FindKey("%alphatexture") )
+					{
+						pBaseTexture = pVMT->FindKey("%alphatexture");
```
ficool2: You are calling `pVMT->FindKey("%alphatexture")` 3 times

ficool2 (changes requested):

src/utils/vrad/vradstaticprops.cpp
SirYodaJedi: I mean, I could store it in a temp variable, but that adds a bunch of complexity for insignificant speed gains.

SirYodaJedi: Would this be preferred? It does mean searching for a likely missing `%alphatexture` even if `$alphatest` and/or `$translucent` are present.
image
Actually, I guess it would be doing that regardless. I'll compile and test, then commit your suggestion.

PR #1154, opened 2025-03-26
Fix TDM negative scoring

### LINKED ISSUE #1153: [HL2MP] Team Deathmatch negative scoring is broken
For non-player-caused deaths, half of the scenarios score a negative one (-1) point to your team, while the other half doesn't change teams scores at all, while always scoring a negative one personally to the player who died. This doesn't make sense.

[Here's a video](https://www.youtube.com/watch?v=vNtpC6fqMOY) which demonstrates the current behavior.

Tripperful:
Fixes #1153

### DIFF
--- a/src/game/server/hl2mp/hl2mp_player.cpp
+++ b/src/game/server/hl2mp/hl2mp_player.cpp
@@ -1322,16 +1322,19 @@ void CHL2MP_Player::Event_Killed( const CTakeDamageInfo &info )
 
 	CBaseEntity *pAttacker = info.GetAttacker();
 
-	if ( pAttacker )
+	if ( HL2MPRules()->IsTeamplay() )
 	{
-		int iScoreToAdd = 1;
-
-		if ( pAttacker == this )
+		// Check for the attacker being in team Unassigned to account
+		// for non-player attackers, which are in this team by default.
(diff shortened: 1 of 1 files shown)


Bitl:
I feel like this would break scores on FFA servers. I feel like the previous code should be added as an alternative case.

example:

```
if (HL2MPRules()->IsTeamplay())
{
	// Check for the attacker being in team Unassigned to account
	// for non-player attackers, which are in this team by default.
	// In TDM, should only happen with deaths to non-player causes.
	if (pAttacker && !pAttacker->InSameTeam(this) && pAttacker->GetTeamNumber() != TEAM_UNASSIGNED)
	{
		pAttacker->GetTeam()->AddScore(1);
	}
	else
	{
		GetTeam()->AddScore(-1);
	}
}
else if (pAttacker)
{
	int iScoreToAdd = 1;

	if ( pAttacker == this )
	{
		iScoreToAdd = -1;
	}

	GetGlobalTeam( pAttacker->GetTeamNumber() )->AddScore( iScoreToAdd );
}
```

Tripperful:
In FFA, team score is never shown in the scoreboard (or anywhere else) and its current value for team Unassigned doesn't make any sense anyways

PR #1155, opened 2025-03-26
[All?] Fix exploit allowing color control characters to be used in chat under certain circumstances

AndrewBetson:
# Description
This PR fixes an oversight that causes the `ReadChatTextString` function to skip over color control characters if they are preceded by either of the hex color control codes.

This happens because the inner for-loop increments `test` one too many times, causing any following color control characters to be skipped over when the outer for-loop increments `test`, itself.

As far as I can tell, this effects all SDK2013 games.

The following screenshots use the example string `color!FFD700test` (`0x07636F6C6F7221070546464437303074657374`), generated with https://sourcecolors.neocities.org/

Before             |  After
chat-color-before  |  chat-color-after

Note that for reasons I'm not entirely sure of, this only seems to effect *dead*/spectate/team chat. (in TF2 at least)

### DIFF
--- a/src/game/client/hud_basechat.cpp
+++ b/src/game/client/hud_basechat.cpp
@@ -178,9 +178,9 @@ wchar_t* ReadChatTextString( bf_read &msg, OUT_Z_BYTECAP(outSizeInBytes) wchar_t
 			{
 				// mark the next seven or nine characters. one for the control character and six or eight for the code itself.
 				const int nSkip = ( *test == COLOR_HEXCODE ? 7 : 9 );
-				for ( int i = 0; i < nSkip && *test != 0; i++, test++ )
+				for ( int i = 0; i < nSkip && *test != 0; i++ )
 				{
-					*test = COLOR_NORMAL;
+					*++test = COLOR_NORMAL;
 				}
 
 				// if we reached the end of the string first, then back up


AzureWoof:
This doesn't seem all too harmful. Would much rather this be a feature, with some restrictions of course.

origintopleft:
So far this has exactly one "malicious" use: Faking item drops. I learned about this when someone faked getting a Valve Rocket Launcher. (I chose something that doesn't actually exist in the game as proof of concept)

20250329150922_crop

I don't know how much this is considered a critical thing to fix, but if we can accept this as a thing people do, I'd vote in favor of people having colored text chat (if changed to something like Quake's system, with ^ suffixed by a value)

universe-breaker:
it would be very cool to just add this as a feature, but that the server hoster can enable and disable this feature instead of just removing this from the game.

BatPisz:
This honestly is such a great thing that's just so happened to arise from a bug. This is akin to the A-Pose glitch(es) in the game - it's harmless fun that has little to no chances of actually causing serious gameplay issues, so it's probably best remaining unpatched.

just-a-github-user:
> This honestly is such a great thing that's just so happened to arise from a bug. This is akin to the A-Pose glitch(es) in the game - it's harmless fun that has little to no chances of actually causing serious gameplay issues, so it's probably best remaining unpatched.

These last few days with new developments, players are beginning to get impersonated using colored chat messages. By using a blank name, and then using this exploit to color a player's name their respective team color, it's been used for griefing and impersonation on public servers
https://www.youtube.com/watch?v=jPEogX5Chjw

BatPisz:
> By using a blank name, and then using this exploit to color a player's name their respective team color, it's been used for griefing and impersonation on public servers

Appears to be more a problem of blank names being exploited and using this as a leverage, rather than the coloring being an issue itself. Which begs the question of what is worth prioritizing - keeping _empty_ usernames, or allowing players to sacrifice text slots for chat coloration?

PR #1165, opened 2025-03-30
Restored unused Rocket Destroyed voice response

rabscootle:
Updated tf_weapon_minigun.cpp to include a trigger for unused MP_CONCEPT_ROCKET_DESTOYED concept. RocketDestroyedHeavy response rule will now play in-game.

Added checks for lines to not be truncated by pain audio (Heavy will typically be under fire when in these situations).

Updated scripts/talker/heavy.txt to lower frequency of rocket deflect vo (50% -> 5%)

https://github.com/user-attachments/assets/1c9e9713-bd60-409b-af49-9c48963e1e91

# Description

### DIFF
--- a/src/game/shared/tf/tf_weapon_minigun.cpp
+++ b/src/game/shared/tf/tf_weapon_minigun.cpp
@@ -629,9 +629,21 @@ void CTFMinigun::AttackEnemyProjectiles( void )
 				int nHitsRequired = m_bCritShot ? 1 : (int)RemapValClamped( iAttackProjectiles, 1, 2, 2, 1 );
 				if ( pProjectile->GetDestroyableHitCount() >= nHitsRequired )
 				{
+					int projectileType = pProjectile->GetBaseProjectileType();
 					pProjectile->Destroy( false, true );
 
 					EmitSound( "Halloween.HeadlessBossAxeHitWorld" );
+
+					// Trigger response rule, we only have vo for Rockets so limit to that type
+					if (projectileType == -1)
+					{
+						// Making sure Heavy has a chance to speak the lines, he will be typically be under fire in these scenarios
+						CMultiplayer_Expresser* pExpresser = pPlayer->GetMultiplayerExpresser();
+						Assert(pExpresser);
+						pExpresser->AllowMultipleScenes();
+						pPlayer->SpeakConceptIfAllowed(MP_CONCEPT_ROCKET_DESTOYED);
+						pExpresser->DisallowMultipleScenes();
+					}
 
 					CTF_GameStats.Event_PlayerAwardBonusPoints( pPlayer, NULL, 2 );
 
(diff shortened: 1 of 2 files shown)


src/game/shared/tf/tf_weapon_minigun.cpp
```diff
@@ -629,10 +629,22 @@ void CTFMinigun::AttackEnemyProjectiles( void )
 				int nHitsRequired = m_bCritShot ? 1 : (int)RemapValClamped( iAttackProjectiles, 1, 2, 2, 1 );
 				if ( pProjectile->GetDestroyableHitCount() >= nHitsRequired )
 				{
+					int projectileType = pProjectile->GetBaseProjectileType();
 					pProjectile->Destroy( false, true );
 
 					EmitSound( "Halloween.HeadlessBossAxeHitWorld" );
 
+					// Trigger response rule, we only have vo for Rockets so limit to that type
+					if (projectileType == -1)
```
brokenparts: It looks like the only projectile that overrides `GetBaseProjectileType` is `CTFWeaponBaseGrenadeProj` , so this check basically equates to "anything that isn't a grenade", which doesn't seem right. Maybe it would be better to explicitly check if the projectile is a rocket?

rabscootle: Latest commit should resolve this.
Discord_yeuY1foPc7

brokenparts: Does this work with cow mangler projectiles?

rabscootle: I didn't add them to the check since I didn't consider them rockets. They are now included in the latest commit.

brokenparts: Fair enough, just wanted to make sure it wasn't an oversight.

src/game/shared/tf/tf_weapon_minigun.cpp
```diff
@@ -633,6 +634,17 @@ void CTFMinigun::AttackEnemyProjectiles( void )
 
 					EmitSound( "Halloween.HeadlessBossAxeHitWorld" );
 
+					// Trigger response rule, we only have vo for Rockets so limit to that type
+					if ( FClassnameIs(pObject, "tf_projectile_rocket") || FClassnameIs(pObject, "tf_projectile_sentryrocket") || FClassnameIs(pObject, "tf_projectile_energy_ball") )
```
FlaminSarge: Rather than string classname checks, it may be better to dynamic_cast pProjectile to CTFProjectile_Rocket. This will cover rockets, sentryrockets, and the short circuit right-click (which I don't think gets deflected in MVM anyways); I'm not sure if it's necessary to include the Cow Mangler projectile, but you can always use CTFProjectile_EnergyBall for that if needed.
```
CTFProjectile_Rocket *pRocket = dynamic_cast< CTFProjectile_Rocket >( pProjectile );
if ( pRocket )
{
 ...
}
```

rabscootle: I updated the code to check this way. I'm okay with leaving the Cow Mangler projectiles out as aren't really used enough officially to warrant another check (they also aren't technically rockets).

PR #1172, opened 2025-04-01
Fixed bot_kick not working on TF2 or HL2 deathmap when using nav_generate.

### LINKED ISSUE #926: bot_kick in nav_generate instead of game-specific commands
https://github.com/ValveSoftware/source-sdk-2013/blob/fe72581dcb5e7bdd18bf385775845cc1c11289e9/src/game/server/nav_generate.cpp#L3437

This command need to be switched to `tf_bot_kick all` or `hl2mp_bot_kick all`, otherwise there is an error of kicking bots.

Unusuario2:
Closes #926 

# Description

### DIFF
--- a/src/game/server/nav_generate.cpp
+++ b/src/game/server/nav_generate.cpp
@@ -3434,7 +3434,13 @@ void CNavMesh::BeginGeneration( bool incremental )
 		}
 	}
 #else
-	engine->ServerCommand( "bot_kick\n" );
+	#if defined(HL2MP)
+		engine->ServerCommand("hl2mp_bot_kick all\n");
+	#elif defined(TF_CLIENT_DLL)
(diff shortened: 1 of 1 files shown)


src/game/server/nav_generate.cpp
```diff
@@ -3434,7 +3434,13 @@ void CNavMesh::BeginGeneration( bool incremental )
 		}
 	}
 #else
-	engine->ServerCommand( "bot_kick\n" );
+	#if defined(HL2MP)
+		engine->ServerCommand("hl2mp_bot_kick all\n");
+	#elif defined(TF_CLIENT_DLL)
```
ficool2: TF_CLIENT_DLL?

Unusuario2: Fixed

PR #1176, opened 2025-04-01
[ALL] Extended some internal BSP limits.

Unusuario2:
# Description
Overlays 512 -> 8192, This is a soft limit, is not check by the engine so it is complety safe to raise up to 65536
Entities 8192 -> 32768, This is an internal limit for compile tools, this allows mappers to use more internal entities without having an error, especially useful if you use a lot of instances, cubemaps, overlays.
Wordlights 8192 -> 65536 Increased number of static lights, internal limit as they dont exist in game.


src/utils/vbsp/detailobjects.cpp
```diff
@@ -500,7 +500,7 @@ static void AddDetailSpriteToLump( const Vector &vecOrigin, const QAngle &vecAng
 	// Insert an element into the object dictionary if it aint there...
 	int i = s_DetailObjectLump.AddToTail( );
 
-	if (i >= 65535)
+	if (i >= MAX_MAP_DETAIL_PROPS)
```
ficool2, line 503: This check and limit can just be removed altogether, as the detail object lump is variable sized

ficool2 (changes requested):

src/utils/vbsp/detailobjects.cpp
Unusuario2: Thanks for the info! I have removed the limit, but I think it makes more sense to warn the user if they exceed 65536 prop_details, as a high number of prop_details is known to overload DirectX and cause performance issues

SirYodaJedi:
> Wordlights 8192 -> 65536 Increased number of static lights, internal limit as they dont exist in game

Static worldlights do technically exist in-game, just not as entities. The two closest ones are used to cast elights on models. 

Ideally, there should be a way to make lights that only contribute to lightmaps and static prop lighting and aren't added to the worldlights lump, but that would require refactoring VRAD to read light info directly from the entities like in QRAD/HLRAD, and wouldn't help for texlights, since VRAD doesn't have a way to make texlights using entities instead of RAD files (light_surface when?).

PR #1177, opened 2025-04-01
[VRAD] Enable super sampling for displacements

Unusuario2:
# Description
Self explanatory, this allows a better lighting transition between brushes and displacements.


ficool2:
This is not enough, you also need to correct the sampling positions

Unusuario2:
> This is not enough, you also need to correct the sampling positions

I forgot. Let me check, i will do it soon.

PR #1183, opened 2025-04-05
Ensure server doesn't process or simulate anything whilst paused

drunderscore:
There are some very common bugs that occur whilst paused:
- Ubercharge being gained
- Cloak being lost
- Gravity and movement effecting players
There are other examples but there are the most critical.

_Most_ of this is caused by the fact that even whilst told not to simulate, players alone would be simulated. This would in turn cause them to think, and propagate through to their weapons, causing miscalculations based on how much time had truly passed.

Competitive communities have been using [this sourcemod plugin](https://github.com/F2/F2s-sourcemod-plugins/blob/master/pause/pause.sp) to try and work _around_ the issue, but it has always been inaccurate and occasionally faulty.

This patch is based on my original patch to p4sstime/p4ss: https://github.com/p4sstime/p4ss/commit/086edd7c62abb7230239a5efd005822e61170557


ficool2:
This is intentional. The engine still processes usercmds while paused for things like `sv_noclipduringpause` to work

Bitl:
> This is intentional and your changes break `sv_noclipduringpause`, a better solution might be to have a convar that forces everything to not be processed

I experimented with this and made this implementation:

https://github.com/FIREFIGHT-RELOADED/FIREFIGHT-RELOADED-src-sdk-2013/commit/870e72dea5e126ce0846eca445bfaadb0747850c

It:

1. Uses a cvar to disable player simulation (sv_pauseplayers).
2. Checks sv_noclipduringpause before doing anything. If it is enabled, it will not allow the game to access the value of  sv_pauseplayers until sv_noclipduringpause is disabled.

drunderscore:
I've changed it so the status-quo is maintained if `sv_noclipduringpause`.

I don't believe a new convar should be created for this -- similar conditionals are already seen in existing code _and_ in the engine, the latter of which would mean an even greater change to status-quo with regards to `sv_noclipduringpause`.

PR #1186, opened 2025-04-06
[TF2] Add generic condition for mini-crit boost

horiuchii:
# Description
This PR adds `TF_COND_MINICRITBOOSTED`, a condition that simply grants the visuals and the effect of a mini-crit buff.

While we already have the existing `TF_COND_OFFENSEBUFF` and `TF_COND_ENERGY_BUFF` conditions to apply the buff, this would be a great way to give it without having to deal with the various side effects these conditions grant.


BringBackQuickPlay:
100% yes.
Asked for this earlier on the feedback repo valve has.
It's ridiculous that we have like 4-5 different conditions to get normal crits and not a single one for mini-crits.

If I may be so bold, could you also add one that provides purely the effects without effecting damage?
kinda like TF_COND_MINICRITBOOSTED_NO_DMG incase one should want just the visual effects and adjust dmg later trough something like a sourcemod plugin? A version of that for the normal crits would also be nice.

antielectronbomb:
This would be a great feature to have, and should make implementing custom game modes much more easier to work with. A good feature to have for modders.

PR #1188, opened 2025-04-06
[TF2] Potential fix for chat window out of position in Casual/Competitive Match Start/End

rabscootle:
Potential fix for https://github.com/ValveSoftware/Source-1-Games/issues/3523

The game uses HUD animations (CompetitiveGame_LowerChatWindow and CompetitiveGame_RestoreChatWindow) to handle moving the chat box for the post-game screens. However, the chat window returns to its in-game position when transitioning to the Competitive 'Results' sequence as it is forced to reposition when players respawn for the podium.

https://github.com/user-attachments/assets/ef85a4e8-ed19-48d2-ae90-bd25e7aac811

This pr adds a check for if player is at end of Casual/Competitive match to retain chat window in lower position. This will minimize overlap with match summary UI on screen.

The live client also fails to relocate the chat window at the start of a match to a lower position. This results in some of the match start UI overlapping significantly:
https://user-images.githubusercontent.com/58406587/153318924-636f0fea-b32b-4a28-8798-d7037d631e25.png

This was previously not an issue in older versions of the game (picture below is from 2016 competitive beta):
image

Adds triggers for chat window to raise/lower during match start sequence to minimize overlap with team lineup UI.

# Description

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -7873,7 +7873,13 @@ void C_TFPlayer::ClientPlayerRespawn( void )
 		m_bNotifiedWeaponInspectThisLife = false;
 
 		// make sure the chat window has been restored to the appropriate place
-		g_pClientMode->GetViewportAnimationController()->StartAnimationSequence( "CompetitiveGame_RestoreChatWindow", false );
+		if ( TFGameRules() && (TFGameRules()->ShowMatchSummary() || TFGameRules()->PlayersAreOnMatchSummaryStage()) )
+		{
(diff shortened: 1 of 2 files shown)


mastercoms:
> The live client also fails to relocate the chat window at the start of a match to a lower position. This results in some of the match start UI overlapping significantly

I suspect this may also have to do with proportional VGUI? Maybe the animation is placing the chat window higher than what was originally intended.

rabscootle:
Possibly, my assumption was that they removed most of the cases handling the chat box moving when they fixed it not always returning to the in-game position:

image

https://www.teamfortress.com/post.php?id=63310

PR #1191, opened 2025-04-07
[TF2] Fix not being able to deploy BASE Jumper parachute in certain situations

horiuchii:
# Description
This PR fixes an issue where walking into freefall after landing with the parachute deployed whilst not holding the jump key will prevent you from redeploying the parachute as the `TF_COND_PARACHUTE_DEPLOYED` condition will still be active as the ground check only happens when the jump key is held. This also occurs when touching the resupply cabinet.

The change I've made to fix this issue is to remove the aforementioned cond whenever we remove the `TF_COND_PARACHUTE_ACTIVE` cond under normal circumstances.

https://github.com/user-attachments/assets/4907b0f9-c47c-44e7-b1ac-19d3f90444c3


horiuchii:
Whoops, can't believe I failed to find that PR, closing this as duplicate.

PR #1192, opened 2025-04-07
[TF2] Make it so Vita-Saw attacks on MvM robots spawn a bolt instead of a spleen (outside of Wave 666)

Bitl:
# Description

When using the Vita-Saw in MVM, it spawns a spleen despite the robots being, well, robots. This PR makes robots spawn a bolt gib rather than a spleen in MVM.

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -9113,7 +9113,7 @@ int CTFPlayer::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 				pRandomInternalOrgan->KeyValue( "origin", buf );
 				Q_snprintf( buf, sizeof( buf ), "%.10f %.10f %.10f", GetAbsAngles().x, GetAbsAngles().y, GetAbsAngles().z );
 				pRandomInternalOrgan->KeyValue( "angles", buf );
-				if (TFGameRules()->IsMannVsMachineMode() && GetTeamNumber() == TF_TEAM_BLUE)
+				if (TFGameRules() && TFGameRules()->IsMannVsMachineMode() && GetTeamNumber() == TF_TEAM_PVE_INVADERS)
 				{
 					//robots don't have spleens....
 					pRandomInternalOrgan->KeyValue("model", "models/player/gibs/gibs_bolt.mdl");
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -2774,6 +2774,8 @@ void CTFPlayer::PrecacheMvM()
 	PrecacheModel( "models/items/currencypack_large.mdl" );
 
 	PrecacheModel( "models/bots/tw2/boss_bot/twcarrier_addon.mdl" );
+    
+    PrecacheModel( "models/player/gibs/gibs_bolt.mdl" );
 
 	PrecacheParticleSystem( "bot_impact_light" );
 	PrecacheParticleSystem( "bot_impact_heavy" );
@@ -3072,7 +3074,6 @@ void CTFPlayer::PrecacheTFPlayer()
 	PrecacheModel( "effects/beam001_blu.vmt" );
 	PrecacheModel( "effects/beam001_white.vmt" );
 	PrecacheModel( "models/player/gibs/random_organ.mdl" );
-    PrecacheModel( "models/player/gibs/gibs_bolt.mdl" );
 
 	PrecacheScriptSound( "Weapon_Mantreads.Impact" );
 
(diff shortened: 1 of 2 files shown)


src/game/server/tf/tf_player.cpp
```diff
@@ -9112,7 +9113,15 @@ int CTFPlayer::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 				pRandomInternalOrgan->KeyValue( "origin", buf );
 				Q_snprintf( buf, sizeof( buf ), "%.10f %.10f %.10f", GetAbsAngles().x, GetAbsAngles().y, GetAbsAngles().z );
 				pRandomInternalOrgan->KeyValue( "angles", buf );
-				pRandomInternalOrgan->KeyValue( "model", "models/player/gibs/random_organ.mdl" );
+				if (TFGameRules()->IsMannVsMachineMode() && GetTeamNumber() == TF_TEAM_BLUE)
```
Tkain: Two things:

- Add `TFGameRules() &&` to the start of your 'if' condition to avoid dereferencing `TFGameRules()` if it's null.
- You should use `TF_TEAM_PVE_INVADERS` rather than `TF_TEAM_BLUE` in the context of MvM.

Bitl: I'll definitely resolve this when I get home

Bitl:
I have improved the implementation based on the code reviews.

Whurrhurr:
you should probably also make it so that if a mission is using EventPopfile Halloween that it spawns spleens instead of bolts (because the robots are turned into zombies)

src/game/server/tf/tf_player.cpp
```diff
@@ -9112,7 +9113,15 @@ int CTFPlayer::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 				pRandomInternalOrgan->KeyValue( "origin", buf );
 				Q_snprintf( buf, sizeof( buf ), "%.10f %.10f %.10f", GetAbsAngles().x, GetAbsAngles().y, GetAbsAngles().z );
 				pRandomInternalOrgan->KeyValue( "angles", buf );
-				pRandomInternalOrgan->KeyValue( "model", "models/player/gibs/random_organ.mdl" );
+				if (TFGameRules() && TFGameRules()->IsMannVsMachineMode() && GetTeamNumber() == TF_TEAM_PVE_INVADERS)
```
ficool2: Don't need the TFGameRules() null check, gamerules will always exist at this point. This whole function already doesn't do null checks on it in several places

As pointed out, you also need a check for `BloodColor() == DONT_BLEED` so zombie robots aren't affected

src/game/server/tf/tf_player.cpp
```diff
@@ -9112,7 +9113,15 @@ int CTFPlayer::OnTakeDamage( const CTakeDamageInfo &inputInfo )
...
+				{
+					//robots don't have spleens....
+					pRandomInternalOrgan->KeyValue("model", "models/player/gibs/gibs_bolt.mdl");
+				}
+				else
+				{
+					pRandomInternalOrgan->KeyValue("model", "models/player/gibs/random_organ.mdl");
+				}
 				pRandomInternalOrgan->KeyValue( "fademindist", "-1" );
```
ficool2, line 9126: Add spaces around the expressions to match Valve's styling

src/game/server/tf/tf_player.cpp
```diff
@@ -3072,6 +3072,7 @@ void CTFPlayer::PrecacheTFPlayer()
 	PrecacheModel( "effects/beam001_blu.vmt" );
 	PrecacheModel( "effects/beam001_white.vmt" );
 	PrecacheModel( "models/player/gibs/random_organ.mdl" );
+    PrecacheModel( "models/player/gibs/gibs_bolt.mdl" );
```
ficool2: Do this in `CTFPlayer::PrecacheMVM` instead

ficool2 (changes requested):

src/game/server/tf/tf_player.cpp
Tkain: (I'll defer to Ficool's advice; sorry for the unnecessary change request, Bitl!)

Bitl:
All of these have been added

src/game/server/tf/tf_player.cpp
```diff
@@ -2774,6 +2774,8 @@ void CTFPlayer::PrecacheMvM()
 	PrecacheModel( "models/items/currencypack_large.mdl" );
 
 	PrecacheModel( "models/bots/tw2/boss_bot/twcarrier_addon.mdl" );
+    
+    PrecacheModel( "models/player/gibs/gibs_bolt.mdl" );
```
FlaminSarge: Indentation should be changed to tabs to match rest of file

Bitl:
the indentation is fixed

src/game/server/tf/tf_player.cpp
```diff
@@ -9112,7 +9114,15 @@ int CTFPlayer::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 				pRandomInternalOrgan->KeyValue( "origin", buf );
 				Q_snprintf( buf, sizeof( buf ), "%.10f %.10f %.10f", GetAbsAngles().x, GetAbsAngles().y, GetAbsAngles().z );
 				pRandomInternalOrgan->KeyValue( "angles", buf );
-				pRandomInternalOrgan->KeyValue( "model", "models/player/gibs/random_organ.mdl" );
+				if (TFGameRules()->IsMannVsMachineMode() && GetTeamNumber() == TF_TEAM_PVE_INVADERS && BloodColor() == DONT_BLEED)
```
Generalisk: you can add ` && !TF_IsHolidayActive( kHoliday_Halloween )` to the end of that if statement to have the bots spawn spleens on halloween mode.

Bitl: there's a problem with this, this means that bots will produce spleens regardless if it's wave 666 when it's halloween. the correct way would be checking the population manager and see if it's using the MVM_EVENT_POPFILE_HALLOWEEN popfile type

Bitl:
this latest commit has also fixed a noticeable formatting problem

PR #1194, opened 2025-04-08
[ALL] Sync new SP bsp limits to MP.

Unusuario2:
# Description
In the Half-Life 2 anniversary update, many limits have been raised, which is especially useful for mappers since it gives them more freedom to create levels. The limits by today's standards are very low; this is not 2004, and people have much more powerful PCs, so increasing the BSP limits is not a threat. Note: CS:GO, Portal 2, and Garry's Mod already have these increased limits, so it won't cause instability while playing on servers.

```
Models 1024 -> 2048,
Brushes 8192 -> 16384, 
TexInfo 12288 -> 16384, 
TexData 2048 -> 8096, 
DispInfo 2048 -> 8096.
```

### DIFF
--- a/src/public/bspfile.h
+++ b/src/public/bspfile.h
@@ -57,12 +57,12 @@
 // Common limits
 // leaffaces, leafbrushes, planes, and verts are still bounded by
 // 16 bit short limits
-#define	MAX_MAP_MODELS					1024
-#define	MAX_MAP_BRUSHES					8192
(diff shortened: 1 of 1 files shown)


bruhhhxyt:
Does it really works modifying it via bspfile.h?

Unusuario2:
> Does it really works modifying it via bspfile.h?

yes, but all the engine binaries needs to be recompiled

PR #1195, opened 2025-04-10
Fix net_graph not rounding interp to nearest tick

treacherousfiend:
# Description
Currently, net_graph is the only place in the entire codebase where the player's cl_interp value is not rounded to the nearest tick instead displaying the raw value of `GetClientInterpAmount()`. This PR fixes that by adding `ROUND_TO_TICKS` to the code that displays it on net_graph.

Because I know that interp is a touchy subject inside the TF2 community and as such I'd like to cover my bases, I would like to clarify that this PR _DOES NOT_ change interpolation or cl_interp in any way shape or form. This is a consistency fix to make net_graph show the same client interp value that the rest of the game reads.
Over the years many people have experimented with interp values in attempting to find the best value for gameplay, and frankly I've seen quite a few people stand by interp values which are technically invalid, its only fair to show them what they're actually getting.
For further clarification, interp is calculated by taking the maximum value of cl_interp and cl_interp_ratio/cl_updaterate, and then that value is rounded to the nearest tick.

There are a couple odd little things that I also would like to mention:
the function that is used to convert interp in ms to ticks, `TIME_TO_TICKS` is `(int)( 0.5f + (float)(dt) / TICK_INTERVAL )`. `dt` is the input, and `TICK_INTERVAL` is set to 0.15, which notably is slightly off from how long 1 tick is supposed to take with an cmdrate/updaterate of 66. Because of this, players used to seeing numbers like 15.2ms on their net_graph will now see 15ms instead.
Also in case people who are not programmers are reading this, the `(int)` on the left of that equation converts the whole thing to an integer, or whole number, so you can't get a value like "1.5" ticks.
And lastly, the default settings of cl_interp 0.1 and cmdrate/updaterate of 20 is actually technically 105ms lerp, which I think is kind of funny
image

### DIFF
--- a/src/game/client/vgui_netgraphpanel.cpp
+++ b/src/game/client/vgui_netgraphpanel.cpp
@@ -768,10 +768,10 @@ void CNetGraphPanel::DrawTextFields( int graphvalue, int x, int y, int w, netban
 
 	g_pMatSystemSurface->DrawColoredText( font, x, y, GRAPH_RED, GRAPH_GREEN, GRAPH_BLUE, 255, "%s", sz );
 
-	Q_snprintf( sz, sizeof( sz ), "lerp: %5.1f ms", GetClientInterpAmount() * 1000.0f );
+	Q_snprintf( sz, sizeof( sz ), "lerp: %5.1f ms", ROUND_TO_TICKS( GetClientInterpAmount() ) * 1000.0f );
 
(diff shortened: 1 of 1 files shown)


wgetJane:
you will be torn limb from limb and splayed across the ground by a horde of spy mains

treacherousfiend:
I'll be honest I dreaded opening up my email today and seeing the response to this PR so I'm glad that yours was the only one

PR #1197, opened 2025-04-11
Fix: player at entity index 99 being treated as server

Pandaptable:
# Description
Currently, on 100 player servers, the player with entity index 99 is treated at dedicated server; this allows them to call Eternaween votes, their votekicks are displayed publicly, and they don't automatically vote yes (and by extension cannot vote on their own vote at all)
If the player count is ever increased to 255 (or higher) then the bug would come back.

### DIFF
--- a/src/game/shared/shareddefs.h
+++ b/src/game/shared/shareddefs.h
@@ -178,7 +178,7 @@ enum
 #define MAX_VOTE_DETAILS_LENGTH 64
 #define INVALID_ISSUE			-1
 #define MAX_VOTE_OPTIONS		5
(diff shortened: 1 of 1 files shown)


mastercoms:
Is there a reason DEDICATED_SERVER wasn't set to 0? This is the index used for chat messages, and can't ever be occupied by a player.

Pandaptable:
honestly? not sure, this was months ago and I'm not sure I ever tested that.

mastercoms:
`CNextLevelIssue::RequestCallVote` checks 99 rather than using DEDICATED_SERVER. Please correct this.

PR #1202, opened 2025-04-16
Add the ability to override weapon slot and position, with very basic TF2 UI support

FlaminSarge:
# Description
This PR adds the ability for vscript/mods/etc. to override the slot and position in which a weapon will appear in a player's weapon selection HUD for all games. It does this by adding `m_iSlot` and `m_iPosition` as networked properties on `CBaseCombatWeapon`, which are initialized to the values previously returned by `GetSlot()` and `GetPosition()`. These two methods now return those properties (and have been moved to be inlined in the header file; this can be reverted if needed). It also makes `m_iSubType` networked so that multiple weapons of the same class can be used by one player without any conflicts with `SelectItem()` behavior, if the vscript/mod so desires.
This PR also updates the TF2 weapon selection HUD to display the currently-selected weapon for the active slot if there are more than one weapon in that position.

If vscript/etc. do not touch these properties, all behavior remains as stock for all games.

Note that due to `tf_weapon_builder`'s unique setup, it is not affected by these changes (yet).

In the future, I plan to create additional UI handling for TF2 community mods to pull in, currently WIP at https://github.com/FlaminSarge/source-sdk-2013/tree/buckets (not yet rebased on this PR's changes).


FlaminSarge:
The intent behind this PR is to provide base functionality with semi-functional UI for VScript to start working with in the base (live) game.

Because GetSlot() and GetPosition() are strictly used for clientside weapon HUD selection, these new properties shouldn't break demos, as far as I can tell.

The followup changes to fully implement a custom UI for extra weapon positions is a longer-term goal that may or may not be viable to merge into live game.

gidi30:
This looks really good!
However it seems to be missing some code in CTFClientScoreBoardDialog and CTFHudPlayerClass to make it show the correct weapon model depending on whats equipped and not just the first weapon in the slot

FlaminSarge:
@gidi30 is that for the 3D player model in HUD? Good catch.

FlaminSarge:
> This looks really good! However it seems to be missing some code in CTFClientScoreBoardDialog and CTFHudPlayerClass to make it show the correct weapon model depending on whats equipped and not just the first weapon in the slot

On second thought, their usage of GetLoadoutSlot seems extremely deliberate for both cases: ```	int nItemSlot = ( pPlayer->IsAlive() && pPlayer->GetActiveTFWeapon() ) ? pPlayer->GetActiveTFWeapon()->GetAttributeContainer()->GetItem()->GetStaticData()->GetLoadoutSlot( nClass ) : LOADOUT_POSITION_PRIMARY;
...
	int nLoadoutSlot = pPlayer->GetActiveTFWeapon() ? pPlayer->GetActiveTFWeapon()->GetAttributeContainer()->GetItem()->GetStaticData()->GetLoadoutSlot( m_nClass ) : LOADOUT_POSITION_PRIMARY;

```

While I could replace this with just getting the active weapon, I'd like to find out the intent behind them sticking to loadout slots instead of just using the active weapon there...

FlaminSarge:
~~Noticing that for singleplayer purposes this may need to add some handling for the 3 props in save data, looking to add that now.~~
This seems to be handled properly already via the addition to the datadesc.

PR #1203, opened 2025-04-17
[TF2] Fix the weapon being invisible after the stun period expires

TheNotoriousFrogposter:
When a stun is applied to the player, the weapon is set to be invisible. But after the stun is over, there isn't any code to set the weapon back to the visible state. This PR fixes this.

Before:

https://github.com/user-attachments/assets/c056985f-caa5-45dd-8a4a-134889212ec9

After:

https://github.com/user-attachments/assets/8612d024-d737-4cc0-9d6a-82fc5bf37cdc

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -7361,6 +7361,11 @@ void CTFPlayerShared::OnRemoveStunned( void )
 #endif
 
 	m_pOuter->TeamFortress_SetSpeed();
+
+	if ( m_pOuter->GetActiveWeapon() && !InCond( TF_COND_TAUNTING ) && !InCond( TF_COND_HALLOWEEN_KART ) )
+	{
+		m_pOuter->GetActiveWeapon()->SetWeaponVisible( true );
+	}
 }
 
 //-----------------------------------------------------------------------------


HyperionCoding:
I could see this easily conflicting with other conditions or content creators wanting to set weapon invisible.

Bradasparky:
> I could see this easily conflicting with other conditions or content creators wanting to set weapon invisible.
This is true, however content creators get around this by setting weapons visibility on tick. See VSH as an example with Hale's fists.

Even so, these checks aren't sufficient enough to make sure you're setting weapon visibility as harmlessly and least often as possible. You should specifically be checking for `TF_STUN_CONTROLS` and `TF_STUN_LOSER_STATE` because those are the only two which affect weapon visibility. For example, getting slowed by a Sandman ball doesn't change weapon visibility. This code should also be moved into `ConditionThink` where you have access to the specific stuns before the condition is removed.

TheNotoriousFrogposter:
> > I could see this easily conflicting with other conditions or content creators wanting to set weapon invisible.
> 
> This is true, however content creators get around this by setting weapon visibility on tick. See VSH as an example with Hale's fists.
> 
> Even so, these checks aren't sufficient enough to make sure you're setting weapon visibility as harmlessly and least often as possible. You should specifically be checking for `TF_STUN_CONTROLS` and `TF_STUN_LOSER_STATE` because those are the only two which affect weapon visibility. For example, getting slowed by a Sandman ball doesn't change weapon visibility. This code should also be moved into `ConditionThink` where you have access to the specific stuns before the condition is removed.

I already though of that. While it is true, I can't think of a hypothetical scenario where it would cause any problem.

Bradasparky:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1203#issuecomment-2823601506

Coming back to this after I've tested my own suggestion, which I've found is unnecessarily complicated to implement and not worth making the `ConditionThink` stun code look uglier. Your implementation is fine and modders have work arounds for keeping weapons invis.

PR #1210, opened 2025-04-22
[VScript] Add ability to send UserMessages

horiuchii:
# Description
This PR adds the `CScriptUserMessage` class which allows VScript to write data to an instance of this class and then call a member function to execute the message inside.
This would allow scripters to call User Messages only called by certain entities like `game_text`, `game_text_tf`, `env_shake` without the need to spawn them or messages that are currently impossible to call with VScript like `ShowMenu`, `VGUIMenu` and various others.

# Usage
To execute a User Message in VScript, the user will call the global function `UserMessageBegin()` which takes a parameter for the message name.
After that, they can write any data they wish and then call the `Execute()` member function with a filter parameter.
This way, User Messages can be stored, modified and executed at any point in time.

# Example Script
```
local UserMessage = UserMessageBegin("KeyHintText")
UserMessage.WriteByte(1)
local UserMessageInstance = UserMessage

UserMessage.WriteString("%+attack% Press this key.")
UserMessageInstance.WriteString("%+attack2% Or this key.")

UserMessage.Execute({
	filter_type = RECIPIENT_FILTER_SINGLE_PLAYER
	filter_entity = GetPlayerFromUserID(2)
})

UserMessage.Execute({
	filter_type = RECIPIENT_FILTER_SINGLE_PLAYER
	filter_entity = GetPlayerFromUserID(3)
})

UserMessageInstance.Execute({
	filter_type = RECIPIENT_FILTER_SINGLE_PLAYER
	filter_entity = GetPlayerFromUserID(4)
})
```

### DIFF
--- a/src/game/server/vscript_server.cpp
+++ b/src/game/server/vscript_server.cpp
@@ -28,6 +28,7 @@
 #include "tier0/vcrmode.h"
 #include "in_buttons.h"
 #include "coordsize.h"
+#include "fmtstr.h"
 #include "team.h"
 
 #ifdef TF_DLL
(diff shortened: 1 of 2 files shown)


src/game/server/vscript_server.cpp
```diff
@@ -1506,6 +1507,208 @@ static void Script_EmitSoundEx( HSCRIPT params )
 	CBaseEntity::EmitSound( filter, pOutputEntity ? pOutputEntity->entindex() : -1, soundParams );
 }
 
+//-----------------------------------------------------------------------------
+HSCRIPT ScriptUserMessageBegin( const char* szMessageName )
+{
+	CScriptUserMessage* pUserMessage = new CScriptUserMessage( szMessageName );
```
ficool2, line 1513: Leaks memory if already registered

PR #1211, opened 2025-04-22
all games: Add sv_infinite_ammo ConVar

### LINKED ISSUE #3673: [HL2/CS:S] [Feature Request] sv_infinite_ammo or mp_infinite_ammo
mp_infinite_ammo was added to Counter-Strike 1.6
-Setting it to 1 makes it a bottomless clip.
-Setting it to 2 makes it infinite, but still reloads.

Considering this feature was also added to CS:GO, I was wondering if this feature could be ported over to Half-Life 2 and Counter-Strike: Source.

### LINKED ISSUE #3828: [TF2] [Feature Request] sv_infinite_ammo
Currently, there's only one option to have infinite ammo, `impulse 101`, however, it also refills clip.
In CS:GO, sv_infinite_ammo has 2 options, without reload and with reload (1 and 2 respectively). I suggest to port this to the game.

copperpixel:
This PR implements the `sv_infinite_ammo` cheat-protected ConVar from Left 4 Dead/Alien Swarm. When enabled, it prevents the player's active weapon from depleting ammo.
Furthermore, added functionality from Counter-Strike : GO where setting the ConVar to 2 or higher will only replenish the weapon's reserve (so the player still has to reload).

In Team Fortress 2, it also refills metal and weapon/item charges.

Partially closes https://github.com/ValveSoftware/Source-1-Games/issues/3673, closes https://github.com/ValveSoftware/Source-1-Games/issues/3828

### DIFF
--- a/src/game/shared/baseplayer_shared.cpp
+++ b/src/game/shared/baseplayer_shared.cpp
@@ -298,19 +298,22 @@ void CBasePlayer::ItemPostFrame()
 	{
 		CBaseCombatWeapon* pWeapon = GetActiveWeapon();
 
-		// dont give clip ammo if set higher than 1 -copperpixel
+		// only refill clip when sv_infinite_ammo == 1 -copperpixel
 		if( sv_infinite_ammo.GetInt() == 1 )
 		{
 #ifdef TF_DLL
-			CTFWeaponBase* pTFWeapon = static_cast< CTFWeaponBase* >( pWeapon );
-			if( pTFWeapon->IsEnergyWeapon() )
+			CTFWeaponBase* pTFWeapon = dynamic_cast< CTFWeaponBase* >( pWeapon );
+			if( pTFWeapon )
 			{
-				pTFWeapon->Energy_SetEnergy( pTFWeapon->Energy_GetMaxEnergy() );
-			}
-			else
-			{
-				pTFWeapon->m_iClip1 = pTFWeapon->GetMaxClip1();
-				pTFWeapon->m_iClip2 = pTFWeapon->GetMaxClip2();
+				if( pTFWeapon->IsEnergyWeapon() )
+				{
+					pTFWeapon->WeaponRegenerate();
+				}
+				else
+				{
+					pTFWeapon->m_iClip1 = pTFWeapon->GetMaxClip1();
+					pTFWeapon->m_iClip2 = pTFWeapon->GetMaxClip2();
+				}
 			}
 #else
 			pWeapon->m_iClip1 = pWeapon->GetMaxClip1();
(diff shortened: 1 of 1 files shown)


brokenparts (changes requested):
This doesn't seem to work with energy weapons in TF2

copperpixel:
> This doesn't seem to work with energy weapons in TF2

Should be fixed now

src/game/shared/baseplayer_shared.cpp
```diff
@@ -289,6 +293,65 @@ void CBasePlayer::ItemPostFrame()
...
+
+	if( sv_infinite_ammo.GetBool() && GetActiveWeapon() )
+	{
+		CBaseCombatWeapon* pWeapon = GetActiveWeapon();
+
+		// dont give clip ammo if set higher than 1 -copperpixel
+		if( sv_infinite_ammo.GetInt() == 1 )
+		{
+#ifdef TF_DLL
+			CTFWeaponBase* pTFWeapon = static_cast< CTFWeaponBase* >( pWeapon );
```
brokenparts: Should this be a `dynamic_cast`? I can't think of any reason why this wouldn't work, but `dynamic_cast` seems to be the preferred way to cast weapons in the rest of the code base.

src/game/shared/baseplayer_shared.cpp
```diff
@@ -289,6 +293,65 @@ void CBasePlayer::ItemPostFrame()
...
+			if( pTFWeapon->IsEnergyWeapon() )
+			{
+				pTFWeapon->Energy_SetEnergy( pTFWeapon->Energy_GetMaxEnergy() );
```
brokenparts: Nit pick, but `CTFWeaponBase::WeaponRegenerate` exists as a shorthand for this.

src/game/shared/baseplayer_shared.cpp
```diff
@@ -289,6 +293,65 @@ void CBasePlayer::ItemPostFrame()
...
+				true
+			);
+		}
+
+#ifdef TF_DLL
+		// tf: additionally replenish spy cloak, engineer metal, etc. -copperpixel
+		CTFPlayer* pTFPlayer = ToTFPlayer( this );
+		if( pTFPlayer )
+		{
+			pTFPlayer->GiveAmmo( 200, TF_AMMO_METAL, true );			// metal
```
brokenparts: `kAmmoSource_Resupply` should be specified to prevent the ammo amount from being multiplied by the `mult_metal_pickup` item attribute:
```c++
pTFPlayer->GiveAmmo( pTFPlayer->GetMaxAmmo( TF_AMMO_METAL ), TF_AMMO_METAL, true, kAmmoSource_Resupply );
```

src/game/shared/baseplayer_shared.cpp
```diff
@@ -289,6 +293,65 @@ void CBasePlayer::ItemPostFrame()
...
+			pTFPlayer->m_Shared.AddToSpyCloakMeter( 100.0f, true );		// spy cloak
+			pTFPlayer->AddToSpyKnife( 100.0f, true );					// spy-cicle recharge 
+			pTFPlayer->m_Shared.SetDemomanChargeMeter( 100.0f );		// shields charge
+			pTFPlayer->m_Shared.SetScoutEnergyDrinkMeter( 100.0f );		// bonk/crit-a-cola
```
brokenparts: Scout lunchbox items are covered by the ammo refill logic so this line is not needed.

brokenparts (changes requested):
This seems to work perfectly for guns.

If you want to refill Soldier banners and rage meters (Soda Popper, Phog, Hitman's Heatmaker, etc) the simpler way to do it would be by calling `CTFPlayerShared::SetItemChargeMeter` on every weapon slot. This is what `impulse 101` does:

https://github.com/ValveSoftware/source-sdk-2013/blob/0565403b153dfcde602f6f58d8f4d13483696a13/src/game/server/tf/tf_player.cpp#L14760-L14766

brokenparts:
That may have just been an issue with my testing setup, then.

ZzZombo:
Should infinite _ammo_ apply to charge meters? Non-weapons in general? Shouldn't that be handled by something separate?

copperpixel:
> Should infinite _ammo_ apply to charge meters? Non-weapons in general? Shouldn't that be handled by something separate?

I think in most use cases where this ConVar would be used in it might be preferred to refill charges as well.

Additionally, a lot of the charge meters already can be refilled with ammo pickups.

src/game/shared/baseplayer_shared.cpp
```diff
@@ -289,6 +293,72 @@ void CBasePlayer::ItemPostFrame()
...
+			CTFWeaponBase* pTFWeapon = dynamic_cast< CTFWeaponBase* >( pWeapon );
+			if( pTFWeapon )
+			{
+				if( pTFWeapon->IsEnergyWeapon() )
+				{
+					pTFWeapon->WeaponRegenerate();
+				}
+				else
+				{
+					pTFWeapon->m_iClip1 = pTFWeapon->GetMaxClip1();
```
Adrad1234, line 314:
```suggestion
					if ( !pTFWeapon->AutoFiresFullClip() )
					{
						pTFWeapon->m_iClip1 = pTFWeapon->GetMaxClip1();
					}
```
Check if the weapon is Beggar's Bazooka, and don't fill it's primary clip? Otherwise it fires non-stop, without being able to switch weapons.

PR #1212, opened 2025-04-23
[TF2] Improve ammo display on target ID

horiuchii:
# Description
This PR makes multiple changes to how ammo displays on the TargetID Display, which appears when healing a patient.

1. Ammo count no longer displays on weapons without ammo
2. Reserve ammo shows up next to the clip ammo when available

These changes also apply to spy's disguises and two new VScript functions `GetDisguiseAmmoReserveCount()` and `SetDisguiseAmmoReserveCount()` have been added to access the disguise reserve ammo.
Similarly to how ammo appears for disguise weapons, the reserve is also randomized when viewed this way.

If this gets merged, a new language token needs to be added to show weapons with reserve ammo:
`"TF_playerid_ammo_reserve"				"    %s1 / %s2"`

### DIFF
--- a/src/game/server/tf/tf_player.h
+++ b/src/game/server/tf/tf_player.h
@@ -747,43 +747,45 @@ public:
 	void				ScriptSetCondDuration( int nCond, float flNewDuration );
 	HSCRIPT				ScriptGetDisguiseTarget();
 
-	bool				ScriptIsCarryingRune()						{ return m_Shared.IsCarryingRune(); }
-	bool				ScriptIsCritBoosted( void ) const			{ return m_Shared.IsCritBoosted(); }
-	bool				ScriptIsInvulnerable( void ) const			{ return m_Shared.IsInvulnerable(); }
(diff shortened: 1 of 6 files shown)


src/game/server/tf/tf_player.h
```diff
@@ -747,43 +747,45 @@ class CTFPlayer : public CBaseMultiplayerPlayer, public IHasAttributes, public I
 	void				ScriptSetCondDuration( int nCond, float flNewDuration );
 	HSCRIPT				ScriptGetDisguiseTarget();
 
-	bool				ScriptIsCarryingRune()						{ return m_Shared.IsCarryingRune(); }
```
HyperionCoding, line 750: Just whitespace changes?

horiuchii: I added an additional tab spacing past the function declaration to make it look uniform with the two new functions I added, in hindsight this with probably unnecessary and creates confusion when trying to diff this.

PR #1215, opened 2025-04-26
[TF2] Fix medigun prediction errors

JustYourOrdinaryNormie:
When healing someone with the medigun, it will cause prediction errors to constantly appear on the medigun due to m_flChargeLevel having no tolerance at all


treacherousfiend:
This is kind of a dupe of #899? I'm mildly biased as the author of that PR but I don't think is a great fix since the issue isn't that it has no tolerance but rather that the client doesn't handle any charge rate modifications. Some tolerance probably would be good but its not really the issue imo.

treacherousfiend:
ah good catch! I wasn't aware of that.

PR #1216, opened 2025-04-26
[VScript] Fix leaked file handle in FileToString

x07x08:
Not closing the handle when files are bigger than 16384 bytes does not allow the file to be deleted until the server is closed.


Rushaway:
Thanks for fixing this massive leak. Need to see this merged asap.

PR #1218, opened 2025-04-27
[TF2] Always update Sentry m_flFireRate

horiuchii:
# Description
This PR moves the `m_flFireRate` check that resided in `Attack()` earlier to `SentryThink()` as this fixes two issues.
1. Unwrangling the Sentry would cause it to beep like a mini sentry due to not updating the m_flFireRate.
2. Using a Crit Boost Canteen would not make the sentry beep with a higher pitch as it should if it has a boosted fire rate.

### DIFF
--- a/src/game/server/tf/tf_obj_sentrygun.cpp
+++ b/src/game/server/tf/tf_obj_sentrygun.cpp
@@ -275,6 +275,21 @@ void CObjectSentrygun::SentryThink( void )
 		CALL_ATTRIB_HOOK_FLOAT_ON_OTHER( GetOwner(), m_flSentryRange, mult_sentry_range );
 	}
 
+	m_flFireRate = 1.f;
+	CALL_ATTRIB_HOOK_FLOAT_ON_OTHER( GetOwner(), m_flFireRate, mult_sentry_firerate );
+	if ( m_bPlayerControlled )
+	{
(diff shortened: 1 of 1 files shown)


src/game/server/tf/tf_obj_sentrygun.cpp
```diff
@@ -275,6 +275,21 @@ void CObjectSentrygun::SentryThink( void )
...
+	CALL_ATTRIB_HOOK_FLOAT_ON_OTHER( GetOwner(), m_flFireRate, mult_sentry_firerate );
+	if ( m_bPlayerControlled )
+	{
+		m_flFireRate *= 0.5f;
+	}
+	if ( IsMiniBuilding() && !IsDisposableBuilding() )
+	{
+		m_flFireRate *= 0.75f;
+	}
+	if ( GetBuilder() && GetBuilder()->m_Shared.InCond( TF_COND_CRITBOOSTED_USER_BUFF ) )
```
ZzZombo, line 288: You shouldn't call the same function with the same arguments twice in a row. Store the result in a variable.

horiuchii: All CBaseObject::GetBuilder() does is return m_hBuilder. Implementing what you're suggesting has no practical benefit.

PR #1220, opened 2025-04-27
[ALL] Resolve vscript naming conflict between CBaseCombatWeapon and CBaseEntity GetName by adding GetTargetname for CBaseEntity

FlaminSarge:
# Description
CBaseEntity defines vscript function GetName which returns the targetname. CBaseCombatWeapon defines vscript function GetName which returns the classname, overriding this. Workarounds are needed to get targetname for CBaseCombatWeapons.

This PR creates GetTargetname for CBaseEntity, while maintaining all existing behavior for GetName for both entities for backwards compatibility.


TeamSpen210:
I don’t think it makes sense to change `CBaseEntity` here. Entity names are far more prevalent (It’s very likely to be used somewhere in any slightly complex VScript), I don’t think it’s tenable to change that or imply that’s wrong.

FlaminSarge:
> I don’t think it makes sense to change `CBaseEntity` here. Entity names are far more prevalent (It’s very likely to be used somewhere in any slightly complex VScript), I don’t think it’s tenable to change that or imply that’s wrong.

Unfortunately, in order to maintain full backwards compatibility, `GetName()`'s behavior has to stay exactly the same here (returning classname for CBaseCombatWeapon and CBaseEntity) as we have no way of knowing where someone would be relying on the existing behavior.
If we enforce that 100% backwards compatibility, only remaining solution is to create a new function for CBaseEntity that returns Targetname at all times, to provide a substitute GetName's bugged behavior of returning classname on CBaseCombatWeapons (and to avoid the current direct m_iName workaround).

PR #1221, opened 2025-04-27
[ALL] Fix functionality of ent_setname command

FlaminSarge:
# Description
ent_setname's docs state that it accepts arguments like so:
`ent_setname {new name} {old name/class/use aim target if empty}`

However, the logic doesn't handle the 2nd parameter properly and tries to search by the first argument instead of the second. It also accepts input even if there are no arguments provided, where it should instead print the help text.

This PR fixes the above issues and restores functionality for the command.


FlaminSarge:
I just noticed that #1217 has some of these fixes, but is missing the ArgC handling/missing arguments fix.

PR #1228, opened 2025-05-01
[TF2] Hotfix for PR #893; Fix for submodel not updating correctly upon Casual late-join

rabscootle:
Hotfix for recently added fix; Added check for if the versus doors haven't been initialized yet before setting skin/submodel

Current code from PR #893 does not check if the versus door model has been initialized yet. This results in the submodel not updating correctly upon the first time it is loaded in client, resulting in the bugged appearance seen below:
image

Issue currently seems to resolve itself after the first time this is triggered (as the model has since been initialized).

Repro steps:
* Join a casual game in-progress from a freshly started TF2 client
* Play match until completion

# Description

### DIFF
--- a/src/game/client/tf/tf_hud_match_status.cpp
+++ b/src/game/client/tf/tf_hud_match_status.cpp
@@ -580,6 +580,12 @@ void CTFHudMatchStatus::FireGameEvent( IGameEvent * event )
 		int nSubModel = 0;
 		if (pMatchDesc->BGetRoundDoorParameters(nSkin, nSubModel))
 		{
+			// Is VS doors model not initialized yet?
+			if (m_pMatchStartModelPanel->m_hModel == NULL)
+			{
+				m_pMatchStartModelPanel->UpdateModel();
+			}
+
 			m_pMatchStartModelPanel->SetBodyGroup( "logos", nSubModel );
 			m_pMatchStartModelPanel->UpdateModel();
 			m_pMatchStartModelPanel->SetSkin( nSkin );


MrConeMan:
Listed in Source-1 games: https://github.com/ValveSoftware/Source-1-Games/issues/7236

rabscootle:
> @rabscootle im not sure, but the bug also introduces a crash for players who don't have the matchmaking elements activated. See: [ValveSoftware/Source-1-Games#7240](https://github.com/ValveSoftware/Source-1-Games/issues/7240)

This is something I’m hoping gets fixed with this additional check… but I’ll take another look.

programmedsun:
Client segfaults when the server goes for Map Voting wo matchmaking UI being active.

https://github.com/user-attachments/assets/22da2b3d-80b5-4075-9528-1787d0a1d4fd

When directly loading the matchmaking UI, the client no longer crashes and the original bug occurs, which this PR claims to fix.


https://github.com/user-attachments/assets/ff169770-843a-4eac-be20-e9c18de82606

programmedsun:
> live TF client

live TF client

rabscootle:
> Client segfaults when the server goes for Map Voting wo matchmaking UI being active.
> 
>  8mb.video-upN-Yd4QLIFe.mp4 
> When directly loading the matchmaking UI, the client no longer crashes and the original bug occurs, which this PR claims to fix.
> 
>  8mb.video-KOS-xgbxIq6w.mp4

I am unable to reproduce this in the live TF client.

image


https://github.com/user-attachments/assets/fc9a93e1-2efc-4adf-9556-a859bc98a135

programmedsun:
> Replying to [#1228 (comment)](https://github.com/ValveSoftware/source-sdk-2013/pull/1228#issuecomment-2849113476)

**outside-of-matchmaking** clients will crash, and this code seems to be the culprit (matchmaking doors doesnt exist for clients who dont have the matchmaking UI elements loaded in yet and **its not related to `ShouldUseMatchHUD()`**)

Since this issue is a bit exclusive to my servers (if u look hard enough, u will find them), but can also be replicated in Casual when Game Coordinator is offline.

rabscootle:
I don't think this would be triggered in Casual, as it's tied to the match information.

PR #1229, opened 2025-05-02
[TF2] Fix to use the correct "Say (PARTY) :" string (replace "chat_party" with "chat_say_party")

gaelcoral:
# Description
Fixes [ValveSoftware/Source-1-Games#4424](https://github.com/ValveSoftware/Source-1-Games/issues/4424#issue-1417031406).

For some reason, the `hud_basechat.cpp` file uses the string `chat_party` instead of the correct one, `chat_say_party`.
With this fix, the text should now display properly localized.

### Before:
image

### After:
image
_btw, unrelated, but in LatAm Spanish, it should be "Decir (GRUPO):" like the other strings, instead of "Informar (GRUPO) :".
However, since those strings come from Half-Life 2, which is part of a private Crowdin project I don’t have access to, I can’t submit a correction 🫠_


gaelcoral:
I changed the PR title to make it easier to understand in the release notes

PR #1230, opened 2025-05-02
Tf2: Fix strange shields count kills from other weapons

1556197:
Strange shields count shield kills, but also every other weapon kills. On wiki, mentioned as a bug.
This patch fixes this.

Fixes https://github.com/ValveSoftware/Source-1-Games/issues/5012. Detailed explanation can also be found there.

### DIFF
--- a/src/game/shared/tf/tf_gamerules.cpp
+++ b/src/game/shared/tf/tf_gamerules.cpp
@@ -10534,6 +10534,9 @@ void EconItemInterface_OnOwnerKillEaterEvent( IEconItemInterface *pEconEntity, C
 		if ( !pWearableItem->GetAttributeContainer() )
 			continue;
 
+		if ( !strcmp( pWearableItem->GetAttributeContainer()->GetItem()->GetStaticData()->GetItemClass(), "tf_wearable_demoshield" ) )
+			continue;
+
 		EconEntity_ValidateAndSendStrangeMessageToGC( pWearableItem->GetAttributeContainer()->GetItem(), pOwner, pVictim, eEventType, nIncrementValue );
 
 		if ( pWearableItem->GetAttributeContainer()->GetItem() == pEconEntity )


Platina6978:
I've noticed this odd behaviour ages ago but never really took it as a bug per say. I understand it as the shields being "meant" to go together with melee's and that's why they count kills from other items. Not entirely sure how it meshes with counting kills from primaries however.

On a side note; if this would get "fixed" now, the strange shields become much more difficult to rack strange kills on compared to before, creating a discrepancy between kills gained before and after this fix is added. A minor issue but one to consider regardless.

1556197:
> I've noticed this odd behaviour ages ago but never really took it as a bug per say. I understand it as the shields being "meant" to go together with melee's and that's why they count kills from other items. Not entirely sure how it meshes with counting kills from primaries however.

It makes no sense to count primary kills, as well as melee kills outside of charging. This is definitely not intended, because the sole reason it exists is because shields have item status "tf_wearable", like Mantreads, Gunboats or hats. Shields don't have weapon slot (in-game) and that's what broke them.

Speaking of Mantreads, they are exactly the same item (can kill and provide passive buffs), but they only count their own kills.

1556197:
> On a side note; if this would get "fixed" now, the strange shields become much more difficult to rack strange kills on compared to before, creating a discrepancy between kills gained before and after this fix is added. A minor issue but one to consider regardless.

I have maybe 30 kills with my Targe, but the counter shows 17639 kills. It doesn't even mean anything at this point, basically "gazillion" kills with I don't even know what weapon. And I still don't know how many shield kills I have, so it's useless as a strange weapon. If this gets fixed, I will gladly reset the counter.

src/game/shared/tf/tf_gamerules.cpp
```diff
@@ -10534,6 +10534,9 @@ void EconItemInterface_OnOwnerKillEaterEvent( IEconItemInterface *pEconEntity, C
 		if ( !pWearableItem->GetAttributeContainer() )
 			continue;
 
+		if ( !strcmp( pWearableItem->GetAttributeContainer()->GetItem()->GetStaticData()->GetItemClass(), "tf_wearable_demoshield" ) )
```
brokenparts: The rest of the code base uses `Q_strcmp` instead of the cstdlib function. `Q_strcmp` has null checks in debug mode.

PR #1235, opened 2025-05-03
[TF2] make disguised spies flinch and emit pain sounds when taking fall damage

kingofings:
# Description

Disguised spies do not emit any pain sounds nor flinch when taking fall damage. This is a dead giveaway that the player is a disguised spy. 

This PR makes sure spies emit the correct disguise pain sound to teammates and enemies.
In addition to also flinching the disguise health is also updated when taking fall damage.

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -15139,7 +15139,7 @@ void CTFPlayer::PainSound( const CTakeDamageInfo &info )
 		// don't play sound for fall stomp event
 		if ( !( pGround && pGround->IsPlayer() && m_Shared.CanFallStomp() ) )
 		{
-			TFPlayerClassData_t* pData = GetPlayerClass()->GetData();
+			TFPlayerClassData_t *pData = GetPlayerClass()->GetData();
 			if ( pData )
 			{
 				CPASFilter filter( GetAbsOrigin() );
@@ -15154,7 +15154,7 @@ void CTFPlayer::PainSound( const CTakeDamageInfo &info )
 
 			if ( m_Shared.InCond( TF_COND_DISGUISED ) )
 			{
-				TFPlayerClassData_t* pDisguiseData = GetPlayerClassData( m_Shared.GetDisguiseClass() );
+				TFPlayerClassData_t *pDisguiseData = GetPlayerClassData( m_Shared.GetDisguiseClass() );
 				if ( pDisguiseData )
 				{
 					CPASFilter disguisedFilter( GetAbsOrigin() );
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -9677,6 +9677,12 @@ int CTFPlayer::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 		{
 			m_Shared.NoteLastDamageTime( m_lastDamageAmount );
 		}
+
+		// Set our disguise health when taking falldamage to make it more believable
+		if ( m_Shared.InCond( TF_COND_DISGUISED ) && info.GetDamageType() == DMG_FALL)
+		{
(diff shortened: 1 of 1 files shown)


FlaminSarge:
Since this technically affects game balance, this may need a ConVar to enable/disable this behavior.

src/game/server/tf/tf_player.cpp
```diff
@@ -15129,7 +15137,7 @@ void CTFPlayer::PainSound( const CTakeDamageInfo &info )
 		return;
 
 	// no pain sounds while disguised, our man has supreme discipline
-	if ( m_Shared.InCond( TF_COND_DISGUISED ) )
+	if ( m_Shared.InCond( TF_COND_DISGUISED ) && !tf_disguise_mimic_fall_damage.GetBool())
```
FlaminSarge: nit: mimic valve's style with the spacing for the outer parentheses  `)`

src/game/server/tf/tf_player.cpp
```diff
@@ -15146,12 +15154,37 @@ void CTFPlayer::PainSound( const CTakeDamageInfo &info )
...
+					disguisedFilter.RemoveRecipientsByTeam( GetGlobalTFTeam( GetTeamNumber() ) );
+					disguisedFilter.AddRecipient( this );
+					EmitSound( disguisedFilter, entindex(), pDisguiseData->GetDeathSound( DEATH_SOUND_GENERIC ) );
+				}
 			}
 		}
 		return;
 	}
 
+	// no pain sounds while disguised, our man has supreme discipline
```
FlaminSarge, line 15181: why does this check happen twice? Would it be better (or at least more clear) to instead have a flag "we already played pain noise, skip the remaining logic" boolean field?

src/game/server/tf/tf_player.cpp
```diff
@@ -236,6 +236,8 @@ ConVar tf_halloween_giant_health_scale( "tf_halloween_giant_health_scale", "10",
 ConVar tf_grapplinghook_los_force_detach_time( "tf_grapplinghook_los_force_detach_time", "1", FCVAR_CHEAT );
 ConVar tf_powerup_max_charge_time( "tf_powerup_max_charge_time", "30", FCVAR_CHEAT );
 
+ConVar tf_disguise_mimic_fall_damage( "tf_disguise_mimic_fall_damage", "1", FCVAR_NOTIFY, "Should disguised spies mimic flinching and pain sounds when taking fall damage?", true, 0, true, 1 );
```
FlaminSarge: Default to current live game behavior 0 rather than 1, Valve will eventually figure out if it makes sense to globally enable for everyone.

src/game/server/tf/tf_player.cpp
kingofings: If the damage received was DMG_FALL it already returns early. The idea behind checking it a 2nd time was to catch if it was not DMG_FALL, since disguised spies shouldn't emit anything then.

FlaminSarge: You might be able to fold the first one into the DMG_FALL check then (and leave the second one intact):
```
	// fall damage and disguise fall damage enabled OR not disguised
	if ( info.GetDamageType() & DMG_FALL && ( !m_Shared.InCond( TF_COND_DISGUISED ) || tf_disguise_mimic_fall_damage.GetBool() ) )
	{
```

wgetJane:
not sure why this is disabled by default or even needs a cvar at all, i think disguised spies not flinching when taking fall damage is just an oversight, this should be labelled as a fix

FlaminSarge:
> i think disguised spies not flinching when taking fall damage is just an oversight

The comments referring to "our man has supreme discipline" seem to indicate it's intentional rather than an oversight. Realistically, it was probably because it was too much of a pain to implement originally, but still...

PR #1239, opened 2025-05-04
fix func_movelinear when parented

bottiger1:
fix func_movelinear when parented to a moving entity

# Description


Bitl:
Potential duplicate of #1054?

bottiger1:
Their version doesn't fix MoveDone

PR #1240, opened 2025-05-04
CLEINT_DLL??

CoolGuy-0000:
# Description


SirYodaJedi:
It's a typo, yes, but fixing it uncomments the code. Does introducing this new early return cause any unexpected side effects? If uncommenting this doesn't fix any bugs, I'd keep it commented out with `#ifdef 0` instead, since it's been commented out for presumably the past 18-ish years (give or take).

PR #1243, opened 2025-05-05
Allow building VBSP on Linux

clubby789:
# Description

This fixes things up to allow VBSP to build on Linux. The changes are generally
- Correcting includes for a case-sensitive filesystem.
- Updating threads/filesystem code to work with POSIX
- Adjust some code that isn't accepted by GCC (Most of these changes are taken from #179)

Currently a draft as I'm getting errors loading `filesystem_steam.so` `libtier0.so` without setting `LD_LIBRARY_PATH`


src/utils/vrad/incremental.cpp
```diff
@@ -310,11 +310,15 @@ void CIncremental::AddLightToFace(
...
 		EnterCriticalSection( &pLight->m_CS );
-			pFace = pLight->FindOrCreateLightFace( iFace, lmSize, &bNew );
+		pFace = pLight->FindOrCreateLightFace( iFace, lmSize, &bNew );
 		LeaveCriticalSection( &pLight->m_CS );
-
+#elif defined(POSIX)
+		pthread_mutex_lock( &pLight->m_CS );
+		pFace = pLight->FindOrCreateLightFace( iFace, lmSize, &bNew );
+		pthread_mutex_unlock( &pLight->m_CS );
+#endif
```
SanyaSho: You can replace this if/elif with CThreadMutex

clubby789:
Split the changes into more atomic commits to make reviewing easier.

src/utils/common/threads.cpp
```diff
@@ -13,14 +13,20 @@
 
 #define	USED
 
-#include <windows.h>
```
Generalisk: This should be encased in a `#ifdef WIN32` instead of being downright removed, as this causes build errors on Windows. This is a re-occurring issue throughout the whole PR.

PR #1245, opened 2025-05-06
[TF2] Add support for item_teamflag killfeed text customisability

main-thing:
Adds 3 keyvalues to item_teamflag:
- text_captured
- text_pickup
- text_defend

**Example of item_teamflag with these keyvalues:**
text_pickup  -> "has stolen gman's suitcase!"
text_defend  -> "has defended gman's intelligence!"

24162ea7a6c96e318149a202120f6d16
(as seen above, this PR **does not** update the max text limit set in hud_basedeathnotice.h from 32)

teamplay_flag_event in **modevents.res** needs to be edited to:
```
	"teamplay_flag_event"
	{
		"player"	"short"			// player this event involves
		"carrier"	"short"			// the carrier if needed
		"eventtype"	"short"			// pick up, capture, defend, dropped
		"home"		"byte"			// whether or not the flag was home (only set for TF_FLAGEVENT_PICKUP) 
		"team"		"byte"			// which team the flag belongs to
		"msg"		"string"		// the text which should override the default killfeed text
	}
```


game/mod_tf/downloadlists/ctf_2fort.lst
FlaminSarge, line 1: Don't commit this file.

game/mod_tf/maps/graphs/cg_facility.ain
FlaminSarge, line 1: Same, don't commit this file.

FlaminSarge (commented):
For localization purposes, this should probably be set up similar to how `teamplay_point_captured` is set up, where the flag's "name" is its own field and the various events handle how it actually gets displayed.

What is the use case for having different verbs in the text? How would those be localized?

main-thing:
> For localization purposes, this should probably be set up similar to how `teamplay_point_captured` is set up, where the flag's "name" is its own field and the various events handle how it actually gets displayed.
> 
> What is the use case for having different verbs in the text? How would those be localized?

I did not think of how maps (which are not added into the game by valve) would handle localization strings. As the code as it stands now does support localization strings as an input. Technically VScript could solve this by reading the cl_language ConVar from the client but that's an ugly workaround.

The use case I had in mind for having different verbs is allowing the teamplay_flag_event be used as a more general way to show messages in the killfeed to show important map goals being achieved.

FlaminSarge:
> The use case I had in mind for having different verbs is allowing the teamplay_flag_event be used as a more general way to show messages in the killfeed to show important map goals being achieved.

This may be better off as an entirely separate event used by VScript, then. Way more flexible too, and will not interfere with people that rely on flag capture events for other logic.

main-thing:
> > The use case I had in mind for having different verbs is allowing the teamplay_flag_event be used as a more general way to show messages in the killfeed to show important map goals being achieved.
> 
> This may be better off as an entirely separate event used by VScript, then. Way more flexible too, and will not interfere with people that rely on flag capture events for other logic.

Sounds good, I'll try to make a separate PR for that does just this. I'll try to get the flag to use the same system as the control points as you've stated above.

main-thing:
Converted the PR to a draft as it has been superseded by #1401

PR #1247, opened 2025-05-07
[TF2] Fix Pomson 6000 projectile firing below crosshair

juliarose:
# Description
The Pomson 6000 currently fires projectiles below the crosshair to align with the muzzle. This leads to shots hitting below their target and sometimes being blocked by ledges which can feel unintuitive and frustrating for the user. The preferred behavior is that the projectile fires straight from the crosshair's location.

Mentioned in Uncle Dane's video https://youtu.be/-iYUMyXLZoo?si=M_k6G6vxJo-TEhGo&t=668

Mentioned in https://github.com/ValveSoftware/Source-1-Games/issues/2592

> The shots fired by the weapon are fired slightly below the reticule.

# Changes
- Fixed the Pomson 6000 projectile firing below the crosshair.

https://github.com/user-attachments/assets/61923d92-908f-4f8b-a494-5e9e20459b13

https://github.com/user-attachments/assets/ad7bc8ad-c5d3-4879-9017-cd7c51c87e10

Unlike the Pomson 6000, the Rescue Ranger's bolt does not share this offset. It looks odd in third-person, but this is rarely noticed in gameplay and is preferred.

https://github.com/user-attachments/assets/d0b45e03-853e-41ce-9259-d8e9873fa98a

### DIFF
--- a/src/game/shared/tf/tf_weapon_raygun.cpp
+++ b/src/game/shared/tf/tf_weapon_raygun.cpp
@@ -241,15 +241,3 @@ void CTFDRGPomson::Precache()
 	PrecacheParticleSystem( "drg_pomson_projectile" );
 	PrecacheParticleSystem( "drg_pomson_muzzleflash" );
 }
-
-
-//-----------------------------------------------------------------------------
-// Purpose: 
-//-----------------------------------------------------------------------------
-void CTFDRGPomson::GetProjectileFireSetup( CTFPlayer *pPlayer, Vector vecOffset, Vector *vecSrc, QAngle *angForward, bool bHitTeammates, float flEndDist )
-{
-	BaseClass::GetProjectileFireSetup( pPlayer, vecOffset, vecSrc, angForward, bHitTeammates, flEndDist );
-
-	// adjust to line up with the weapon muzzle
-	vecSrc->z -= 13.0f;
-}
--- a/src/game/shared/tf/tf_weapon_raygun.h
+++ b/src/game/shared/tf/tf_weapon_raygun.h
@@ -95,8 +95,6 @@ public:
 
 	virtual const char *GetMuzzleFlashParticleEffect( void )	{ return "drg_pomson_muzzleflash"; }
 	virtual const char *GetIdleParticleEffect( void )			{ return "drg_pomson_idle"; }
-
-	virtual void GetProjectileFireSetup( CTFPlayer *pPlayer, Vector vecOffset, Vector *vecSrc, QAngle *angForward, bool bHitTeammates = true, float flEndDist = 2000.f );
 };
 
 


juliarose:
> This is the case with all the rocket launchers as well, I don't think it will be changed.

Except for the Original, which spawns from Soldier's face. A popular choice for this reason.

After more review, the Scorch Shot, Flare Gun, Crusader's Crossbow, Grenade Launcher, Iron Bomber, Loch-n-Load, Stickybomb Launcher, Quickiebomb Launcher, Scottish Resistance, and maybe a few other projectile weapons I haven't tested. None of the projectiles for these weapons are aligned with the muzzle. Medic's needle guns (Syringe Gun, Blutsauger, Overdose) spawn a bit lower and align with the muzzle but aren't really something anyone complains about considering the arch of these projectiles.

https://github.com/user-attachments/assets/4422a959-b1d8-459c-8592-644610c5af29

wgetJane:
the alignment is supposed to be in first-person

juliarose:
> the alignment is supposed to be in first-person

Yes.

I didn't include this earlier but if you check the commit there's a comment which mentions the offset was added to align with the muzzle. I'm not sure why the developers added this when every other projectile weapon doesn't function this way, aside from rocket launchers and needle guns it seems. It should be removed.

FlaminSarge:
> the alignment is supposed to be in first-person

Even in the original "intended" case it's looking like it's still not aligned with the muzzle, but rather with where the muzzle would be with viewmodel minmode enabled, so something's definitely not lining up here.
image

juliarose:
> Separately, what does the Bison do for this?

The class for the Pomson `CTFDRGPomson` inherits from the Bison `CTFRaygun` and explicitly adds an override for `GetProjectileFireSetup` to offset the projectile so that it spawns lower. The Bison has no such override, only the Pomson. Presumably the intention from the developers was to have the projectile align with the muzzle in third-person as the Engineer holds the weapon closer to his waist, while the Soldier holds the Bison like a pistol, staring down its sights. Although it should be aligned in first-person instead.

FlaminSarge:
> muzzle in third-person as the Engineer holds the weapon closer to his waist, while the Soldier holds the Bison like a pistol, staring down its sights. Although it should be aligned in first-person instead.

Makes sense. Would be worth trying out different offsets to see if you can get it to sort-of look right in both thirdperson and firstperson; the current offset is definitely too low, but firing from eye position may be too high (despite making a bit more sense for gameplay). Try -4 just to see how it looks?

mastercoms:
How about moving the extra offset to `CTFWeaponBaseGun::FireEnergyBall`?

```
	if ( GetWeaponID() == TF_WEAPON_DRG_POMSON )
	{
		vecOffset.z -= 13.0f;
	}
```

The problem is that unlike every other projectile, the offset is applied after projectile fire setup is determined. Projectile fire setup is done precisely for this reason, that it realigns the projectile path to travel along the crosshair, based upon the muzzle offset location. The weapon applying an additional offset after this causes the projectile to be moved downwards, with no such crosshair path realignment.

PR #1254, opened 2025-05-10
[ALL] Fixed a potential crash with entity damage modifiers

Bitl:
# Description

In TF2, damage modifiers are used (mainly on tf_weapon_medigun). However, in other games that don't use damage modifiers, there's a rare potential crash where it tries to find damage modifiers that don't exist. This PR attempts to fix that issue by checking if the damage modifier list is full before proceeding into CBaseEntity::GetAttackDamageScale and CBaseEntity::GetReceivedDamageScale respectively.

### DIFF
--- a/src/game/server/baseentity.cpp
+++ b/src/game/server/baseentity.cpp
@@ -1770,6 +1770,10 @@ int CBaseEntity::TakeDamage( const CTakeDamageInfo &inputInfo )
 float CBaseEntity::GetAttackDamageScale( CBaseEntity *pVictim )
 {
 	float flScale = 1;
+    
+	if ( m_DamageModifiers.Count() <= 0 )
+		return flScale;
+    
 	FOR_EACH_LL( m_DamageModifiers, i )
 	{
 		if ( !m_DamageModifiers[i]->IsDamageDoneToMe() )
@@ -1786,6 +1790,10 @@ float CBaseEntity::GetAttackDamageScale( CBaseEntity *pVictim )
 float CBaseEntity::GetReceivedDamageScale( CBaseEntity *pAttacker )
 {
 	float flScale = 1;
+    
+	if ( m_DamageModifiers.Count() <= 0 )
+		return flScale;
+    
 	FOR_EACH_LL( m_DamageModifiers, i )
 	{
 		if ( m_DamageModifiers[i]->IsDamageDoneToMe() )


brokenparts:
Just curious, do you have any more info on what causes this crash to occur? I'm wondering if there is an underlying issue with how `m_DamageModifiers` is populated.

Bitl:
This crash seems to be caused when there are no damage modifiers in the array (which is the default behavior on HL2DM). I found this crash when playtesting my own HL2 mod and that was the main conclusion I came to, as no other entity in HL2 or HL2DM based mods used damage modifiers.

Bitl:
Fixed another crash where the `info.ScaleDamage( GetReceivedDamageScale( info.GetAttacker() ) );` line was outside of the `if ( info.GetAttacker() )` block. Without this change, calling this line would crash the game if info.GetAttacker() returned NULL.

This change is related to the nature of the PR, so I'll keep it included here for consistency's sake.

src/game/server/baseentity.cpp
```diff
@@ -1697,11 +1697,11 @@ int CBaseEntity::TakeDamage( const CTakeDamageInfo &inputInfo )
 		if ( info.GetAttacker() )
 		{
 			info.ScaleDamage( info.GetAttacker()->GetAttackDamageScale( this ) );
+            
+			// Scale the damage by my own modifiers
+			info.ScaleDamage( GetReceivedDamageScale( info.GetAttacker() ) );
```
FlaminSarge: Moving this doesn't seem right; if the attacker is worldspawn/-1 the recipient should still take increased damage via their increased damage scale modifier. If it's crashing due to this, maybe pass an explicit NULL to the method if info.GetAttacker() isn't valid?

In addition, FOR_EACH_LL becomes:
```
	for( intp iteratorName=(listName).Head(); (listName).IsUtlLinkedList && iteratorName != (listName).InvalidIndex(); iteratorName = (listName).Next( iteratorName ) )
```
If m_damageModifiers is valid but empty, shouldn't this automatically skip the for loop? Where is it crashing if not? Is m_damageModifiers null, rather than empty?

Bitl: The move was done specifically because info.getattacker() could be a null value in some rare instances, and this blocks any potential crashes from happening in that instance

FlaminSarge: GetReceivedDamageScale doesn't even use its *pAttacker parameter, though. Was it info.getattacker() returning null, or was info itself null? If the latter, `info ? info.GetAttacker() : NULL` is probably the way to do this, instead of moving the call.

src/game/server/baseentity.cpp
```diff
@@ -1698,9 +1698,9 @@ int CBaseEntity::TakeDamage( const CTakeDamageInfo &inputInfo )
 		{
 			info.ScaleDamage( info.GetAttacker()->GetAttackDamageScale( this ) );
 		}
-
+		
 		// Scale the damage by my own modifiers
-		info.ScaleDamage( GetReceivedDamageScale( info.GetAttacker() ) );
+		info.ScaleDamage( GetReceivedDamageScale( info.GetAttacker() ? info.GetAttacker() : NULL ) );
```
FlaminSarge, line 1703: if you want to be extra safe, `info && info.GetAttacker() ? info.GetAttacker() : NULL`

src/game/server/baseentity.cpp
```diff
@@ -1698,9 +1698,9 @@ int CBaseEntity::TakeDamage( const CTakeDamageInfo &inputInfo )
 		{
 			info.ScaleDamage( info.GetAttacker()->GetAttackDamageScale( this ) );
 		}
-
+
```
FlaminSarge: remove the tabs?

src/game/server/baseentity.cpp
FlaminSarge: FYI something still doesn't seem right here, I don't see why info.GetAttacker() returning null would cause a crash but `info.GetAttacker() ? info.GetAttacker() : NULL` wouldn't, given that it'd just be `NULL ? NULL : NULL` in that case. Something else is going wrong, or that isn't actually what's crashing it.

src/game/server/baseentity.cpp
```diff
@@ -1770,6 +1770,10 @@ int CBaseEntity::TakeDamage( const CTakeDamageInfo &inputInfo )
 float CBaseEntity::GetAttackDamageScale( CBaseEntity *pVictim )
 {
 	float flScale = 1;
+    
+	if ( m_DamageModifiers.Count() <= 0 )
+		return flScale;
+
```
FlaminSarge, line 1776: Rather than an empty check, may be worth doing a null check here for m_DamageModifiers; I don't think the GetAttacker() change actually does anything, and I suspect your null pointer crash is coming from damagemodifiers being fully null instead. Same for the other case.

FlaminSarge: The other possibilities are that `m_DamageModifiers[i]` is null, or `i` is ending up out of bounds for `m_DamageModifiers`

Bitl:
closing this as i believe this is more of an issue with the mod i was working on rather than it being an issue with the base. this does not seem to happen on hl2dm.

PR #1256, opened 2025-05-10
[ALL] Add a command to clear chat history.

main-thing:
# Description
Adds a console command to allow players to clear chat history.

# Reasoning
After playing on populated server with active chat for a while, the chat history gets clogged with thousands of messages. 

Due to this, when opening chat your game will freeze for a short period of time. 

This freezing scales with how much messages are stored in chat. 

So, the simplest solution is to allow the chat history to be cleared by a console command.

### DIFF
--- a/src/game/client/hud_basechat.cpp
+++ b/src/game/client/hud_basechat.cpp
@@ -1870,3 +1870,12 @@ void CBaseHudChat::FireGameEvent( IGameEvent *event )
 	}
 #endif
 }
+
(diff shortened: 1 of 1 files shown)


BoogerKingLmao:
"Bro we gotta flee the country the TF2 chatlogs got leaked" but actually nice fix

wgetJane:
imo should be renamed from `cl_clearchathistory` to something like `clearchat`, since the former's naming scheme is more typical for convars than for concommands

PR #1261, opened 2025-05-12
[ALL] Improved vgui_fpspanel/cl_showfps.

Bitl:
# Description
This PR aims to improve the behavior of vgui_fpspanel.

This change does 5 things:

1. All ConVars used for the CFPSPanel class use FCVAR_ARCHIVE so it's saved upon restarting all games.
2. cl_showbattery is now shown in Competitive matchmaking. cl_showpos wasn't enabled because it could give players a potential advantage.
3. Fixed a bug where the cl_showbattery overlaps the cl_showpos display.
4. cl_showbattery will now display if cl_showfps and/or cl_showbattery are off.
5. Fixed some formatting issues in the file itself.

### DIFF
--- a/src/game/client/vgui_fpspanel.cpp
+++ b/src/game/client/vgui_fpspanel.cpp
@@ -24,9 +24,9 @@
 // memdbgon must be the last include file in a .cpp file!!!
 #include "tier0/memdbgon.h"
 
-static ConVar cl_showfps( "cl_showfps", "0", FCVAR_ALLOWED_IN_COMPETITIVE, "Draw fps meter at top of screen (1 = fps, 2 = smooth fps)" );
-static ConVar cl_showpos( "cl_showpos", "0", 0, "Draw current position at top of screen" );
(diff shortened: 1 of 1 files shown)


Bitl:
The most recent commit fixed an issue where there was an odd space between the FPS and the battery panels if both were disabled but cl_showpos wasn't.

Here's is is now if cl_showpos was disabled, but cl_showfps and cl_showbattery are enabled.
image

Here's another screenshot demonstrating the 4th change.

image

copperpixel:
I think it would be a good idea to add a ConVar that determines if the font should be proportional or not similarly to net_graph since the text can get hard to read on high resolutions

Bitl:
> I think it would be a good idea to add a ConVar that determines if the font should be proportional or not similarly to net_graph since the text can get hard to read on high resolutions

I'll definitely experiment with that

Bitl:
Implemented the proportional FPS panel fix!

image

PR #1269, opened 2025-05-13
[ALL] Refactor how kills with VPhysics objects are handled.

Bitl:
# Description
For whatever reason, VPhysicsShadowCollision in CBaseCombatCharacter doesn't set the owner entity properly when calculating collision damage between entities. This results in rare instances where entities aren't credited for the kill, and is presumably the reason why this block of code was written:

```
// FIXME: is there a better way for physics objects to keep track of what root entity responsible for them moving?
CBasePlayer* pPlayer = pOther->HasPhysicsAttacker(1.0);
if (pPlayer)
{
	dmgInfo.SetAttacker( pPlayer );
}
```

This PR correctly sets it so any VPhysics kills (mainly with the combine ball, but could effect other instances) will always mark the owner entity as the attacker, rather than the entity that attacked. This addresses the FIXME comment above, while also addressing other more rare instances (such as NPCs that fire insta-kill energy balls in some mods).


Bitl:
While testing this, I noticed that some VPhysics kills weren't registering properly, so the block of code was re-enabled

This PR now addresses rare instances of VPhysics collision kills.

PR #1270, opened 2025-05-14
[ALL] Fix major oversight with nav area visibility check in IVision::IsAbleToSee

CookieCat45:
# Description
This PR fixes an oversight in `IVision::IsAbleToSee()` with the way it determines visibility based on NavMesh areas. Normally, the function will check if there is a potential line of sight between the last known areas of the bot itself and the entity that it is trying to see. If there isn't, it will then save on performance by skipping the actual raytrace done to determine if there is actually any visible line to the target entity.

However, an issue arises if the target entity is not currently standing in an area. Since the function uses the last known areas of the bot and the target entity to determine visibility at first, it may attempt to use an area that the target is no longer standing in, as once the target isn't standing in the NavMesh anymore, it will use the last area that they were known to be in. This leads to situations where the visibility check will fail despite the target being in plain sight. If the target is standing on an area, and climbs up onto a tall prop that has no NavMesh on top of it, the visibility check will use the area that the target climbed onto the prop from. This can cause the visibility check to fail due to it being blocked by said prop.

A simple fix for this issue is to check if the target is actually standing in an area after doing the visibility check, to determine if the raytrace should still be skipped or not.

These videos show an example of the problem in TF2, on the map mvm_bigrock. Before the fix, the Sniper is able to climb up onto the rock (which has no NavMesh on top of it) to avoid being seen by bots. After the fix is applied, the bots are successfully able to see the Sniper that climbed onto the rock as soon as they come out of the gate.

## Before Fix

https://github.com/user-attachments/assets/8ec178ae-2b8e-469d-bb75-d6752474d49a

## After Fix

https://github.com/user-attachments/assets/2056601b-250a-4870-9cb5-ee33fd66fa4d


CookieCat45:
This also closes https://github.com/ValveSoftware/Source-1-Games/issues/3980

misyltoad:
I am curious as to what subject area is in this instance. I imagine that _should_ probably be NULL in this case?

CookieCat45:
> I am curious as to what subject area is in this instance. I imagine that _should_ probably be NULL in this case?

That would be whatever area the target entity was last standing in. If the entity leaves the nav mesh, it'll still be whatever nav area they were last standing on before they did, which is what causes the issue. The game updates it every think in [CBaseCombatCharacter::UpdateLastKnownArea](https://github.com/ValveSoftware/source-sdk-2013/blob/39f6dde8fbc238727c020d13b05ecadd31bda4c0/src/game/server/basecombatcharacter.cpp#L3457), but it doesn't appear set it to null if the entity isn't standing the nav mesh anymore.

mastercoms:
Some usages of `ClearLastKnownArea()` seem to imply that leaving the nav mesh should clear the last known area. However, at the same time, "LastKnownArea" in general implies this is a reference to the last known nav area the player was in, i.e, even if the player leaves, the AI can still use it to know where they last were before they left the nav mesh. Otherwise, it probably would have been called "

mastercoms:
Some usages of `ClearLastKnownArea()` seem to imply that leaving the nav mesh should clear the last known area. However, at the same time, "LastKnownArea" in general implies this is a reference to the last known nav area the player was in, i.e, even if the player leaves, the AI can still use it to know where they last were before they left the nav mesh. Otherwise, it probably would have been called "CurrentArea" or something like that. Plus, `UpdateLastKnownArea` explicitly bails if the area is null, rather than clearing the LastKnownArea. As far as usages for `GetLastKnownArea`, it's very unclear what the expectations are. Still, it seems like some usages are aware of this, for example `CNavMesh::GetNavArea( CBaseEntity *pEntity, int nFlags, float flBeneathLimit )` checking if the current position overlaps with the LastKnownArea.

I would then wager this is most likely an oversight with IsAbleToSee, as adjusted in the PR itself. It _may_ be beneficial to use a similar check as the aforementioned `GetNavArea` function, to null out subject area, like so:

```cpp
CNavArea *subjectArea = combat->GetLastKnownArea();
if ( subjectArea && !subjectArea->IsOverlapping( combat->GetAbsOrigin() ) )
{
  subjectArea = NULL;
}
```

This way, we catch any case in which the nav area may not be suitable for a visibility check, rather than only when the NavArea is null. Plus, this IsOverlapping check is much less expensive compared to the GetNavArea used in the PR.

CookieCat45:
> Some usages of `ClearLastKnownArea()` seem to imply that leaving the nav mesh should clear the last known area. However, at the same time, "LastKnownArea" in general implies this is a reference to the last known nav area the player was in, i.e, even if the player leaves, the AI can still use it to know where they last were before they left the nav mesh. Otherwise, it probably would have been called "CurrentArea" or something like that. Plus, `UpdateLastKnownArea` explicitly bails if the area is null, rather than clearing the LastKnownArea. As far as usages for `GetLastKnownArea`, it's very unclear what the expectations are. Still, it seems like some usages are aware of this, for example `CNavMesh::GetNavArea( CBaseEntity *pEntity, int nFlags, float flBeneathLimit )` checking if the current position overlaps with the LastKnownArea.
> 
> I would then wager this is most likely an oversight with `IsAbleToSee`, as adjusted in the PR itself. It _may_ be beneficial to use a similar check as the aforementioned `GetNavArea` function, to null out subject area, like so:
> 
> ```c++
> CNavArea *subjectArea = combat->GetLastKnownArea();
> if ( subjectArea && !subjectArea->IsOverlapping( combat->GetAbsOrigin() ) )
> {
>   subjectArea = NULL;
> }
> ```
> 
> This way, we catch any case in which the nav area may not be suitable for a visibility check, rather than only when the current NavArea is null. Plus, this IsOverlapping check is much less expensive compared to the GetNavArea used in the PR.


```
CBaseCombatCharacter *combat = subject->MyCombatCharacterPointer();
if ( combat )
{
    CNavArea *subjectArea = combat->GetLastKnownArea();
    if ( subjectArea && subjectArea->IsOverlapping( combat->GetAbsOrigin() ) )
    {
        CNavArea *myArea = GetBot()->GetEntity()->GetLastKnownArea();
        if ( myArea && myArea->IsOverlapping( GetBot()->GetEntity()->GetAbsOrigin() ) && !myArea->IsPotentiallyVisible( subjectArea ) )
        {
            // subject is not potentially visible, skip the expensive raycast
            return false;
        }
    }
}
```

In that case, I suppose this could work? The `IsOverlapping` check on the bot's area may be unnecessary, but it's there to account for the possibility of the bot itself also being outside of the nav mesh. Otherwise, this seems to have no functional differences from the previous fix after testing.

GodspotEnjoyer:
please dont do this please please please, keep the godspots please please please

Lyn3y:
please do this please please please, fix the godspots please please please

thanks you for fixing this idlespot!

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -589,16 +589,16 @@ bool IVision::IsAbleToSee( CBaseEntity *subject, FieldOfViewCheckType checkFOV,
...
-		CNavArea *myArea = GetBot()->GetEntity()->GetLastKnownArea();
-		if ( myArea && subjectArea )
-		{
-			if ( !myArea->IsPotentiallyVisible( subjectArea ) )
-			{
-				// subject is not potentially visible, skip the expensive raycast
-				return false;
-			}
-		}
+	    CNavArea *subjectArea = combat->GetLastKnownArea();
```
FlaminSarge: nit: Should match the tab style of SDK instead of 4 spaces

CookieCat45: fixed

CookieCat45:
I'm adding a Z height check here, as I realized that the subject could possibly manipulate their position by standing on the edge of a prop in order to trip the `IsOverlapping()` check. It might still happen with *really* short props, but I think 100 should suffice.

alkaris2:
I don't see why it needs fixing, let people abuse the exploit.

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -590,10 +590,12 @@ bool IVision::IsAbleToSee( CBaseEntity *subject, FieldOfViewCheckType checkFOV,
...
 	{
 		CNavArea *subjectArea = combat->GetLastKnownArea();
-		CNavArea *myArea = GetBot()->GetEntity()->GetLastKnownArea();
-		if ( myArea && subjectArea )
+		Vector subjectPos = combat->GetAbsOrigin();
+		if ( subjectArea && subjectArea->IsOverlapping( subjectPos ) && fabs( subjectPos.z - subjectArea->GetZ( subjectPos ) ) <= 100.0 )
 		{
-			if ( !myArea->IsPotentiallyVisible( subjectArea ) )
+			CNavArea *myArea = GetBot()->GetEntity()->GetLastKnownArea();
+			Vector myPos = GetBot()->GetEntity()->GetAbsOrigin();
```
ZzZombo: If I were you, I'd avoid these repeated `GetBot()->GetEntity()` calls by stuffing the result into a temporary variable.

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -590,10 +590,12 @@ bool IVision::IsAbleToSee( CBaseEntity *subject, FieldOfViewCheckType checkFOV,
 	if ( combat )
 	{
 		CNavArea *subjectArea = combat->GetLastKnownArea();
-		CNavArea *myArea = GetBot()->GetEntity()->GetLastKnownArea();
-		if ( myArea && subjectArea )
+		Vector subjectPos = combat->GetAbsOrigin();
+		if ( subjectArea && subjectArea->IsOverlapping( subjectPos ) && fabs( subjectPos.z - subjectArea->GetZ( subjectPos ) ) <= 100.0 )
```
ZzZombo, line 594: Why did you choose `100.0` here? Why not twice or half as much?

src/game/server/NextBot/NextBotVisionInterface.cpp
CookieCat45: I figured this wasn't really necessary since `GetBot()->GetEntity()` or similar is being used in many different places in the file.

src/game/server/NextBot/NextBotVisionInterface.cpp
CookieCat45: It's more of a middleground. I didn't choose a value that's too low so players don't unintentionally trigger it from jumping, but also not too high either so that it will still trigger for players that are really high above the ground.

src/game/server/NextBot/NextBotVisionInterface.cpp
ZzZombo: I bet they are not repeated within the same function, are they? The point was not to perform redundant work.

Generalisk: I suggest doing so as well as it'll reduce the amount of workload the server would have to perform, even if not by much, but could make quite A difference when in A server with dozens of bots (doesn't even have to be MvM bots, could be training bots as well), especially on servers ran on lower-end hardware.

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -593,12 +593,16 @@ bool IVision::IsAbleToSee( CBaseEntity *subject, FieldOfViewCheckType checkFOV,
 		Vector subjectPos = combat->GetAbsOrigin();
 		if ( subjectArea && subjectArea->IsOverlapping( subjectPos ) && fabs( subjectPos.z - subjectArea->GetZ( subjectPos ) ) <= 100.0 )
 		{
-			CNavArea *myArea = GetBot()->GetEntity()->GetLastKnownArea();
-			Vector myPos = GetBot()->GetEntity()->GetAbsOrigin();
-			if ( myArea && myArea->IsOverlapping( myPos ) && fabs( myPos.z - myArea->GetZ( myPos ) ) <= 100.0 && !myArea->IsPotentiallyVisible( subjectArea ) )
+			CBaseCombatCharacter* botEnt = GetBot()->GetEntity();
+			if (botEnt)
```
mastercoms: We already know botEnt is non-null, since code above this used `GetBot()->GetEntity()`. No need to check with an if statement.

CookieCat45: I suppose that's true, was putting it there just in case

PR #1271, opened 2025-05-14
[TF2] NPC damage notice fixes

Megawolf7339:
# Description
Adds a damage event to Halloween NPCs that didn't have them, meaning they give feedback with hitsounds/killsounds.

### DIFF
--- a/src/game/server/tf/bot_npc/bot_npc.cpp
+++ b/src/game/server/tf/bot_npc/bot_npc.cpp
@@ -526,7 +526,7 @@ int CBotNPC::OnTakeDamage_Alive( const CTakeDamageInfo &rawInfo )
 	if ( event )
 	{
 		event->SetInt( "entindex", entindex() );
-		event->SetInt( "health", MAX( 0, GetHealth() ) );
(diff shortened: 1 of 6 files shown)


Borloxos:
Do these changes also apply to the HHH, Merasmus, Monoculus, Skeleton King/Toastmaster, Scarecrows/Mobsters/Salmann (probably, as these are reskinned skeletons), and the RD robots? If not, could you add them to these entities as well?

Megawolf7339:
they apply to all bosses, including custom ones, though krampus is implemented in a very strange way where his health never reaches 0, so the kill sound may not work

Megawolf7339:
you were right i forgot those

Borloxos:
Nice, thank you! I cannot test your branch right now, does this only add damage sounds and kill sounds or are damage numbers also included?

gidi30:
> Nice, thank you! I cannot test your branch right now, does this only add damage sounds and kill sounds or are damage numbers also included?

Damage sounds and damage numbers are handled by the same hud element so both will happen

ZzZombo:
There seems to be some code duplication, can't you refactor the common code into functions?

PR #1273, opened 2025-05-14
[TF2] Add explosion creep

Megawolf7339:
# Description
Adds the 'tf_explosion_creep' ConVar, allowing explosives to pass over small edges, much like in CS:GO.

### DIFF
--- a/src/game/shared/tf/tf_gamerules.cpp
+++ b/src/game/shared/tf/tf_gamerules.cpp
@@ -5715,6 +5715,8 @@ void CTFRadiusDamageInfo::CalculateFalloff( void )
 	}
 }
 
+ConVar tf_explosion_creep( "tf_explosion_creep", "12" );
+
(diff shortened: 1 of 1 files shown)


wgetJane:
how does this interact with demo's sticky traps? unless im misunderstanding how this works, wouldn't the default value of 12 let the demoman kill people like this?

image

wgetJane:
imo explosion "creep" around corners is potentially problematic, but it's probably generally acceptable for explosions to "creep" over steps/slopes

something like this:
1) do LoS check from explosion point, if it's unobstructed then target should be hit
2) if LoS check is obstructed, then trace down by 18 units (step height)
3) if the downward trace hits nothing, then target should not be hit
4) otherwise, from the downward trace's endpos do another trace up by 18 units
5) from the upward trace's endpos, perform final LoS check to target

PR #1274, opened 2025-05-14
[TF2] Add optional colored names to in-game party chat

Megawolf7339:
# Description
Adds the 'tf_party_use_member_color' ConVar, allowing for names in chat to be colored with the player's respective party color.
image


ficool2:
I would say make this on by default as its a nice quality-of-life improvement

teufortressIndustries:
Aren't all party chat messages fully colored green?... correct me if I'm wrong.

PR #1277, opened 2025-05-14
all games: Add ConVars for configuring proximity voice chat

copperpixel:
# Description
Currently, proximity voice chat exists in the engine but it is hard-coded to be disabled. This PR adds ConVars allowing proximity voice chat to be configured by server admins or custom gamemodes. By default it is disabled.

| Name  | Default value | Description |
| ------------- | ------------- | ------------- |
| `sv_proximity_voice_enable` | `0` | Enable proximity voice chat |
| `sv_proximity_voice_distance` | `2048` | Max distance at which voice will be projected |

Note that the default value for the proximity voice distance has been cherry picked to be ideal for the fixed volume fade-out. This fade-out can't be changed without modifying the engine, but small maps might still want to mute voice at a lower distance either way.

[proximity_vc_convars.webm](https://github.com/user-attachments/assets/cbd05923-83b5-48d2-900e-e8cb4e53c9cb)

When proximity voice chat is enabled, the `proximity_voice` tag is appended to server's `sv_tags`.

### DIFF
--- a/src/.vscode/settings.json
+++ b/src/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "files.associations": {
-        "xiosbase": "cpp"
-    }
-}


Megawolf7339:
it might be a good idea to add this to the auto tag list in tf2

copperpixel:
> it might be a good idea to add this to the auto tag list in tf2

Implemented

PR #1282, opened 2025-05-14
IgnitePlayerEx to let players set burn duration in vscript

RobinsAviary:
# Description

This is a small change that fixes a bug in `IgnitePlayer`, a Vscript function that is part of `CTFPlayer`. In the original implementation the burn time is left to its default, resulting in the player to instantly start and stop burning. While this does trigger sound effects that could be useful to developers, it would also be nice to be able to directly set the duration, thus `IgnitePlayerEx` aims to fix this. It has a default burn time set of 5 seconds, but does still extinguish after a normal amount of time, should you set it long enough (This seems to be ~10-12 seconds).


FlaminSarge:
Does this differ from #805 ?

PR #1285, opened 2025-05-16
Fix class select animation playback

brokenparts:
Hi. This fixes a bug where class select animations will freeze if you go back and forth in the menu.

## Before:

https://github.com/user-attachments/assets/2a1c625e-ba25-4048-bf60-08cc4140d0f0

## After

https://github.com/user-attachments/assets/da9ab140-d697-4ad5-bdff-888d0d6e86db

## Details:

When reusing the shared CTFPlayerModelPanel in the class select screen, `SetupFlexWeights` would incorrectly calculate a large negative `m_flSceneTime`. This would result in class select animations freezing when going back and forth in the class select menu.

-0.1s start times are allowed to give the system time to lerp to the looping animation. This constant is already used in neighboring code.

The 2nd commit just replaces this `0.1f` magic number with a macro constant. This is optional, I just thought it would be a nice improvement.

### DIFF
--- a/src/game/client/tf/vgui/tf_playermodelpanel.cpp
+++ b/src/game/client/tf/vgui/tf_playermodelpanel.cpp
@@ -2295,6 +2295,7 @@ void CTFPlayerModelPanel::SetupFlexWeights( void )
 		}
 
 		m_flSceneTime += (m_RootMDL.m_MDL.m_flTime - m_flLastTickTime);
+		m_flSceneTime = Max( m_flSceneTime, -0.1f );
 		m_flLastTickTime = m_RootMDL.m_MDL.m_flTime;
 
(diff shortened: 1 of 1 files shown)


condor00fr:
Finally someone has adresse this! Is there any ways the sound effects could be re-introduced as well?

TheNotoriousFrogposter:
There's also another bug where the player models have their mouth wide open and then slowly decay to neutral expression for over a second when you select them.
agape

https://github.com/ValveSoftware/source-sdk-2013/blob/39f6dde8fbc238727c020d13b05ecadd31bda4c0/src/game/client/tf/vgui/tf_playermodelpanel.cpp#L2261
You could fix it by simply changing the number from 0.95 to 0.0f.

PR #1288, opened 2025-05-17
[ALL} Allow the Color class to accept hex code strings as colors.

Bitl:
# Description
This PR aims to implement a constructor to the Color class that allows users to define hex code strings for colors instead of RGB(a) values. This implementation is similar to hud_basechat's, and as such supports alpha channels.

Due to the nature of the Color class, the default alpha is set to 0, so passing in any normal hex code will result in a transparent color if whatever we put this in uses the full RGBA values.

# Usage:
This example defines a color macro named COLOR_GENUINE with a hex code of #4D7455, the TF2 Genuine color hex code provided by the TF2 Wiki (https://wiki.teamfortress.com/wiki/Item_quality). FF was added to ensure the color is opaque.
```
#define COLOR_GENUINE		Color("4D7455FF")
```

image

The COLOR_GENUINE color macro, but it is transparent
```
#define COLOR_GENUINE		Color("4D745580")
```

image

Shout out to @KaleiAlma for suggesting I do this 😄

### DIFF
--- a/src/public/Color.h
+++ b/src/public/Color.h
@@ -12,6 +12,8 @@
 #pragma once
 #endif
 
+#include "tier1/strtools.h"
+
 //-----------------------------------------------------------------------------
 // Purpose: Basic handler for an rgb set of colors
 //			This class is fully inline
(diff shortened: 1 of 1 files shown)


Bitl:
The newest version of this PR now supports wchar_t strings and also deals with syntax

The above example can now work with proper hex code strings!

```
#define COLOR_GENUINE		Color("#4D7455FF")
```

Bitl:
The latest version now adds KeyValues support, allowing for the use of hex values in KeyValues scripts!

Bitl:
Due to the new null pointer check, "4D7455FF" is no longer considered a valid value

The original post will be updated.

Bitl:
additional note: this modification doesn't work for schemes as the override for IScheme::GetColor needs to be modified to support it.

Bitl:
UPDATE: this PR now supports Color1 by the request of people in the Source Modding Community. You may now enter in the above value as the following:

```
#define COLOR_GENUINE		Color(0.3, 0.45, 0.33)
```

Bitl:
Fixed an issue where 3 digit hex codes didn't work as expected.

Bitl:
Currently, Color1 values don't work with the current keyvalues implementation and would require a rewrite of how colors are handled. i'll probably be leaving it alone in case this gets implemented as it is

Issue #1226, opened 2025-05-01
[TF2] MvM mini boss backgrounds don't work on mission support bots


Mentrillum:
The May 1st, 2025 patch added the red background for support bots if they're classified as a giant (or MiniBoss). However if the bot is marked as a mission support, such as a spy, sniper, sentry buster, or engineer, the wavebar will not use the red background on those bots as shown in the screenshot.

![Image](https://github.com/user-attachments/assets/cea7b00d-1446-4aa9-b6c8-7d4585d5c228)

Here is a basic POP file you can use to see this issue, I put this on mvm_bigrock. As of writing I can't check for myself to see this issue since the SDK isn't updated to include this change.

```
#base robot_giant.pop
#base robot_standard.pop

WaveSchedule
{
	StartingCurrency		400
	RespawnWaveTime 		7
	CanBotsAttackWhileInSpawnRoom no

	Mission // Spy
	{
		Objective Spy  

		InitialCooldown 10
		Where spawnbot_mission_spy
		BeginAtWave 1
		RunForThisManyWaves 1
		CooldownTime 30
		DesiredCount 4

		TFBot
		{
			Class Spy
			Skill Expert
			Name Spy
			Health 1200
			Attributes MiniBoss
		}
	}
	
	Mission // Sniper
	{
		Objective Sniper

		InitialCooldown 10
		Where spawnbot_mission_sniper
		BeginAtWave 1
		RunForThisManyWaves 1
		CooldownTime 20
		DesiredCount 4

		TFBot
		{
			Class Sniper
			Skill Hard
			Name Sniper
			MaxVisionRange 3000
			Health 1200
			Attributes MiniBoss
		}
	}
	
	Mission // Engineer
	{
		Objective Engineer  

		InitialCooldown 10
		Where spawnbot
		BeginAtWave 1
		RunForThisManyWaves 1
		CooldownTime 60
		DesiredCount 1

		TFBot
		{
			Template T_TFBot_Engineer_Sentry_Teleporter
			Health 1200
			Attributes MiniBoss
		}
	}
	
	Wave
	{
		StartWaveOutput
		{
			Target wave_start_relay
			Action Trigger
		}
		DoneOutput					
		{
			Target wave_finished_relay
			Action trigger
		}
		WaitWhenDone 65
		Checkpoint Yes
		
		WaveSpawn
		{
			Where spawnbot
			TotalCount 1
			MaxActive 1
			SpawnCount 1
[...]

PR #1289, opened 2025-05-17
[TF2] 2 MvM Wavebar Changes

Mentrillum:
This PR has two changes relating to the MvM wavebars.

The first change addresses issue #1226 where the change relating to red icon backgrounds didn't carry over to mission supports even though they were added for regular support bots.

image

The second change allows MvM mission creators to utilize their own tank icons, removing the hardcoded bits that forced the default tank icon, previously you needed an extension to get custom tank icons working and it wouldn't work for the boss bars. This also fixes a bug where if a wave uses the tank icon but doesn't have a tank would play the tank wave music on wave start, now it just checks if the wave has any tank spawns.

image (1)

### DIFF
--- a/src/game/client/tf/player_vs_environment/c_tf_tank_boss.cpp
+++ b/src/game/client/tf/player_vs_environment/c_tf_tank_boss.cpp
@@ -8,6 +8,7 @@
 
 IMPLEMENT_CLIENTCLASS_DT(C_TFTankBoss, DT_TFTankBoss, CTFTankBoss)
 	//RecvPropVector(RECVINFO(m_shadowDirection)),
+	RecvPropString( RECVINFO(m_iszClassIcon) ),
 END_RECV_TABLE()
 
 LINK_ENTITY_TO_CLASS( tank_boss, C_TFTankBoss );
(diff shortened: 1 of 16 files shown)


StardustSpie:
updoot

SargentoMesa:
Stuff like this, mission makers would rejoice if merged, especially for the upcoming mvm thing

src/game/client/tf/player_vs_environment/c_tf_tank_boss.cpp
```diff
@@ -22,3 +23,11 @@ void C_TFTankBoss::GetGlowEffectColor( float *r, float *g, float *b )
 	TeamplayRoundBasedRules()->GetTeamGlowColor( GetTeamNumber(), *r, *g, *b );
 }
 
+const char* C_TFTankBoss::GetBossProgressImageName() const
+{
+	if (m_iszClassIcon[0] == '\0')
```
ZzZombo, line 28: Why do you check the same string for `null` in two different ways in this PR?

src/game/server/tf/player_vs_environment/tf_populator_spawners.cpp
```diff
@@ -1467,6 +1473,15 @@ bool CTankSpawner::Spawn( const Vector &here, EntityHandleVector_t *result )
 	return false;
 }
 
+string_t CTankSpawner::GetClassIcon(int nSpawnNum)
+{
+	if (m_iszClassIcon != NULL_STRING)
```
ZzZombo, line 1478: Like here.

Professor-Heavy:
> Stuff like this, mission makers would rejoice if merged, especially for the upcoming mvm thing

Commence the rejoicing.

JoriKos:
I would like to point out that these commit messages should be more descriptive in the future. 'Init' and 'fixes' doesn't say anything about what was actually changed.

Mentrillum:
> I would like to point out that these commit messages should be more descriptive in the future. 'Init' and 'fixes' doesn't say anything about what was actually changed.

For the record I always forget that when PRs push, they use the commit name and not the actual PR name.

PR #1296, opened 2025-05-20
[TF2] Fix players sometimes getting stuck in team unassigned when joining a server

bakugo:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/3651

If a player connects to a server and tries to join a specific team right as the team becomes full (such that the client-side team balance check passes but the server-side one fails), then tries again within 2 seconds, the second `jointeam` command will be rejected by the team change rate limit without re-opening the team selection panel. This results in the player becoming permanently stuck in team unassigned unless they use console commands to manually join a team, as the team selection panel cannot be reopened.

Fixed by skipping the team change rate limit check if the player is still in team unassigned.

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -7332,7 +7332,7 @@ bool CTFPlayer::ClientCommand( const CCommand &args )
 	if ( FStrEq( pcmd, "jointeam" ) )
 	{
 		// don't let them spam the server with changes
-		if ( GetNextChangeTeamTime() > gpGlobals->curtime )
+		if ( GetNextChangeTeamTime() > gpGlobals->curtime && GetTeamNumber() != TEAM_UNASSIGNED )
 			return true;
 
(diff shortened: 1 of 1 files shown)


ZzZombo:
This may open a possibility of abuse if a malicious client repeatedly joins and leaves the unassigned team, but I don't know if it is possible to join it after changing teams once in TF2.

bakugo:
> This may open a possibility of abuse if a malicious client repeatedly joins and leaves the unassigned team, but I don't know if it is possible to join it after changing teams once in TF2.

I checked before opening this PR. As far as I can tell, there is no way to rejoin team unassigned after leaving it.

PocketSpade:
Does this help with malicious players / bots staying on unassigned to avoid being kicked and clogging up game servers?

bakugo:
> Does this help with malicious players / bots staying on unassigned to avoid being kicked and clogging up game servers?

No, this does not affect that.

PR #1301, opened 2025-05-21
[TF2] Uncomment Attribute Code for ammo_becomes_health to Make it Function Again

antielectronbomb:
The "ammo_becomes_health" attribute currently does not function as intended because the underlying code for said attribute has been commented out.

By uncommenting the code for the said attribute, this should make the attribute to function properly again and allow TF2 modders to have another useful attribute that can be used with VScript or SourceMod.


Professor-Heavy:
I think there are plenty of other attributes that we could uncomment as well. I can't list them off in their entirety, but while going through the code, I see things like `addperc_ondmgdone_tmpbuff`, `disguise_speed_penalty`, `set_charged_airblast`, etc.

PR #1302, opened 2025-05-21
[TF2] Implement "mod_medic_buff_range" attribute

SanyaSho:
# Description
This is a copy of "mod_soldier_buff_range" attribute but for Medic which is might be useful for mapmakers and custom servers.

## Default:
https://github.com/user-attachments/assets/4b901516-2ab7-4873-aa66-dbb9351a39cd
## Modified (attribute value is set to 0.4):
https://github.com/user-attachments/assets/8e34dcbc-145e-4f3b-8b59-ee629c0bf203

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -14187,6 +14187,10 @@ void CTFPlayerShared::PulseMedicRadiusHeal( void )
 #ifdef GAME_DLL
 	if ( gpGlobals->curtime >= m_flRadiusHealCheckTime )
 	{
+		float fMaxRadius = 450.0f;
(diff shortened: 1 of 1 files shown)


FlaminSarge:
Will this also apply to the set_weapon_mode 1 fists taunt handling? e.g. https://gist.github.com/sigsegv-mvm/f72e3047c58dc2674493#file-set_weapon_mode-txt-L67

Separately, any thoughts on an attribute that scales the range at which a Medigun can heal targets?

SanyaSho:
> Will this also apply to the set_weapon_mode 1 fists taunt handling?

No, the radial buff fists are using [it's own code](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/game/server/tf/tf_player.cpp#L18576-L18603), which is also using hardcoded range values.

> Separately, any thoughts on an attribute that scales the range at which a Medigun can heal targets?

Should be easy to implement, the code is looking in the weapon info table for "range" which we can modify using the attribute.

PR #1305, opened 2025-05-23
[TF2] Set the class loadout player model team based on current team if available, add setteam2/3 commands for class loadout panel for custom huds to add Team buttons

FlaminSarge:
# Description
TF2 currently defaults to RED skins in the class loadout at all times other than when selecting the "Edit Loadout" button on the bottom-right of the class selection menu. This PR allows the loadout to take on the player's current class if they are in a game.
This also allows custom HUDs to use the commands "setteam2" and "setteam3" instead of the current "sv_cheats 1; r_skin 0/1" to set the playermodel's team.


GrubPL:
image
image
it doesnt seem to work with most team colored weapons, but only when in the main menu, while in game it seemed to work fine

just thought you'd want to know this, and potentially fix it if possible

PR #1307, opened 2025-05-25
[TF2] Fix #618 causing money to be aggressively collected in the playable area

mtxfellen:
# Description

The PR #618, merged in 52e568b374ef3dd9f9309b8aa9f1fba7b0a3dbc0 happened to introduce a bug in `CCurrencyPack::ComeToRest()` that this PR aims to address:

- Use `PointIsWithin()` instead of `WorldSpaceSurroundingBounds()` so that complex trigger shapes are accounted for.
  - Using `WorldSpaceSurroundingBounds` effectively ignores the actual shape of the spawnroom and does an AABB check, which is particularly problematic for maps like Mannhattan that have multiple spawnroom areas controlled by a single entity with a complex collision shape, instead of multiple distinct entities.
  - As a result, cash ends up being auto-collected in perfectly accessible, playable space on pre-existing maps.

This PR also fixes some other minor bugs in this method:
- ~~Add missing early returns that could've rarely allowed currency to be auto-collected multiple times.~~ Fixed in d757e9117d0bb1afebe60e0fd665465523cb57a5.
- Prevent money being collected in disabled respawn rooms.
- Move `m_bDistributed` check so that red money can be auto-collected if it drops outside the playable space.
- Normalise `GetCollisionOrigin()` and `GetAbsOrigin()` usage as `GetAbsOrigin()`.
  - They are the same for this entity.


src/game/server/tf/entity_currencypack.cpp
```diff
@@ -174,59 +174,59 @@ void CCurrencyPack::BlinkThink( void )
...
 //-----------------------------------------------------------------------------
-// Become touchable when we are at rest
+// Purpose: Collect this pack when it comes to rest if it is outside the playable space.
 //-----------------------------------------------------------------------------
 void CCurrencyPack::ComeToRest( void )
 {
 	BaseClass::ComeToRest();
 
-	if ( IsClaimed() || m_bDistributed )
+	// I'm not when this should ever actually return true, but it doesn't seem to cause any issues so I'm not touching it.
```
brokenparts:
```suggestion
	// I'm not sure when this should ever actually return true, but it doesn't seem to cause any issues so I'm not touching it.
```

src/game/server/tf/entity_currencypack.cpp
```diff
@@ -174,59 +174,59 @@ void CCurrencyPack::BlinkThink( void )
...
 	}
 
 	// Or a func_respawnroom (robots can drop money in their own spawn)
 	for ( int i = 0; i < IFuncRespawnRoomAutoList::AutoList().Count(); i++ )
 	{
 		CFuncRespawnRoom *pRespawnRoom = static_cast<CFuncRespawnRoom *>( IFuncRespawnRoomAutoList::AutoList()[ i ] );
-		Vector vecMins, vecMaxs;
-		pRespawnRoom->GetCollideable()->WorldSpaceSurroundingBounds( &vecMins, &vecMaxs );
-		if ( IsPointInBox( GetCollideable()->GetCollisionOrigin(), vecMins, vecMaxs ) )
+		if ( !pRespawnRoom->m_bDisabled && pRespawnRoom->PointIsWithin( GetAbsOrigin() ) )
```
FlaminSarge: If the respawn room gets disabled before the money comes to rest, will this result in the money not autocollecting? Some maps may disable certain respawn rooms for robots mid-game but may not make those areas walkable for players.

mtxfellen: There's little reason for a map to do this in an inaccessible area (where the problem would arise), as the enabled/disabled status of a func_respawnroom entity doesn't control whether the containing spawnpoints can be targeted by the populator.

Mentrillum:
```suggestion
		if ( !pRespawnRoom->m_bDisabled && ( pRespawnRoom->PointIsWithin( GetAbsOrigin() ) || pRespawnRoom->IsTouching( this ) )
```
Correct me if I'm wrong but couldn't we also double check to see if the cash entity is touching the trigger using IsTouching?

mtxfellen: `IsTouching()` would require the spawnroom to have spawnflags set that allow cash entities to register as touching to return true, but they usually only have `SF_TRIGGER_ALLOW_CLIENTS` set.

PR #1314, opened 2025-05-27
[TF2] fix: remove early return in coach request check

CosminPerRam:
# Description
Early returning `false` skips the threshold of hours check and because of the previous checks, makes the function always return `false` unconditionally.


CosminPerRam:
Possible indeed, an official clarification if this is intentional would be nice to have to add a documentation comment instead.

PR #1319, opened 2025-05-31
[TF2] Fix various CYOAPDA/ConTracker animation bugs ("iPad cancelling")

jus7n:
This patch fixes a bug in `CTFPlayerAnimState::CheckCYOAPDAAnimtion()` that prevented the server from ever playing ConTracker animations which caused desync between client and server hitboxes.

Here's a video showing some silly effects of the bug: https://youtu.be/Qgl0uRBl9_E

Before patch (server hitbox does not match client's):
image

After patch:
image

Additional changes:

- Update checks that prevented certain player actions while viewing the ConTracker to instead check for actively playing ConTracker animations. This fixes wonky behavior like players being able to move/shoot while viewing the ConTracker.

- Add new checks to prevent player input while viewing the ConTracker.

- Remove client check preventing the quest panel from being opened while taunting.

- Players are now put in third person to make it obvious when ConTracker animations are playing (and thus are unable to move/shoot).

- Add ConVar `tf_cyoa_pda_animations` to control ConTracker animations.

- Add `IsAllowedToViewCYOAPDA()` and `IsInCYOAPDAAnimation()` script functions.

### DIFF
--- a/src/game/server/tf/tf_player.h
+++ b/src/game/server/tf/tf_player.h
@@ -748,6 +748,7 @@ public:
 	bool				IsTaunting( void ) const { return m_Shared.InCond( TF_COND_TAUNTING ); }
 	void				DoTauntAttack( void );
 	bool				IsAllowedToTaunt( void );
+	bool                IsAllowedToViewCYOAPDA( void );
 	bool				FindOpenTauntPartnerPosition( const CEconItemView *pEconItemView, Vector &position, float *flTolerance );
 	bool				IsAllowedToInitiateTauntWithPartner( const CEconItemView *pEconItemView, char *pszErrorMessage = NULL, int cubErrorMessage = 0 );
 	void				CancelTaunt( void );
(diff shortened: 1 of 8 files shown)


src/game/server/tf/tf_player.h
```diff
@@ -748,6 +748,7 @@ class CTFPlayer : public CBaseMultiplayerPlayer, public IHasAttributes, public I
 	bool				IsTaunting( void ) const { return m_Shared.InCond( TF_COND_TAUNTING ); }
 	void				DoTauntAttack( void );
 	bool				IsAllowedToTaunt( void );
+	bool                IsAllowedToViewCYOAPDA( void );
```
skmedix, line 751:
```suggestion
	bool				IsAllowedToViewCYOAPDA( void );
```

src/game/shared/tf/tf_player_shared.cpp
```diff
@@ -762,6 +764,71 @@ bool CTFPlayer::IsAllowedToTaunt( void )
 	return true;
 }
 
+//-----------------------------------------------------------------------------
+// Purpose: Check if a player is allowed to view the ConTracker. Mostly borrowed from IsAllowedToTaunt
```
henke37, line 768: "Borrowed" as in copy pasted. Factor the code out into a shared function!

PR #1323, opened 2025-06-01
add func_detail_nonsolid

RealGamerX:
Currently, if you want to add a non-solid brush to your map, your only option is to use a brush entity, such as func_illusionary or func_brush. Brush entities take up valuable edicts, and are generally overkill when all you want is a brush you don't collide with.

This PR adds func_detail_nonsolid, which, just like func_detail, gets turned into a world brush on compile, thus not consuming any edicts. It basically acts like the %compilenonsolid VMT property.

Just like func_detail, func_detail_nonsolid has no keyvalues.

image

Here's the entity in action:

https://github.com/user-attachments/assets/4d6b0cf8-9eba-49c0-8db6-515af67ef741

Huge thanks to @gidi30 and @flurbury for helping me with stopping windows from breaking the entity.

### DIFF
--- a/src/utils/common/map_shared.cpp
+++ b/src/utils/common/map_shared.cpp
@@ -29,7 +29,7 @@ ChunkFileResult_t LoadEntityKeyCallback(const char *szKey, const char *szValue, 
 		{
 			pLoadEntity->nBaseContents = CONTENTS_DETAIL;
 		}
-		else if (!stricmp( szValue, "func_detail_nosolid"))
+		else if (!stricmp( szValue, "func_detail_nonsolid"))
 		{
 			pLoadEntity->nBaseContents = CONTENTS_DETAIL | CONTENTS_OPAQUE;
 		}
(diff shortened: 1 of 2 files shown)


chanticleerhegemony:
`func_detail_illusionary` would be a better name to ensure consistency with other quake derivatives

SirYodaJedi:
> func_detail_illusionary would be a better name to ensure consistency with other quake derivatives

As well as consistent with `func_illusionary`.

Edit: VBSP++ also calls it func_detail_illusionary.

dashishmeister:
don't see why this functionality can't be added to normal func_detail as a spawnflag or a keyvalue

SirYodaJedi:
> don't see why this functionality can't be added to normal func_detail as a spawnflag or a keyvalue

It could, but
1. This way is clearer at a glance, because you can tell just by reading the classname in the 2D view of hammer that it's supposed to be nonsolid. (same reason func_illusionary is still used instead of a func_brush with collision disabled)
2. It prevents them from being treated as solid by other game's VBSPs when porting (it's much more obvious that tinkering needs to be done when the geometry is flat out missing)
3. `func_detail_illusionary` is already an established term, used by VBSP++ and Quake's ericw-tools.

PR #1324, opened 2025-06-01
 [TF2] Decouple TF_COND_HALLOWEEN_THRILLER from blocked input, instead adding FREEZE_INPUT when it is added to provide the same effects

FlaminSarge:
# Description

- HALLOWEEN_THRILLER now only plays the music, rather than stopping inputs, plus it forces the Halloween taunt when attempting to taunt. It also maintains thirdperson camera if the player enters thirdperson camera, which is the same as stock behavior.
- To maintain backwards compatibility with custom maps that may use it, HALLOWEEN_THRILLER also adds FREEZE_INPUT when added.
	- This results in the same behavior as stock for all maps including custom maps.
- This now allows HALLOWEEN_THRILLER to open console, which helps avoid a softlock when using `addcond 54` and offers greater flexibility for using the Thriller condition (maps may add it then remove FREEZE_INPUT immediately to use it by itself).
- This prevents using action slot during FREEZE_INPUT.
- This also prevents using econ taunts while in HALLOWEEN_THRILLER, forcing the Thriller taunt instead.
- In order for the functionality of the Dance events in various maps to remain consistent, players are allowed to look around during FREEZE_INPUT if they also have HALLOWEEN_THRILLER. This is so that SlamPosAndAngles/other Teleports can set the player's facing angle properly.

This PR also includes #1331 as it removes some behavior from that PR due to the change in THRILLER condition behavior.
To view the difference between that PR and this one, please see https://github.com/FlaminSarge/source-sdk-2013/compare/bot-freeze...FlaminSarge:source-sdk-2013:thriller-console

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -6496,11 +6496,16 @@ bool C_TFPlayer::CreateMove( float flInputSampleTime, CUserCmd *pCmd )
 	static float flTauntTurnSpeed = 0.f;
 	
 	bool bNoTaunt = true;
-	bool bInTaunt = m_Shared.InCond( TF_COND_TAUNTING ) || m_Shared.InCond( TF_COND_HALLOWEEN_THRILLER );
+	bool bInTaunt = m_Shared.InCond( TF_COND_TAUNTING );
 
 	if ( m_Shared.InCond( TF_COND_FREEZE_INPUT ) )
 	{
-		pCmd->viewangles = angMoveAngle; // use the last save angles
+		if ( !m_Shared.InCond( TF_COND_HALLOWEEN_THRILLER ) )
+		{
+			// if NOT halloween taunt, force their last-saved view angles
+			// (thriller allows it so teleports can set the correct view angles)
+			pCmd->viewangles = angMoveAngle;
+		}
 		pCmd->forwardmove = 0.0f;
 		pCmd->sidemove = 0.0f;
 		pCmd->upmove = 0.0f;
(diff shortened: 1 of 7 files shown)


FlaminSarge:
If anyone can try to find any more exploits for this, please let me know. Thriller blocking KeyInput but not the underlying commands is a bit strange, but fully locking out console to paper over it doesn't seem correct.

Whurrhurr:
i know TFBots can somehow break out of the control-locking TF_COND_HALLOWEEN_THRILLER is supposed to do, and it's been happening for almost a decade now, but i'm not quite sure what causes that to happen. (i've also seen human players break out of the control-locking TF_COND_HALLOWEEN_THRILLER puts on you. maybe it's the same way TFBots are using to break out of it?)

if you want to quickly test it, load up ghost fort, add some tfbots and force-spin the wheel with ent_fire wheel_of_fortress spin until it lands on the thriller outcome. bots will freely attack eachother, but be unable to move, and get forced to use the thriller taunt when the condition forces them to use it.

FlaminSarge:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1324#issuecomment-2932959089

Chances are that the actual attack methods aren't blocked during it, solely the _keybind_ for attacking. Which is why blocking only the KeyInputs is suboptimal. I think I'll have to add cases for +attack/etc here. I'm going to see if there's a better way to handle this.

FlaminSarge:
> i know TFBots can somehow break out of the control-locking TF_COND_HALLOWEEN_THRILLER is supposed to do

I have opened a PR for this at https://github.com/ValveSoftware/source-sdk-2013/pull/1331

PR #1325, opened 2025-06-01
Fix mount path in sdk_container

xVemu:
# Description
Current script throws an error on Ubuntu 22.04 LTS: `Error: "source: invalid mount option`. This PR fixes that.


xVemu:
Duplicate #645

PR #1327, opened 2025-06-03
[TF2] fix: Client not updating workshop maps

Pandaptable:
# Description
Currently the client waits for Steam to notice a pending update for workshop maps, during this entire time GetItemState will say the map is up-to-date. This even happens when the workshop map is manually fetched using SendQueryUGCRequest.

This issue exists also on dedicated servers and the old code already accounted for this issue, where the game always called DownloadItem, regardless of GetItemState saying the workshop map needed an update or not.

The result is that the dedicated server will update the workshop map but the client will not, resulting in a map mismatch and the client getting kicked.

This PR makes the dedicated server behavior also apply to the client, DownloadItem does an update check for us so no need to worry about double-downloading workshop maps, if there is no update nothing will be downloaded.

### DIFF
--- a/src/game/server/tf/workshop/maps_workshop.cpp
+++ b/src/game/server/tf/workshop/maps_workshop.cpp
@@ -217,42 +217,19 @@ void CTFWorkshopMap::Steam_OnQueryUGCDetails( SteamUGCQueryCompleted_t *pResult,
 	}
 
 	uint32 state = steamUGC->GetItemState( m_nFileID );
-	if (( state & k_EItemStateNeedsUpdate ) ||
-	    !( state & ( k_EItemStateDownloading | k_EItemStateDownloadPending | k_EItemStateInstalled ) ) )
+	if ( ! ( state & k_EItemStateDownloading ) )
 	{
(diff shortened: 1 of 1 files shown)


mastercoms:
This change seems to ignore the apparent bug for `DownloadItem` that is outlined in the comment: `NOTE There is another bug where calling DownloadItem() on the *non-gameserver* api on a fully up to date item`. If this bug was not fixed in Steam, this PR should not be merged as it's not a viable fix without that.

Pandaptable:
> If this bug was not fixed in Steam, this PR should not be merged as it's not a viable fix without that.
To my knowledge it has been fixed in steam 👍

Pandaptable:
I can confirm despite valve saying in a steam client beta update "Re-worked system that managed what subscribed items needed to be updated." on [July 10th](https://steamcommunity.com/groups/SteamClientBeta/announcements/detail/667244800007733950), the issue still persists.
I tested same as always:
1. Upload map to workshop
2. Make a server run the map
3. Update the map on the server and reload map
4. Server updates but the client doesn't

PR #1328, opened 2025-06-03
[ANY] feat: Allow mp_disable_respawn_times to work with specific team numbers

Pandaptable:
# Description
Makes `mp_disable_respawn_times` work with the team number, for TF2 for example `mp_disable_respawn_times 2` would disable respawn times for RED team only.


henke37:
This should probably be a bitfield, not a single team number.

PR #1330, opened 2025-06-03
[TF2] Add thousands separators to various values

horiuchii:
# Description
This PR adds thousands separators to various values across the game that tend to become very large and difficult to read at first glance without separators, this previously only existed on the player list stats on the MvM scoreboard.

I moved and renamed `ConvertScoreboardValueToString` inside of `tf_hud_mann_vs_machine_scoreboard.cpp` to `V_AddThousandSeparators` inside of `strtools.cpp` to make it a more accessible function, this was to avoid using `V_pretifynum` as it's around 4x slower and results in the same desired output.

The thousands separator applies to the following values:
- Strange item counts
- All stats displayed on scoreboards
- PVP Panel stats (Casual XP, Comp MMR and unused MYM matchmaking stats)
- Competitive match history stats
- MvM Currency (Counts near the player health and the upgrade menu and upgrade cost)

# Examples:

### DIFF
--- a/src/game/client/tf/player_vs_environment/c_tf_upgrades.cpp
+++ b/src/game/client/tf/player_vs_environment/c_tf_upgrades.cpp
@@ -1175,10 +1175,7 @@ void CHudUpgradePanel::UpdateUpgradeButtons( void )
 					pUpgradeBuyPanel->m_pShortDescriptionLabel->SetText( pDescription );
 				}
 
-				locchar_t wzCost[64];
-				loc_sprintf_safe( wzCost, LOCCHAR( "%d" ), pUpgradeBuyPanel->m_nPrice );
-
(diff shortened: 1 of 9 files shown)


gidi30:
Sounds great but it might be good to add some way to localize this since certain languages use `.` as the thousands seperator and `,` as the decimal seperator

gaelcoral:
> Sounds great but it might be good to add some way to localize this since certain languages use `.` as the thousands seperator and `,` as the decimal seperator

In Spanish both for Latin America and Spain, four-digit numbers do not use a period or comma as a separator. For five digits or more, a space is used as the thousands separator (ideally a non-breaking space).

So, if this pull request were to be implemented, the values should be implemented as localized strings with format variables, so localizers can properly apply the correct number formatting for their language (like Deadlock).
image

PR #1331, opened 2025-06-05
[TF2] Add CTFBotFreezeInput behavior to make bots respect TF_COND_FREEZE_INPUT

FlaminSarge:
# Description
As described in a comment on https://github.com/ValveSoftware/source-sdk-2013/pull/1324#issuecomment-2932959089 bots do not respect the 2 conditions that prevent player input, TF_COND_HALLOWEEN_THRILLER and TF_COND_FREEZE_INPUT. This PR causes both conditions to make bots stop all other behavior while they have the conditions.

As a note, with #1324 I plan to decouple THRILLER from blocked inputs and solely tie it to dancing and music, leaving FREEZE_INPUT as the only frozen input condition, which would also update this to remove the THRILLER case here.

This has been tested on Lakeside Event for THRILLER and with manually adding FREEZE_INPUT. If necessary I can test on Doomsday Event and Helltower as well, but I feel that's not necessary.


FlaminSarge:
~~I'm investigating whether this can be replaced with a `Continue()` in the initial FREEZE_INPUT check instead of a whole new behavior. However, a new behavior seems like the 'correct' way to handle this, rather than calling Continue() early to skip CTFBotMainAction's handling, particularly when other triggers in CTFBotMainAction could potentially interrupt the behavior (e.g. OnInjured/OnContact/etc.). I'm not sure whether OnKilled needs special handling.~~
The new behavior class is necessary; the bots attempt to jump/etc. as if they're stuck if the new behavior isn't used.

PR #1337, opened 2025-06-07
[TF2] Prevent bots from targeting ghosts

FlaminSarge:
# Description
This includes selecting heal targets, attacking, and includes NPCs and tf_bots


FlaminSarge:
Currently testing for any side effects of the m_lifeState change before I move this out of draft.

PR #1344, opened 2025-06-10
[TF2] Properly remove/scale Spy cigarette/Medic glasses with decapitation/head smash/big head transformations

FlaminSarge:
# Description
This addresses a handful of missing handling cases for player model transformations:
* Decapitated and head-smashed ragdolls keep the Spy's cigarette and the Medic's glasses floating above the severed neck/shortened neck model; this patch hides them under these conditions
* The Big Head transformations do not scale the cigarette/glasses, resulting in them disappearing underneath the enlarged model; this patch scales them so that they appear alongside the enlarged (or shrunken) heads in the correct positions
* Removes iClass param from BuildTorsoScaleTransformations as it is entirely unused

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -8393,10 +8393,12 @@ void BuildBigHeadTransformations( CBaseAnimating *pObject, CStudioHdr *hdr, Vect
 	// Scale the head.
 	MatrixScaleBy ( flScale, transform );
 
-	const int cMaxNumHelms = 2;
+	const int cMaxNumHelms = 4;
 	int iHelmIndex[cMaxNumHelms];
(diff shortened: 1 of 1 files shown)


src/game/client/tf/c_tf_player.cpp
```diff
@@ -8393,10 +8393,12 @@ void BuildBigHeadTransformations( CBaseAnimating *pObject, CStudioHdr *hdr, Vect
 	// Scale the head.
 	MatrixScaleBy ( flScale, transform );
 
-	const int cMaxNumHelms = 2;
+	const int cMaxNumHelms = 4;
 	int iHelmIndex[cMaxNumHelms];
```
henke37, line 8397: Perhaps it is time to rename these variables.

Issue #1343, opened 2025-06-10
[VRAD] Add Source SDK 2007 / Pre-SteamPipe Static Prop Lighting & AO


EternallyEmber:
Ambient static prop lighting was darker when the TF2's Source SDK was updated to the 2013 SteamPipe version. it resulted in some maps having their lighting broken, 2fort, being one example. (below is a photo of pre-steampipe 2fort.)

![Image](https://github.com/user-attachments/assets/7a47c7b0-12dc-4210-aaff-8aa65a1f9ce3)

Issue #6250 in Source-1-Games: https://github.com/ValveSoftware/Source-1-Games/issues/6250
talked about this issue.

i talked to SlartyBarty, the creator of the Slammin' Source Map tools about the issue.

![Image](https://github.com/user-attachments/assets/2f2b9249-a6fe-41e6-88a0-70e0ed0b94e7)

his tools do not only fix this issue, but it also adds AO to VRAD 2013.

he's holding back from submitting a pull request, because submitting a pull request with valve's own old code to fix an issue, would be a bit ironic, and maybe legally questionable??? ( i dont know how exacly that works tho )

please dont try to contact him to ask to make a pull request, i just hope valve would maybe give a green flag,
or that someone else might go ahead and fix it after seeing this pull request
if valve gives a green flag im sure to let him know, or better yet,
im just sending him a link on this github issue.

if you wanna know what this fix + added AO looks like, i have a youtube video on a workshop submission i made:
https://steamcommunity.com/sharedfiles/filedetails/?id=3479828308 
i also enhanced some other things, so please only take notice of the fixed lighting, and not of the water reflections, and new animations on the map.

ficool2:
https://github.com/ValveSoftware/source-sdk-2013/pull/1345

hourai-branch:
https://github.com/Biohazard90/g-string_2013/commit/d3b4d1402693d75fe0fcdc67ae49bcfaa22a895e#diff-f476b67a0982d8e0481ebaf002a8b1d95dd8fc3196bcc6e61b0c8f5aabc1cc47

^this old SDK fork adds AO. it would be nice if it could get ported and pulled.
im currently doing my best trying with pull requests and all but not knowing C++ makes this hard.

i can imagine someone else doing it in a couple of hours, or minutes.

maybe i shouldn't judge coding time, im a hobbyist sysadmin, and not a coder.

if i was a coder i'd probably have more success getting this done.

[...]

PR #1345, opened 2025-06-10
Fix VRAD prop indirect lighting regression

ficool2:
VRAD's prop lighting has been noted to regress since Orangebox particularly in TF2. Specifically, prop lighting appears to be darker and more harsh overall. After some digging, this is caused by a faulty calculation for how indirect (bounced) lighting is sampled on props. This has been broken since around 2014 (when `rd_asteroid` was released.)

Orangebox VRAD did not take into account any falloff or dot from a lighting sample. However because of this, it lead to an unnaturally bright result, although some users opinionated this is an improvement over current VRAD.

Current (TF2) VRAD accounts falloff from the sample, using an inverse square law, presumably to fix the bright results of the Orangebox's method. Unfortunately, this falloff is quite harsh and therefore creates an unnaturally dark result.

This PR implements a middleground of those both in VRAD. It only takes into account the dot of a sample, which leads to a decent spot in-between where it's not too dark or not bright. In-fact, this is the same change that CS:GO VRAD implemented to have more natural prop lighting. A bonus is this also more closely matches the indirect lighting calculation for brushes.

However, for the sake of compatibility with maps that might depend on the previous algorithms' results, a new VRAD option named `-StaticPropIndirectMode` is added to allow users to switch to the old TF2 or Orangebox algorithm.

Comparison:
2fort

### DIFF
--- a/src/utils/vrad/vrad.cpp
+++ b/src/utils/vrad/vrad.cpp
@@ -119,7 +119,7 @@ bool		g_bStaticPropLighting = false;
 bool        g_bStaticPropPolys = false;
 bool        g_bTextureShadows = false;
 bool        g_bDisablePropSelfShadowing = false;
-
+int			g_nIndirectPropLightingMode = 0;
 
 CUtlVector<byte> g_FacesVisibleToLights;
 
@@ -2408,6 +2408,18 @@ int ParseCommandLine( int argc, char **argv, bool *onlydetail )
 		{
 			g_bDisablePropSelfShadowing = true;
 		}
+		else if ( !Q_stricmp( argv[i], "-StaticPropIndirectMode" ) )
+		{
+			if ( ++i < argc )
+			{
+				g_nIndirectPropLightingMode = atoi( argv[i] );
+			}
+			else
+			{
+				Warning( "Error: expected a value after '-StaticPropIndirectMode'\n" );
+				return -1;
+			}
+		}
 		else if ( !Q_stricmp( argv[i], "-textureshadows" ) )
 		{
 			g_bTextureShadows = true;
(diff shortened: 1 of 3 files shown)


hourai-branch:
oh, hi! Ficool!

This solution Looks amazing! ❤️
i suppose adding AO too as #1343 mentioned is a optional enhancement better left to be done separately someday.

now i just really hope this gets in!!!!

SirYodaJedi:
This is a huge improvement; I had assumed static props weren't taking indirect lighting into account at all, as I was getting some parts of props as pitch black when recompiling DoDS maps.
I'm definitely going to be adding this to my personal VRAD fork, but I do hope that if this PR is approved, the other first party SDK2013 games besides TF2 get the update as well, so everyone can benefit.

For clarification, does the lighting issue also affect the singleplayer SDK? The comments mention TF2 specifically, but it sounds like the change was a 2013 thing (so it'd also affect HL2).

Bitl:
> This is a huge improvement; I had assumed static props weren't taking indirect lighting into account at all, as I was getting some parts of props as pitch black when recompiling DoDS maps. I'm definitely going to be adding this to my personal VRAD fork, but I do hope that if this PR is approved, the other first party SDK2013 games besides TF2 get the update as well, so everyone can benefit.
> 
> For clarification, does the lighting issue also affect the singleplayer SDK? The comments mention TF2 specifically, but it sounds like the change was a 2013 thing (so it'd also affect HL2).

this also effects the SP sdk as well i think.

ficool2:
> For clarification, does the lighting issue also affect the singleplayer SDK? The comments mention TF2 specifically, but it sounds like the change was a 2013 thing (so it'd also affect HL2).

Games not derived from SDK 2013 MP have the orangebox algorithm (and so suffer from being too bright)

hourai-branch:
I think its kind of sad to see valve has done nothing with this yet...

PR #1347, opened 2025-06-10
all games: Fix segfaults in CBaseEntity when VScript is disabled

copperpixel:
# Description
When VScript is disabled using the `-noscripting` commandline param or when an invalid scripting language is specified, the VScript VM doesn't get initialized which shouldn't be fatal. However the server will attempt to dereference a NULL `g_pScriptVM` pointer due to missing checks in `CBaseEntity::RunVScripts`, `CBaseEntity::RunPrecacheScripts` and `CBaseEntity::RunOnPostSpawnScripts` leading to a server crash on map load if there are any entities with scripts assigned to them.

The fix for this is to make the functions return early if `ValidateScriptScope` fails.

Closes https://github.com/ValveSoftware/Source-1-Games/issues/5122

### DIFF
--- a/src/game/server/baseentity.cpp
+++ b/src/game/server/baseentity.cpp
@@ -8311,12 +8311,10 @@ bool CBaseEntity::ValidateScriptScope()
 //-----------------------------------------------------------------------------
 void CBaseEntity::RunVScripts()
 {
-	if( m_iszVScripts == NULL_STRING )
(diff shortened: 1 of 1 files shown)


MrSquid47:
We're hitting these crashes on our TF2 servers where we disable VScript for most maps. I can work around it by hooking these functions but merging this proper fix would be great.

PR #1350, opened 2025-06-12
Always lowercase filenames in CUtlFilenameSymbolTable to fix loading packed files on Linux (fixes #865)

### LINKED ISSUE #865: [Bug] Linux CUtilFilenameSymbolTable probably conflicts with packed-file casing
Per https://github.com/ValveSoftware/Source-1-Games/issues/6868 loading packed files from BSPs (& VPKs!) has some casing issue that popped up recently for Linux users.

The problem might stem from `CZipPackFile::GetFileInfo()` (and `CPackFile`?) which is used when trying to open packed files.

Previously (at least in CS:S `previous_build`) it would do something like this:
```C++
CPackFileEntry entry;
entry.hash = HasStringCaselessConventional(filename);
int i = this->files.Find(entry);
if (i != -1)
{
	// return file info here...
	return 1;
}
return 0;
```

Now it looks like it does something like this:
```C++
CPackFileEntry entry;
entry.hash = HasStringCaselessConventional(filename);
int i = this->files.Find(entry);
if (i != -1)
{
	FileHandle_t f = this->filesystem_maybe->FindFileName_maybe(filename);
	if (this->files[i].handle == f)
	{
		// return file info here...
		return 1;
	}
}
return 0;
```

I think this is making its way to `CUtilFilenameSymbolTable::FindFileName()`
https://github.com/ValveSoftware/source-sdk-2013/blob/aea94b32cbefeba5d16ef6fc70eff9508cf11673/src/tier1/utlsymbol.cpp#L362-L381
where it'll `Q_strlower(filename)` *only on Windows*.

So even if the packed file *did* find a match case-insensitively, the `FindFileName()` might be causing it to fail.

(Same with `CUtilFilenameSymbolTable::FindOrAddFileName()`)
https://github.com/ValveSoftware/source-sdk-2013/blob/aea94b32cbefeba5d16ef6fc70eff9508cf11673/src/tier1/utlsymbol.cpp#L312-L332

rtldg:
# Description
Tracked down in https://github.com/ValveSoftware/source-sdk-2013/issues/865, loading packed files (e.g. from bsps & vpks) is often broken on community maps for Linux users (& servers!!!) leading to missing textures and broken model collisions.

Also reported at https://github.com/ValveSoftware/Source-1-Games/issues/6868

I'm making this PR to bring more visibility to the cause and to hopefully be merged.
(The cause being that `CUtlFilenameSymbolTable` only lowercases filenames on Windows, which is called from `CZipPackFile::GetFileInfo()` to use a cached file-handle.)

## Alternate approach
@mutezero https://github.com/ValveSoftware/source-sdk-2013/issues/865#issuecomment-3149873648

### DIFF
--- a/src/tier1/utlsymbol.cpp
+++ b/src/tier1/utlsymbol.cpp
@@ -327,9 +327,7 @@ FileNameHandle_t CUtlFilenameSymbolTable::FindOrAddFileName( const char *pFileNa
 	char fn[ MAX_PATH ];
 	Q_strncpy( fn, pFileName, sizeof( fn ) );
 	Q_RemoveDotSlashes( fn );
-#ifdef _WIN32
 	Q_strlower( fn );
-#endif
 
 	// Split the filename into constituent parts
 	char basepath[ MAX_PATH ];
@@ -370,9 +368,7 @@ FileNameHandle_t CUtlFilenameSymbolTable::FindFileName( const char *pFileName )
 	char fn[ MAX_PATH ];
 	Q_strncpy( fn, pFileName, sizeof( fn ) );
 	Q_RemoveDotSlashes( fn );
-#ifdef _WIN32
 	Q_strlower( fn );
-#endif
 
 	// Split the filename into constituent parts
 	char basepath[ MAX_PATH ];


Tolfx:
Have you compiled and tested it works on all platforms like normal? As well fixing the issue https://github.com/ValveSoftware/Source-1-Games/issues/6868?

rtldg:
> Have you compiled and tested it works on all platforms like normal? As well fixing the issue [ValveSoftware/Source-1-Games#6868](https://github.com/ValveSoftware/Source-1-Games/issues/6868)?

It's used in engine code (`CZipPackFile`) which we can't rebuild, so it can't be tested by compiling, but it has been confirmed to fix https://github.com/ValveSoftware/Source-1-Games/issues/6868 after hooking the relevant functions:
- client-side https://github.com/ValveSoftware/Source-1-Games/issues/6868#issuecomment-2707662934
- server-side https://github.com/rtldg/linux_file_lowercaser

rtldg:
~~A user of the server-side plugin reported that it could conflict with packed models that use capital letters https://github.com/rtldg/linux_file_lowercaser/issues/1#~~

~~But there's not much I can investigate for it right now.~~

EDIT: Seems to be fine after all so never mind 😇

rtldg:
@mutezero has posted about another approach to handle case-insensitive file name/path comparisons that could be considered too https://github.com/ValveSoftware/source-sdk-2013/issues/865#issuecomment-3149873648

PR #1355, opened 2025-06-14
[TF2] Allow self inspection

horiuchii:
# Description
This PR adds the Client `tf_inspect_allow_self` CVar, which defaults to 0, and the `inspect_self` Command.

When `tf_inspect_allow_self` is set to 1 and you attempt to inspect without looking at another player, you will instead inspect your equipped items as if you were inspecting another player, this is useful for custom gamemodes like VScript / SM VSH that often modify the attributes of your weapons, of which provide no simple or intuitive way to view them.

Executing `inspect_self` will always inspect yourself, good for those who wish to bind a key to inspect yourself while keeping the previous functionality of the inspect key.

This PR also allows you to close the inspect panel with `+attack2` and `+attack3`, to more easily close the inspect panel when `tf_inspect_allow_self` is set to 1, just like it already does with `+attack`.

### DIFF
--- a/src/game/client/tf/tf_hud_inspectpanel.cpp
+++ b/src/game/client/tf/tf_hud_inspectpanel.cpp
@@ -28,6 +28,7 @@
 // memdbgon must be the last include file in a .cpp file!!!
 #include "tier0/memdbgon.h"
 
+static ConVar tf_inspect_allow_self("tf_inspect_allow_self", "0", FCVAR_ARCHIVE, "When attempting to inspect without looking at another player, Inspect yourself.");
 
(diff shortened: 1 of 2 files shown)


HyperionCoding:
How does this affect MvM? Inspecting without target opens your upgrade menu.

horiuchii:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1324#issuecomment-2932959089

`tf_inspect_allow_self` has no effect over inspecting in MvM and the `inspect_self` command will inspect your own upgrades as expected.

PR #1358, opened 2025-06-15
[TF2] Add ConVar for Tournament HUD to ensure client visiblity

irql-notlessorequal:
This adds a new ConVar (`cl_hud_tournament_always_centered`) that creates an additional pass to `CHudTournament::RecalculatePlayerPanels` that when enabled and past 9 defenders, will swap out our client with a player in the middle of the panel list, ensuring that we are once again visible.

Useful for community MvM gamemodes utilizing above normal defender counts.

Before:
Screenshot_2025-06-16_01-27-35

After and enabled:
Screenshot_2025-06-16_01-27-08

### DIFF
--- a/src/game/client/tf/tf_hud_tournament.cpp
+++ b/src/game/client/tf/tf_hud_tournament.cpp
@@ -48,7 +48,10 @@ void AddSubKeyNamed( KeyValues *pKeys, const char *pszName );
 
 using namespace vgui;
 
+#define TOURNAMENT_PANEL_OVERFLOW_THRESHOLD 9
 #define TOURNAMENT_PANEL_UPDATE_INTERVAL 0.25f
+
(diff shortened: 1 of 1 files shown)


FlaminSarge:
It may be worth allowing this UI to overflow onto a 2nd line or use wider bounds instead, even if that would need hud file updates as well.

irql-notlessorequal:
> It may be worth allowing this UI to overflow onto a 2nd line or use wider bounds instead, even if that would need hud file updates as well.

In my opinion having it overflow into a 2nd line could work however it'd cause visual clutter, and would get worse with more defenders a community gamemode would have.

On the otherhand, using wider bounds doesn't help much for 4:3 and 5:4 users and isn't enough to for example have all 24 players be visible (maybe at 4K or higher resolutions it would?) considering the HUD does scale, see below screenshot as a proof of concept.

Screenshot_2025-06-23_12-46-08

Maybe having a new variant of the tournament HUD that would show a percentage of the ready state of players would be a much better idea but that would be additional work.

PR #1363, opened 2025-06-18
[ALL] (Vscript) Make/Destroy Physics

ALIEN31ITA:
MakePhysical( nSolidType, nSolidFlags, asleep )
DestroyPhysics()

# Description
These vscript function can be called on any entity, to make, or destroy physics, letting the user experiment more for their creations, instead on relaying on prop_phyisics or other already physical entities.

this doesn't come with safety checks, so for example, you can run this on players if you ever wanted to, so use at own risk.


kiwitf2:
Tested, LGTM

PR #1365, opened 2025-06-21
tf2: Fix MvM bots not being able to use their items

Mikusch:
In some cases, robots in MvM mode can not use their items (Buff Banner, Bonk, etc.), due to the timer in `CTFBotUseItem` ending before the weapon is fully deployed. This leads to a loop of robots constantly switching to their items without activating it.

We should wait until the bot is actually able to "attack" aka use the item.

Resubmission of #875 due to a deleted fork

### DIFF
--- a/src/game/server/tf/bot/behavior/tf_bot_use_item.cpp
+++ b/src/game/server/tf/bot/behavior/tf_bot_use_item.cpp
@@ -47,9 +47,12 @@ ActionResult< CTFBot >	CTFBotUseItem::Update( CTFBot *me, float interval )
 	{
 		if ( m_cooldownTimer.IsElapsed() )
 		{
-			// use it
-			me->PressFireButton();
-			m_cooldownTimer.Invalidate();
+			if ( me->GetNextAttack() <= gpGlobals->curtime )
+			{
+				// use it
+				me->PressFireButton();
+				m_cooldownTimer.Invalidate();
+			}
 		}
 	}
 	else // used


KaelaSavia:
This will affect balance of Mann Up/Boot Camp heavily, I'd recommend making it convar that is turn-onable for community servers instead

Mikusch:
This is fixing broken behavior introduced in Gun Mettle, due to the increased switch speed. I do not see how this affects the balance.

src/game/server/tf/bot/behavior/tf_bot_use_item.cpp
```diff
@@ -47,9 +47,12 @@ ActionResult< CTFBot >	CTFBotUseItem::Update( CTFBot *me, float interval )
 	{
 		if ( m_cooldownTimer.IsElapsed() )
```
zxz41, line 48: I know this is a bit of format nitpicking, but the if statements can be combined into one like so.
```suggestion
		if ( m_cooldownTimer.IsElapsed() && me->GetNextAttack() <= gpGlobals->curtime )
```

KimmyTF2:
> This will affect balance of Mann Up/Boot Camp heavily, I'd recommend making it convar that is turn-onable for community servers instead

What? This is a bug messing with the gameplay. How is that related to balance in any way?

henke37:
> > This will affect balance of Mann Up/Boot Camp heavily, I'd recommend making it convar that is turn-onable for community servers instead
> 
> What? This is a bug messing with the gameplay. How is that related to balance in any way?

Bugs affecting gameplay can change the balance. And this PR is worth any changes.

PR #1371, opened 2025-06-23
[HL2MP] Finish bot health charger support

Blixibon:
HL2MP bots have unfinished support for using `item_healthcharger` and `func_healthcharger` entities to heal themselves. This pull request allows them to be supported to the same extent as regular health kits.

Specific changes made:

- Chargers with no juice are not considered when searching for health kits.
- Bots navigate to a point in front of the chargers, rather than directly into the chargers.
	- Positions can be displayed using `hl2mp_bot_debug_health_scavenging`, which was previously unused.
- Chargers are now identified by finding `item_health*`/`func_health*` and then searching for "charger" in the classname directly, rather than through leading wildcards with `*charger*`/`*_health*`, which Source does not support by default.
	- Previously, this was causing the health entity filter to evaluate *all* entities, and for all health kits to be considered chargers.
- Range increased from `50` to the actual player use radius, which is `80` by default.
- Bots drop any prop they're holding with the gravity gun before interacting with a charger.
- Bots now look at the charger while using it.
- Bots now press the +USE button directly rather than invoking the `Use()` function on the chargers, as the latter had issues.
- Fix critical health not being evaluated correctly when deciding whether to abandon a charger.
- Bots on "Expert" difficulty will now move left and right while using a charger to avoid standing still. This behavior can be toggled with `hl2mp_bot_health_charger_weave`.

Here's a video of how health scavenging behavior worked prior to this pull request:

https://github.com/user-attachments/assets/a6b1c5f7-1155-4cd7-8f54-e08359c45396

Here's a video showing how it works with the changes included:

https://github.com/user-attachments/assets/aa6121f5-396e-4246-96fa-a3f0cdcb8b4d

Here's a video showing a bot on "Expert" difficulty using the aforementioned "weaving" behavior:

https://github.com/user-attachments/assets/96617a09-b412-400a-86d1-d195ece47bb1


Blixibon:
I made additional changes for bots considering threats and reflected that in the change log.

src/game/server/hl2mp/bot/behavior/hl2mp_bot_get_health.cpp
```diff
@@ -146,10 +289,10 @@ ActionResult< CHL2MPBot >	CHL2MPBotGetHealth::OnStart( CHL2MPBot *me, Action< CH
 	}
 
 	m_healthKit = s_possibleHealth;
-	m_isGoalCharger = m_healthKit->ClassMatches( "*charger*" );
+	m_isGoalCharger = s_possibleIsCharger;
```
ZzZombo, line 292: Are two variables even needed? Why can you not use just the existing one?

src/game/server/hl2mp/bot/behavior/hl2mp_bot_get_health.cpp
```diff
@@ -184,9 +327,16 @@ ActionResult< CHL2MPBot >	CHL2MPBotGetHealth::Update( CHL2MPBot *me, float inter
...
 	{
 		// we need to get near and wait, not try to run over
-		const float nearRange = 50.0f;
-		if ( ( me->GetAbsOrigin() - m_healthKit->GetAbsOrigin() ).IsLengthLessThan( nearRange ) )
+		const float nearRange = PLAYER_USE_RADIUS;
+		Vector vecDir = (me->EyePosition() - m_healthKit->WorldSpaceCenter());
+		if ( vecDir.IsLengthLessThan( nearRange ) )
 		{
+			// if we're carrying a prop, drop it
+			if ( me->Physcannon_GetHeldProp() )
```
ZzZombo, line 335: I don't see an equivalent code for holding a prop without the gravity gun.

PR #1379, opened 2025-06-24
Add nullptr check for pTFAttacker

Kenzzer:
# Description

This fixes a very rare crash in 

https://github.com/ValveSoftware/source-sdk-2013/blob/39f6dde8fbc238727c020d13b05ecadd31bda4c0/src/game/shared/tf/tf_gamerules.cpp#L10659-L10661

As you can see `pOwner` is never null checked for. And in very rare instances 

https://github.com/ValveSoftware/source-sdk-2013/blob/39f6dde8fbc238727c020d13b05ecadd31bda4c0/src/game/server/tf/tf_player.cpp#L9840-L9855

That function is called when `pTFAttacker` is nullptr.

Since `pTFAttacker` is already null-checked in lot of places inside `int CTFPlayer::OnTakeDamage( const CTakeDamageInfo &inputInfo )` let's just add one extra check there (see PR changes).

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -9837,21 +9837,24 @@ int CTFPlayer::OnTakeDamage( const CTakeDamageInfo &inputInfo )
 	}
 
 
-	CTFWeaponBase *pTFWeapon = GetKilleaterWeaponFromDamageInfo( &info );
-	if ( !pTFWeapon )
+	if ( pTFAttacker )
 	{
-		// Check Wearable instead like demoshields or manntreads
-		CTFWearable *pWearable = dynamic_cast< CTFWearable* >( info.GetWeapon() );
-		if ( pWearable )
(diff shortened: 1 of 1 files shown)


FlaminSarge:
Would it make more sense to add the check and early return inside the method that the crash actually occurs in, so that any future callers get the safety check for free? (Potentially in addition to the check added here, since skipping the logic makes sense if invalid attacker).

Kenzzer:
The thought has crossed my mind before making this PR however I ultimately decided against it, on the grounds that `pTFAttacker` is null checked in a ton of places inside `CTFPlayer::OnTakeDamage` and it made more sense to continue in that spirit. Furthermore it dispels the wrong assumption that `pTFAttacker` isn't nullptr if `pTFWeapon` is valid. There could be more functions than `EconEntity_OnOwnerKillEaterEvent_Batched` called under those if statements in the future.

If Valve prefers to move the nullptr check I'm sure they'll comment on the PR or do it themselves.

PR #1380, opened 2025-06-25
all games: Add support for animated avatars

### LINKED ISSUE #5566: [TF2] [Feature Request] Make animated avatars be animated
Like in CS2
(obviously, there should be an option to disable animated avatars if this gets implemented)

copperpixel:
# Description
This PR implements support for displaying Steam animated avatars on scoreboard, friendslist, etc.
Players can disable animated avatars using `cl_animated_avatars` ConVar.

The `CGifMaterial` class can be used to read GIF data into a material to be displayed in-game.

Demonstration video:

[animated_avatars2_c.webm](https://github.com/user-attachments/assets/3a4efa9b-bd93-4f00-aabd-ee4587a8ca20)

Closes https://github.com/ValveSoftware/Source-1-Games/issues/5566

GIFLIB v6.1.3 dependency is added for decoding GIFs pulled from Steam. It has been sourced from [here](https://giflib.sourceforge.net/) and is released under a permissive MIT/X Consortium-like license:

https://github.com/ValveSoftware/source-sdk-2013/blob/3068fd6518f1e6a83937aad7d3fde98735d16c8e/game/thirdpartylegalnotices.txt#L914-L936

There were only few modifications made to the original library:
- Added CMake support for more streamlined building
- Replaced OpenBSD array realloc with custom implementation

For convenience, I have included precompiled amd64 binaries of GIFLIB for Windows and Linux. You can verify their integrity using respective SHA-256 checksums:
<details>
<summary>`giflib.lib` (Windows)</summary>

```sha256
ab1b7789e13d9533945e9615867f96b39bac03b223e999eecb09ff5429cf1ab5  giflib.lib
```
</details>
<details>
<summary>`libgiflib.a` (Linux)</summary>

```sha256
04c855d3f0228f16d753eb935cf9b48a40fda3bb952c76a70613f37d0faf6662  libgiflib.a
```
</details>

The modified GIFLIB source is available in `src/thirdparty/giflib-6.1.3`. You can find build instructions in `BUILD.MD` file where the source is located.


gaelcoral:
Please Valve add this 🙏

Anjo2807:
cool stuff

Bitl:
Would probably be better to have pre-compiled binaries for windows/linux to make it easier to implement in mod projects

copperpixel:
> Would probably be better to have pre-compiled binaries for windows/linux to make it easier to implement in mod projects

When this PR will be ready to merge I'll include them with a way of verifying their integrity

Bitl:
I don't think this PR is stable yet. (although i might be wrong as i'll describe later)

I implemented it into my mod and it works perfectly initially. However, if I change a static avatar to a animated avatar (or another animated avatar if I already have one equipped), it produces a ``this was 0xFFFFFFFFFFFFFFEF.`` crash.

devenv_9BqDB4lH2d

The call stack leads to the ``delete pAvatar;`` portion of ``CAvatarImage::OnHTTPRequestCompleted``. This DOES work if the ``delete pAvatar;`` line is commented, but it corrupts some animated avatars as shown below, and I originally thought this was why ``delete pAvatar;`` was added to begin with.

image

I didn't know if this is the fault of the PR however. My mod disables the TF2 main menu in favor of the GameUI source menu, so that change could have been a factor here. Since the normal menu was just a preprocessor definition away in my mod base, I re-enabled it, recompiled it, and it didn't crash, although it had the same corruption issue in the scoreboard as before.

image

I worry that this PR (in its current state) could crash in HL2MP due to the above issue, as that doesn't contain any animated avatar code anywhere but the scoreboard.

I'm not sure what's causing the corruption problem too, some avatars work fine while others break. It seems to be scoreboard related but i'm not sure. The effected avatars DO work when launching the mod, but corrupt when switching to them from Steam.

copperpixel:
I have tested this again using stock SDK2013 on two vastly different setups and although I couldn't replicate your crash when destroying pAvatar, I did notice some artifacts on GIFs when GameUI isn't overriden (on TF tested by commenting out line 330 in `tf_hud_mainmenuoverride.cpp`), which right now doesn't make any sense to me and I'll need to do further investigation on it.

I'm gonna clone your fork later and see why is the crash happening exactly, however my blind guess is that a change made in Valve's STL makes this specific scenario segfault.

I'll update this thread when I find anything.

copperpixel:
The artifacting should be fixed now, caused by the previous framebuffer not being initialized correctly sometimes making disposal not work properly.

Fix just ensures that it is set with the background color.

copperpixel:
I have also tested your mod however I never managed to get the delete pAvatar crash you described, both with legacy mainmenu disabled and enabled. Maybe I didn't have some preprocessor definitions you had turned on?

Bitl:
> I have also tested your mod however I never managed to get the delete pAvatar crash you described, both with legacy mainmenu disabled and enabled. Maybe I didn't have some preprocessor definitions you had turned on?

After I sent the bug report i put an ifndef for BDSBASE_LEGACY_MAINMENU around the "delete pAvatar" line, which suppresses the crash so it works without crashing in each instance

Bitl:
Just tested the bug fix and the artifacting still seems to be a problem. I'll try rebuilding the solution to see if that fixes the issue and update this

image

EDIT: this seems to be fixed with the Space Core avatar, but it appears to still not be fixed with others (like the aperture desk job avatar)

image

The purple Cat Face avatar was one of the affected avatars from before the fix. Although now it got a black background rather than the light bluish-yellow background it had before

image

copperpixel:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1380#issuecomment-3042199798

I did comment out the ifndef when testing, should've specified that, sorry.

> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1380#issuecomment-3042208547

I'm still debugging this however the broken background in GIFs seems to only occur in TF without the GameUI override. Both normal TF and HL2MP display them fine. (EDIT: The only difference I noticed between non-GameUI override TF and HL2MP is that TF calls `CGIFHelper::OpenImage` during the loading screen, so I'm guessing it has to do something with that)

copperpixel:
image
image

The background bug should be resolved now.

Bitl:
I can confirm that it works properly now!

copperpixel:
Went over and improved some things that I have initially missed:
- Reset animation speed for very fast GIFs to match web browsers (technically old behavior was more correct, but it's probably preferred for avatars to act the same way as on Steam)
- Optimizations in `CGIFHelper::GetRGBA`, also made it a bit easier to look at
- Overhauled animated avatar handling in `CAvatarImage`, e.g. reference counting is performed automatically, faster cache lookup

copperpixel:
Reduced memory usage by around 80% by storing GIF textures in DXT1 format instead of using raw RGBA8888 ~~(drawback being more processing needs to be done, but I think the trade off is worth it)~~. GIFs are processed on a separate worker thread now.

~~`CGIFHelper` also now supports exporting to other 8-bit channel texture formats.~~

Sub-Zero1324:
Is there a way to add the animated Avatar frames that will work on animated avatars and non animated avatars?

copperpixel:
> Is there a way to add the animated Avatar frames that will work on animated avatars and non animated avatars?

You'd need to implement support for reading APNG data into materials and draw them in CAvatarImage::Paint. The URL for a user's avatar frame image can be received the same way as animated avatars.

PR #1381, opened 2025-06-25
[TF2][MVM] Allow non giant robots and humans to properly gib in man vs machine

TheShermanTanker:
# Description

In man vs machine, you will quickly notice that no matter what you do, nothing aside from Giants will ever gib. The reason this is the case is that [Valve disabled gibbing in this mode a long time ago due to severe issues man vs machine had with gibbing](https://www.youtube.com/watch?v=Aqu2Lb2G2ME). This special case for man vs machine can be found within CTFPlayer::ShouldGib. To my knowledge, this bug has been fixed ages ago (By that I mean when I remove this special case stopping gibbing from happening in man vs machine, the old bug that caused Valve to remove gibbing in the first place doesn't resurface. Indeed, libtf2mod [also now only contains logic to remove this check from the game, and no longer contains the old fix it used to have](https://github.com/sigsegv-mvm/libtf2mod/blob/master/patch/mvm_enable_human_gibbing.c)), meaning that this check to stop gibbing in man vs machine is now redundant and can be removed. This restores proper gibbing to man vs machine, and brings this mode closer to parity with the other modes.

This is a visual only change, so game balance is not affected whatsoever. Regardless, this pull request can still be closed if it is decided that this would be detrimental to the game for some reason.

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -10910,10 +10910,6 @@ bool CTFPlayer::ShouldGib( const CTakeDamageInfo &info )
 			return true;
 	}
 
-	// normal players/bots don't gib in MvM
-// 	if ( TFGameRules()->IsMannVsMachineMode() )
-// 		return false;
-
 	// Suicide explode always gibs.
 	if ( m_bSuicideExplode )
 	{
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -10911,8 +10911,8 @@ bool CTFPlayer::ShouldGib( const CTakeDamageInfo &info )
 	}
 
 	// normal players/bots don't gib in MvM
-	if ( TFGameRules()->IsMannVsMachineMode() )
-		return false;
+// 	if ( TFGameRules()->IsMannVsMachineMode() )
+// 		return false;
 
 	// Suicide explode always gibs.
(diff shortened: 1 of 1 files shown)


src/game/server/tf/tf_player.cpp
```diff
@@ -10911,8 +10911,8 @@ bool CTFPlayer::ShouldGib( const CTakeDamageInfo &info )
 	}
 
 	// normal players/bots don't gib in MvM
-	if ( TFGameRules()->IsMannVsMachineMode() )
-		return false;
+// 	if ( TFGameRules()->IsMannVsMachineMode() )
```
ZzZombo: Please, just remove the code.

TheShermanTanker:
I have addressed the review comment and removed the code entirely.

FrozenDragon0:
Should also mention, in the vanilla game the only weapon that's able to gib the small robots is the cow mangler's charged shot.

TheShermanTanker:
Sorry, is there anything else I have to do on my end to possibly get this in?

TheShermanTanker:
@EricS-Valve Sorry for paging for you directly, but is there anything else I need to do on my end?

FrozenDragon0:
> @EricS-Valve Sorry for paging for you directly, but is there anything else I need to do on my end?

There is nothing you can do but wait I'm afraid, it's completely random as to what pull requests Eric chooses to merge.

Issue #683, opened 2025-02-20
[HL2] Vertices stretch toward world origin on specific citizen models


Blixibon:
### Describe the bug
The following models in Half-Life 2 have random vertices seemingly stretching to the map origin and other positions:
* `models/humans/group01/male_02.mdl`
* `models/humans/group02/male_02.mdl`
* `models/humans/group03m/male_02.mdl`
* `models/humans/group03m/male_04.mdl`
* `models/humans/group03m/male_05.mdl`

The issue becomes less intense as the LOD level increases (i.e. as the player gets farther away). This issue appears to exist in both HL2 and HL2:DM, although the models involved are not used in HL2:DM by default.

This is on the latest version of the Source 2013 `master` branch.

### Steps to reproduce
1. Spawn an entity which uses one of the above models. The following command creates a `prop_dynamic` with the first listed model: `prop_dynamic_create "humans/group01/male_02"`
2. Walk around the model in close proximity
3. See vertices stretching toward the world origin

### Media
Here is a short video demonstrating the issue:

https://github.com/user-attachments/assets/258e2181-d29e-461d-94de-8c9f7126b38a

Blixibon:
After further investigation, the issue appears to be coming from the models' mouths. The issue stops occurring when `r_teeth` is disabled.

I also confirmed that this issue occurs with other users as well, although it does not appear to be present on Linux.

misyltoad:
You might be able to get more info if you compile your own shader DLL containing teeth and play around with this.

But I will add this to my backlog of things to look at.

TotallyMehis:
Has anybody found a proper fix for this, perchance?

PR #1382, opened 2025-06-25
[HL2MP] Fix Teeth Vertices stretch on specific citizen models

JohnMega:
### Bug
Fixes a bug, discribed in the issue #683 
Bug: Some citizen models have random teeth vertices stretching to the map origin and other positions.

### Media

Video demonstrating the bug:

https://github.com/user-attachments/assets/f659da80-6b83-4e91-8b1f-50e4de6ce594

Video demonstrating the fix:

https://github.com/user-attachments/assets/9f47369a-e198-4c76-840b-3f25e5c99c5e


ficool2:
This is a hack and certainly not the right way to fix this

blackletum:
> This is a hack and certainly not the right way to fix this

what would the proper way to fix it entail?

PR #1385, opened 2025-06-28
[TF2] Allow server admins to change powerup mode convars without forcing cheats to be enabled.

bijlpieter:
A few years ago, Valve updated powerup mode with some very significant balance changes that:
- forcibly autobalances players that are performing well, without notice; and
- puts players that are performing well into a special dominant mode, forcing them to take minicrit damage and drastically reducing their powerup effectiveness, for 10 minutes.

The ConVars controlling this behavior are currently defined with the `FCVAR_CHEAT` flag, meaning that changing them only takes effect if `sv_cheats` is set to 1.

This proposed change would allow third party server admins to change these settings without the need to also allow cheats, meaning that those who prefer to play without the balance changes can do so.


bijldavid:
Good change, I like 👍

samoalis:
Excellent suggestion for mannpower

PR #1386, opened 2025-06-28
Fix Wrangler shield not moving with parented sentrygun

ficool2:
Wrangler shields don't attach to sentryguns that are parented to a moving object, causing them to float ominously. This PR addresses that by parenting the shield to the sentrygun as well. This is a common problem with community maps that have moving platforms, vehicles etc

Before:

https://github.com/user-attachments/assets/a2287103-b0a6-4ad5-8786-9ce4ed67d9ac

After:

https://github.com/user-attachments/assets/31ce13c9-8c73-4a8b-8f31-1a398205f6fb

### DIFF
--- a/src/game/client/tf/c_obj_sentrygun.cpp
+++ b/src/game/client/tf/c_obj_sentrygun.cpp
@@ -60,7 +60,7 @@ C_ObjectSentrygun::C_ObjectSentrygun()
 	m_nShieldLevel = SHIELD_NONE;
 	m_nOldShieldLevel = SHIELD_NONE;
 	m_hLaserBeamEffect = NULL;
-	m_pTempShield = NULL;
+	m_hShieldModel = NULL;
 	m_bNearMiss = false;
 	m_flNextNearMissCheck = 0.f;
 
@@ -312,7 +312,7 @@ void C_ObjectSentrygun::SetDormant( bool bDormant )
 	if ( IsDormant() && !bDormant )
 	{
 		// Make sure our shield is where we are. We may have moved since last seen.
-		if ( m_pTempShield )
+		if ( m_hShieldModel )
 		{
 			m_bRecreateShield = true;
 			m_bRecreateLaserBeam = true;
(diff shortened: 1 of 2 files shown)


src/game/client/tf/c_obj_sentrygun.cpp
```diff
@@ -758,4 +755,49 @@ const char* C_ObjectSentrygun::GetStatusName() const
...
+// Purpose:
+//-----------------------------------------------------------------------------
+void C_SentrygunShield::StartFadeOut( float flDuration )
+{
+	SetRenderMode( kRenderTransTexture );
 
+	m_flFadeOutStartTime = gpGlobals->curtime;
+	m_flFadeOutEndTime = gpGlobals->curtime + flDuration;
+	SetNextClientThink( gpGlobals->curtime );
+}
```
TheShermanTanker, line 803: Newline deletion intentional?

FlaminSarge:
Would it be worth asking for an engine change for AttachTentToEntity in the same way AttachTentToPlayer exists, instead of doing this? Or adding a parent param to SpawnTempModel? 
~~In fact, it's looking like AttachTentToPlayer doesn't actually care that the entity passed in isn't a client, so that might work already?~~ _one_ check to remove to get that working, I guess

Also, do maps allow players to build sentries on moving platforms and properly parent them to the moving platforms once built? Or is that logic coming from vscript/etc and not built into the game?

ALIEN31ITA:
> Would it be worth asking for an engine change for AttachTentToEntity in the same way AttachTentToPlayer exists, instead of doing this? Or adding a parent param to SpawnTempModel? ~In fact, it's looking like AttachTentToPlayer doesn't actually care that the entity passed in isn't a client, so that might work already?~ _one_ check to remove to get that working, I guess
> 
> Also, do maps allow players to build sentries on moving platforms and properly parent them to the moving platforms once built? Or is that logic coming from vscript/etc and not built into the game?

There's func_doors, breakable and _TANK_train that allow you to build on top of them
then i assume via the "object_built" event a trace gets fired aiming down from the building and called Setparent, that's the best approach i could think of, and used on one of my maps.

FlaminSarge:
> There's func_doors, breakable and _TANK_train that allow you to build on top of them
then i assume via the "object_built" event a trace gets fired aiming down from the building and called Setparent, that's the best approach i could think of, and used on one of my maps.

So parenting buildables to moving objects does not occur natively and needs to be set up by the map/mod in some way?
+I wonder if there's a way to keep handling this as a Tempent instead of as a Client Entity (though I guess from a performance perspective there really isn't that much difference).

gaelcoral:
Hi, I noticed that the Wrangler shield outline is now rendered when it shouldn't be
image

bakugo:
> Hi, I noticed that the Wrangler shield outline is now rendered when it shouldn't be

Making the shield an entity parented to the sentry unfortunately means that it now inherits the glow effect from the sentry itself. The shield and the sentry are rendered onto the stencil buffer as a single object, too, so it ends up effectively hiding the sentry inside.

Logic is in `CGlowObjectManager::GlowObjectDefinition_t::DrawModel()`. That function skips drawing move children that have their own glow object, so this might be fixable by creating a `CGlowObject` for the shield with both `bRenderWhenOccluded` and `bRenderWhenUnoccluded` set to false.

PR #1389, opened 2025-06-28
all games: Make `tf_delete_temp_files` available in non-TF games

copperpixel:
# Description
This PR renames `tf_delete_temp_files` ConVar to `cl_delete_temp_files` and makes it available in other Source2013 games (CStrike, DoD, HL1MP and HL2MP) than just Team Fortress 2.

**NOTE:** To ensure that the delete temp files setting still works in TF advanced options menu, any references to `tf_delete_temp_files` must be changed to `cl_delete_temp_files` in `cfg/user.scr`.


FlaminSarge:
~~I may be wrong about this, but I could have sworn there was a way to hook one cvar up to another in the case of cvar renames, so that the old cvar still functions if necessary (re: needing to update user.scr). Similar to how entity classnames/etc can be remapped in server code.~~ I'm probably mistaken

Bitl:
user_default.scr must be changed alongside user.scr for fresh mod/game installs to have the proper option

Dessensce:
Garry's Mod has "gmod_delete_temp_files", i'd assume these are the same on the back end, just different names. so i don't see any reason why Valve wouldn't do this.

copperpixel:
Left 4 Dead also has a `l4d_delete_temp_files`

PR #1390, opened 2025-06-29
[TF2] Add CVar to adjuct spectator bar alpha

horiuchii:
# Description
This PR adds `cl_spec_baralpha`, a CVar that sets the transparency of the spectator bars.
These bars are often seen as obtrusive, with the only way to reduce their presence is by modifying how tall they are, which in the case of the top bar, is directly linked to the position of the kill feed and engineer/spy building status y position when spectating.

Being able to freely modify the alpha of these bars would allow players using the default hud to improve their spectating 
experience hassle free and custom hud creators to be more creative without the fear of making new elements that occupy the same space as existing elements.

The CVar defaults at 200, which should be a good compromise between the bars being visible without being obtrusive to vision.

### DIFF
--- a/src/game/client/tf/vgui/tf_spectatorgui.cpp
+++ b/src/game/client/tf/vgui/tf_spectatorgui.cpp
@@ -68,7 +68,7 @@ CTFSpectatorGUI *GetTFSpectatorGUI()
 //-----------------------------------------------------------------------------
 ConVar cl_spec_carrieditems( "cl_spec_carrieditems", "1", FCVAR_ARCHIVE, "Show non-standard items being carried by player you're spectating." );
 
-void HUDTournamentSpecChangedCallBack( IConVar *var, const char *pOldString, float flOldValue )
+void HUDSpecInvalidateLayout( IConVar *var, const char *pOldString, float flOldValue )
 {
 	CTFSpectatorGUI *pPanel = (CTFSpectatorGUI*)gViewPortInterface->FindPanelByName( PANEL_SPECGUI );
 	if ( pPanel )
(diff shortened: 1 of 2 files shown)


src/game/client/tf/vgui/tf_spectatorgui.cpp
```diff
@@ -68,15 +68,16 @@ CTFSpectatorGUI *GetTFSpectatorGUI()
...
+void HUDSpecInvalidateLayout( IConVar *var, const char *pOldString, float flOldValue )
 {
 	CTFSpectatorGUI *pPanel = (CTFSpectatorGUI*)gViewPortInterface->FindPanelByName( PANEL_SPECGUI );
 	if ( pPanel )
 	{
 		pPanel->InvalidateLayout( true, true );
 	}
 }
-ConVar cl_use_tournament_specgui( "cl_use_tournament_specgui", "0", FCVAR_ARCHIVE, "When in tournament mode, use the advanced tournament spectator UI.", HUDTournamentSpecChangedCallBack );
+ConVar cl_use_tournament_specgui( "cl_use_tournament_specgui", "0", FCVAR_ARCHIVE, "When in tournament mode, use the advanced tournament spectator UI.", HUDSpecInvalidateLayout);
```
henke37, line 79: Copy past error?

henke37:
Normally the default should preserve existing behavior unless there is a strong case for changing the behavior.

src/game/client/tf/vgui/tf_spectatorgui.cpp
horiuchii: `HUDTournamentSpecChangedCallBack()` was renamed to `HUDSpecInvalidateLayout()`

PR #1397, opened 2025-07-05
[TF2] Fix Buggy movement when colliding with players

JohnMega:
### Bug

When colliding with players it tends to produce prediction errors: A player gets stuck in another player.

### Media

Video demonstrating the bug:

https://github.com/user-attachments/assets/d5759744-3556-4e7c-a436-2d9204d343ee

Video demonstrating the fix:

https://github.com/user-attachments/assets/59356753-7a90-41f4-8298-d5c768b0d3a6


RiverHornet3209 (approved):

ficool2:
This is not the correct fix for the issue and is just masking away the problem, the prediction errors are due to a lack of networking precision in player origin, this was fixed in later games by sending origins at bit-level precision

Adding a new `CONTENTS` flag just for this hack is also a bad way of going about it

PR #1399, opened 2025-07-06
[HL2MP] Fix empty icons in weapon selector

JohnMega:
### Bug

There's a blank space in the 1st slot for one of the melee weapons (depending on the team), which isn't cycled when I am selecting weapons, but still displayed.

### Media

image1

image2


Bitl:
This isn't a bug, this is an intentional design choice. This can be disabled with the hud_showemptyweaponslots ConVar.

PR #1400, opened 2025-07-06
[VRAD] Fixed VRAD always forcing verbose to true.

Unusuario2:
# Description
Self explanatory.


ficool2:
This was likely done on purpose, disabling verbose mode hides the map statistics after finishing a compile. Those are useful to see how close the map is to limits

Unusuario2:
Fixed. Now VRAD prints the Bounce Msg information and also the PrintBSPFileSizes function.

PR #1402, opened 2025-07-08
tf: Fix mouse mode not working in MvM scoreboard

copperpixel:
# Description
This PR makes mouse mode properly function in the Mann vs Machine scoreboard, allowing players to check detailed stats of their teammates and use the right click context menu.

A change in `SectionedListPanel` was needed to be made so that enable mouse input state was correctly updated in it's items.

Demonstration video:

[mvm_scoreboard_contextmenu_c.webm](https://github.com/user-attachments/assets/be4f6ea2-aedd-449c-9bd7-44b6ef4b8c89)

Closes issues:
- https://github.com/ValveSoftware/Source-1-Games/issues/2061
- https://github.com/ValveSoftware/Source-1-Games/issues/7375

### DIFF
--- a/src/game/client/tf/tf_hud_mann_vs_machine_scoreboard.cpp
+++ b/src/game/client/tf/tf_hud_mann_vs_machine_scoreboard.cpp
@@ -9,6 +9,7 @@
 #include <filesystem.h>
 #include <time.h>
 #include "tf_lobby_server.h"
+#include <vgui_controls/ScrollBarSlider.h>
 
 using namespace vgui;
 
@@ -271,6 +272,20 @@ char *ConvertScoreboardValueToString( int iValue )
 //-----------------------------------------------------------------------------
 void CTFHudMannVsMachineScoreboard::UpdatePlayerList () 
 {
+	int iSelectedPlayerIndex = GetLocalPlayerIndex();
+
+	// Save off which player we had selected
+	int itemID = m_pPlayerList->GetSelectedItem();
+
+	if( itemID >= 0 )
+	{
+		KeyValues* pInfo = m_pPlayerList->GetItemData( itemID );
+		if( pInfo )
+		{
+			iSelectedPlayerIndex = pInfo->GetInt( "playerIndex" );
+		}
+	}
+
 	m_pPlayerList->ClearSelection();
 	m_pPlayerList->RemoveAll();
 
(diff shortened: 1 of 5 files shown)


src/game/client/tf/tf_hud_mann_vs_machine_scoreboard.cpp
```diff
@@ -673,3 +693,25 @@ void CTFHudMannVsMachineScoreboard::UpdatePopFile( void )
 	}
 }
 
+//-----------------------------------------------------------------------------
+void CTFHudMannVsMachineScoreboard::InitializeInputScheme( bool bUseMouse )
+{
+	SetKeyBoardInputEnabled( false );
```
FlaminSarge, line 705: This might need
```
	if ( !IsVisible() )
		return;
```
 that `CTFClientScoreBoardDialog::InitializeInputScheme` also has, as well as the handling for if m_pPlayerList is missing

src/game/client/tf/vgui/tf_clientscoreboard.cpp
```diff
@@ -2298,7 +2304,12 @@ SectionedListPanel *CTFClientScoreBoardDialog::GetSelectedPlayerList( void )
 	SectionedListPanel *pList = NULL;
 
 	// navigation
-	if ( m_pPlayerListBlue->GetSelectedItem() >= 0 )
+	if ( m_bIsPVEMode )
+	{
+		// only one list in MvM
+		pList = m_pMvMScoreboard->GetPlayerList();
```
FlaminSarge, line 2310: Could this whole setup be simplified if m_pPlayerListRed was the sole list being used in MVM? As in, CTFHudMannVsMachineScoreboard.m_pPlayerList is assigned to m_pPlayerListRed at some point? A bit hacky but might work.

copperpixel: My main concern with this is that it could potentially break custom HUDs as the MvM player list panel signature would get changed to "RedPlayerList" from "MvMPlayerList" in the scheme file (unless there is a way to circumvent that which I'm unaware of)

FlaminSarge: No, I think you're probably right. There's probably a way to do it but not worth trying to hack it together.

PR #1405, opened 2025-07-08
[TF2] Remove ammo metal prediction if the weapon has add_onhit_ammo

horiuchii:
# Description

This PR removes ammo removal prediction when the weapon ammo type is TF_AMMO_METAL and it has the `add_onhit_addammo` attribute.
This fixes an issue where when firing the Widowmaker a -30 and (on successful hit) a +30 will overlap, creating issues where numbers become hard to read or unreadable.

This changes will delay the resulting ammo when the shot is registered into a single metal ammo account delta change, resulting in a clean and readable number.

https://github.com/user-attachments/assets/6ebbe5e3-757d-40d1-81a4-c6904fcd4e94


Platina6978:
I love these small QoL changes being produced here on the SDK, really hope they get pulled into the live game eventually!

wgetJane:
looks good, but imo if possible it would be more ideal for the metal counter itself to stay predicted, but just unpredict the metal change number

PR #1407, opened 2025-07-09
[ALL] Make trigger_hurt double w/ forgiveness actually forgive

horiuchii:
# Description

This PR fixes an issue with trigger_hurt's Doubling w/forgiveness damage model.

Currently, the trigger doesn't check for whether the curtime is past the forgiveness reset time before dealing damage and pushing forgiveness reset time into the future.

Now, curtime is instead checked to see if it has passed the forgiveness reset time at the start of the think, preventing this.

### DIFF
--- a/src/game/server/triggers.cpp
+++ b/src/game/server/triggers.cpp
@@ -834,6 +834,12 @@ void CTriggerHurt::EndTouch( CBaseEntity *pOther )
 #define TRIGGER_HURT_FORGIVE_TIME	3.0f	// time in seconds
 int CTriggerHurt::HurtAllTouchers( float dt )
 {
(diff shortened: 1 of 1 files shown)


ficool2:
This might cause issues with maps that depend on the broken behavior, should add a new keyvalue for this instead

horiuchii:
>Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1407#issuecomment-3052519908

Made the fix optional by moving `DAMAGEMODEL_DOUBLE_FORGIVENESS` down in the enum by 1 and having `DAMAGEMODEL_DOUBLE` take it's place.

FGDs probably should to be updated to reflect this so that's now in the PR description.

PR #1408, opened 2025-07-09
[TF2] Fix MoveableSubPanel showing while dead

horiuchii:
# Description
Fixes ValveSoftware/Source-1-Games/issues/4155

Adds an extra check to CTFPlayer::CanPickupBuilding() to see if the player is dead, if they are, they can't pickup the building, and shouldn't display the MovableSubPanel.


src/game/shared/tf/tf_player_shared.cpp
```diff
@@ -12339,6 +12339,9 @@ bool CTFPlayer::CanPickupBuilding( CBaseObject *pPickupObject )
 	if ( pPickupObject->GetUpgradeLevel() != pPickupObject->GetHighestUpgradeLevel() )
 		return false;
 
+	if ( !IsAlive() )
```
ZzZombo, line 12342: This check should be further up. For optimization the more complex checks should be gated behind the more trivial ones.

PR #1409, opened 2025-07-09
[TF2] Fix sentry "buffering" wrangler inputs

horiuchii:
# Description
Fixes ValveSoftware/Source-1-Games/issues/4237 by setting `m_bFireNextFrame` and `m_bFireRocketNextFrame` to false after when the sentry would've attacked that think.
If it did attack that think, it's already false, if it didn't, it shouldn't be true for the next think.


blackletum:
I actually just came across this while investigating my own fix I had made.

My fix was very similar but in `tf_obj_sentrygun.h` I changed:

`void	ClearTarget( void ) { m_hEnemy = NULL; }`

into

`void	ClearTarget( void ) { m_hEnemy = NULL; m_bFireNextFrame = false; m_bFireRocketNextFrame = false; }`

Which I feel might be a better way to handle this, but I'd like to hear your thoughts on it.

PR #1416, opened 2025-07-12
[TF2] Fix Force-A-Nature, Soda Popper and Shortstop only fire a single crit tracer visually

JohnMega:
### Bug

Force-A-Nature, Soda Popper and Shortstop only create a single crit tracer effect when they fire with crits, only one pellet gets a tracer.
The problem was the "Batch Particle Systems" enabled flag in some of the tracer effects. If this flag is disabled in these effects, everything returns to normal.

This PR also closes [#7384](https://github.com/ValveSoftware/Source-1-Games/issues/7384)

### Media

 **Force-A-Nature:**

https://github.com/user-attachments/assets/05d6a6e4-8d94-4648-80ea-ba035664add1

**Soda Popper:**

https://github.com/user-attachments/assets/66e40865-9d9c-4f8c-b0b1-03db0118a00e

**Shortstop:**

https://github.com/user-attachments/assets/0a1c5328-42c9-47c4-99e7-ea98621ba14c

### DIFF
--- a/src/game/shared/particle_property.cpp
+++ b/src/game/shared/particle_property.cpp
@@ -113,6 +113,7 @@ CNewParticleEffect *CParticleProperty::Create( const char *pszParticleName, Part
 // Purpose: Create a new particle system and attach it to our owner
 //-----------------------------------------------------------------------------
 static ConVar cl_particle_batch_mode( "cl_particle_batch_mode", "1" );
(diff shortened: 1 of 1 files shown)


ficool2:
This should be fixed in the .pcfs themselves, not by hardcoding it by name in code

JohnMega:
> This should be fixed in the .pcfs themselves, not by hardcoding it by name in code

Of course, the correct solution would be to edit this flag directly in the effect itself. But within this repository, it is not possible to edit the effect itself.

PR #1420, opened 2025-07-12
[TF2] Allow the Shortstop to shove while reloading

Bradasparky:
# Description
This PR allows the Shortstop to shove during a reload, thereby canceling the reload. You can shoot while reloading, so why not shove? I've heard others who shared this sentiment a lot ever since the addition of the shove, so figuring out that it's been possible all along is really cool.

**Note:** I want to briefly mention that the code block checking for `m_bInReload` was taken from `CTFWeaponBase::ItemBusyFrame()` where `pOwner->m_flNextAttack = gpGlobals->curtime`  allows for `ItemPostFrame()` to run, which then calls the shove function `Push()` . If you find something wrong with this implementation, please let me know. Thank you.

https://github.com/user-attachments/assets/2866315f-c183-4ab8-9724-0f7bd594c27c

### DIFF
--- a/src/game/shared/tf/tf_weapon_pistol.cpp
+++ b/src/game/shared/tf/tf_weapon_pistol.cpp
@@ -200,6 +200,33 @@ void CTFPistol_ScoutPrimary::Push( void )
 // Purpose: 
 // Input  :  - 
 //-----------------------------------------------------------------------------
+void CTFPistol_ScoutPrimary::ItemBusyFrame()
+{
+	BaseClass::ItemBusyFrame();
+
+	CTFPlayer *pOwner = ToTFPlayer( GetOwner() );
+	if ( !pOwner )
+		return;
+
+	if ( !m_bInAttack2 )
+		return;
+
+	// These are checked in SecondaryAttack too, but reload shouldn't be aborted if these fail
+	if ( !CanAttack() || m_flNextSecondaryAttack > gpGlobals->curtime )
+		return;
+
+	if ( m_bInReload )
+	{
+		AbortReload();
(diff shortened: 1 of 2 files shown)


src/game/shared/tf/tf_weapon_pistol.cpp
```diff
@@ -196,6 +196,33 @@ void CTFPistol_ScoutPrimary::Push( void )
 #endif
 }
 
+//-----------------------------------------------------------------------------
+// Purpose: 
+// Input  :  - 
+//-----------------------------------------------------------------------------
```
ZzZombo, line 202: Please either fill the comment out or remove it altogether. Why add this boilerplate and not even fill it?

Bradasparky: The point of these boilerplate comment blocks are to leave a space for future comments. Just because I don't leave one now, doesn't mean there won't be one from another developer in the future.

ZzZombo: Sorry, but it is rather bad practice to add boilerplate and not actually make use of it. This is not physical medium, there is no need to reserve space for future additions in the code. And why do you defer to some other unknown developers in an indeterminate future to fill it in when you're the author of the code? That's rather inconsiderate. Please reconsider your stance.

P. S.: I know the code throughout the codebase is littered by similar comment boilerplate, but all such instances should be cleaned up or actually given proper thought regarding their content, and not copied blindly everywhere.

Bradasparky: I've given this more thought and left a comment within the block.

PR #1424, opened 2025-07-13
[TF2] Auto-equip powerup canteen when ForceUpgrades is enabled (Freaky Fair)

Lizard-Of-Oz:
# Description

**TL;DR:** this PR auto-equips a free powerup canteen on cp_freaky_fair (and other non-MvM maps with Upgrades), a QoL change already applied to other gameplay-relevant action items, and even canteens in MvM.

When playing MvM, you get a powerup canteen as a default Action Item in your loadout menu that gets auto-equipped if your action slot is empty.
Similarly, you auto-equip a free Spellbook Magazine on maps that use Spells and the Grappling Hook in Mannpower.

This, however, isn't the case on non-MvM maps with Upgrades enabled via `ForceEnableUpgrades`. One such map, cp_freaky_fair, was added in Scream Fortress 2024.
You have to manually equip your powerup canteen in your loadout... Assuming you have one... Which will then remain equipped and prevent you from using the Spells on other Halloween maps.

This Pull Request fixes that and gives you a temporary powerup canteen if `ForceEnableUpgrades` is set to `2`, which is how non-MvM maps enable Upgrades.

This PR also adds mode `3` to `ForceEnableUpgrades`. This new mode enables the Upgrades but does NOT auto-equip the canteen.
This is made for potential maps that would want to use MvM money for purposes not related to Upgrades.
Because most (if not all) maps that currently use `ForceEnableUpgrades` do so specifically for the Upgrades, it's worth having canteens auto-equipped by default should Upgrades be enabled, unless specifically told not to.

image

### DIFF
--- a/src/game/shared/tf/tf_item_inventory.cpp
+++ b/src/game/shared/tf/tf_item_inventory.cpp
@@ -686,7 +686,7 @@ CEconItemView *CTFInventoryManager::GetBaseItemForClass( int iClass, int iSlot )
 		}
 
 		static CSchemaItemDefHandle pItemDef_MvMCanteen( "Default Power Up Canteen (MvM)" );
-		if ( TFGameRules() && TFGameRules()->IsMannVsMachineMode() && pItemDef_MvMCanteen )
+		if ( TFGameRules() && TFGameRules()->ShouldProvidePowerupBottle() && pItemDef_MvMCanteen )
 		{
 			stockActionItemDefIndices.AddToTail( pItemDef_MvMCanteen->GetDefinitionIndex() );
 		}
--- a/src/game/shared/tf/tf_gamerules.cpp
+++ b/src/game/shared/tf/tf_gamerules.cpp
@@ -2124,7 +2124,7 @@ bool CTFGameRules::GameModeUsesUpgrades( void )
 	if ( m_nForceUpgrades == 1 )
 		return false;
 
-	if ( m_nForceUpgrades == 2 )
+	if ( m_nForceUpgrades == 2 || m_nForceUpgrades == 3 )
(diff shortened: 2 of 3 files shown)


src/game/shared/tf/tf_item_inventory.cpp
mtxfellen, line 1: Should perhaps also add this to `CTFInventoryManager::SlotContainsBaseItems()` at line 648.
```cpp
if ( TFGameRules() && TFGameRules()->ShouldProvidePowerupBottle() )
	return true;
```

Mikusch:
Changing the current behavior of `ForceEnableUpgrades 2` is not a good idea, as it is also commonly used to enable the currency counter in the HUD. Any maps or scripts using this feature would now get the popup even if they don't want it.

This should really be its own input/netprop instead of overloading the meaning of `m_nForceUpgrades` and introducing potentially undesirable changes.

Lizard-Of-Oz:
The popup doesn't appear outside of MvM - the only change this does is replacing an empty action slot with a "stock" powerup canteen if the conditions are met.

If the behavior of existing `ForceEnableUpgrades` values _must_ be preserved, we could make it so that `ForceEnableUpgrades 3` gives the players a "stock" canteen, which would require 1 line change to this PR.
```diff
bool CTFGameRules::ShouldProvidePowerupBottle( void )
{
-	return GameModeUsesUpgrades() && m_nForceUpgrades != 3;
+	return GameModeUsesUpgrades() && m_nForceUpgrades != 2;
}
```
This change will provide a free canteen if the `m_nForceUpgrades` value is either (`3`) or (`0` and it's MvM).

PR #1426, opened 2025-07-14
[TF2] Separate Voice and Chat mutes

horiuchii:
# Description
This PR separates the option to mute players into two separate options, voice and chat. The mute menu and scoreboard context menu have been updated to display these options.

In order to achieve two separate mute options, the method of writing to `voice_ban.dt` was updated to include saving per player ban flags. All functions responsible for parsing mute statuses between players have been modified to use a ban flag parameter instead of a boolean.

Additionally, the `MutePlayerDialog.res` was updated to use new translation strings within `mod_tf_english.txt`

# Showcase
Mute Dialog

Scoreboard context menu mute

**Potential Volume Warning**

https://github.com/user-attachments/assets/49f69deb-2f49-4557-bbc4-c703284fdf0f


src/game/client/tf/vgui/drawing_panel.cpp
```diff
@@ -317,7 +317,7 @@ void CDrawingPanel::FireGameEvent( IGameEvent *event )
 			// If a player is muted for voice, also mute them for lines because jerks gonna jerk.
```
henke37, line 317: Is this comment still accurate? Looks like it probably is. But should it remain like this?

RiverHornet3209:
The video doesn't load

PR #1428, opened 2025-07-15
Fix steam input not working for the upgrade station menu

Ashe-tf:
# Description
The upgrade station menu has partial support for controllers, only for Accept, Back and Done. This adds support for viewing upgrades, changing weapon pages and the Refund button.

https://github.com/user-attachments/assets/bc99e511-eca0-4a03-a22c-93cc71ac174f

Sadly, this doesn't seem to work with commands but by binding to gamepad buttons ~and gamepad must be enabled in the options (no -nojoy too)~, which isn't ideal. I could do a better implementation if Valve releases inputsystem to the sdk.

imagen

### DIFF
--- a/src/game/client/tf/player_vs_environment/c_tf_upgrades.cpp
+++ b/src/game/client/tf/player_vs_environment/c_tf_upgrades.cpp
@@ -1197,11 +1197,11 @@ void CHudUpgradePanel::UpdateJoystickControls( void )
 	}
 
 	bool bUp = ::input->Joystick_GetForward() < 0.0f || ::input->Joystick_GetPitch() < 0.0f || vgui::input()->IsKeyDown( KEY_XBUTTON_UP ) || vgui::input()->IsKeyDown( KEY_UP ) || vgui::input()->IsKeyDown( STEAMCONTROLLER_DPAD_UP );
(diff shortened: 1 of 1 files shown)


ghost:
This game absolutely needs more controller support. As of now you cannot swap out weapons in your inventory, use the main menu or even close the mvm loosing panel. If theses could be adressed I think it would benefit steamdeck users greatly.

Ashe-tf:
> This game absolutely needs more controller support. As of now you cannot swap out weapons in your inventory, use the main menu or even close the mvm loosing panel. If theses could be adressed I think it would benefit steamdeck users greatly.

The main menu had controller support until JI, which hardcoded the main menu hud and valve didn't add support. If you wanna take a look on it, check tf_controls.cpp

Ashe-tf:
I've disabled the Joystick check, which is something other menus don't do, and now it should work regardless of the "Enable gamepad" check in the options

Ashe-tf:
I've done some changes 1. to prevent a very unlikely access violation crash. This one in particular was happening constantly because of a change ("hiding" the cursor, see bottom) I've ended up not publishing. In the end, it also ended up fixing a visual glitch that happens every time a player opens the panel.

Before:

https://github.com/user-attachments/assets/76c085e5-9220-47c7-aebd-074aabc85664

After:

https://github.com/user-attachments/assets/31459962-c584-43af-b312-5b13d87a6b70

2. Hide the highlight box when it isn't positioned correctly. This happens because of the crash fix, but, IIRC, it also happens in vanilla very rarely, afaik mostly when playing with high ping.
3. The panel can no longer be opened if the player has the main menu opened, to prevent some weird input stuff. The panel immediately opens after closing the menu anyways.

Notes: This is the change I wanted to implement, but this PR has already too much changes. If you want to use it, be my guest.

devenv_GdICeqicJP

PR #1433, opened 2025-07-17
[TF2] Add client ConVars for changing viewmodel position and angle offsets

Bradasparky:
# Description
Currently, the only option for modifying your viewmodel offset is `tf_use_min_viewmodels`, which is a hard coded `z` value. This PR allows players to modify their viewmodel offsets however they want. 

**Added ConVars**
\- `viewmodel_offset_x`
\- `viewmodel_offset_y`
\- `viewmodel_offset_z`
\- `viewmodel_offset_pitch`
\- `viewmodel_offset_yaw`
\- `viewmodel_offset_roll`

**Notes**
- Modified `tf_viewmodels_offset_override` to include angle arguments.
- When  `tf_use_min_viewmodels 1` is used with a non-zero value of `viewmodel_offset_z`, both offsets will be added together.
- Works with `+inspect` and weapon lowering during the game over round state or when in `TF_COND_PHASE`.

https://github.com/user-attachments/assets/38e420ab-4d99-497c-a127-115ec30cc6e6

### DIFF
--- a/src/game/shared/tf/tf_viewmodel.cpp
+++ b/src/game/shared/tf/tf_viewmodel.cpp
@@ -124,7 +124,15 @@ ConVar cl_gunlowerspeed( "cl_gunlowerspeed", "2", FCVAR_CLIENTDLL | FCVAR_CHEAT 
 
 ConVar tf_use_min_viewmodels( "tf_use_min_viewmodels", "0", FCVAR_ARCHIVE, "Use minimized viewmodels." );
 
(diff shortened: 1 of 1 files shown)


copperpixel:
IMO it would be better to do this in baseviewmodel_shared and apply it to other games similarly to Valve's implementation in Alien Swarm

Bradasparky:
> IMO it would be better to do this in baseviewmodel_shared and apply it to other games similarly to Valve's implementation in Alien Swarm

I'll look into doing that. I see that the Alien Swarm SDK's baseviewmodel_shared.cpp has the same positional offset convars.

PR #1436, opened 2025-07-19
[All VScript] Add ScriptOverlayMaterialEx

horiuchii:
(Reopened due to an accidental deletion of the original forked branch)

# Description

Closes https://github.com/ValveSoftware/Source-1-Games/issues/6112

Adds (Get/Set)ScriptOverlayMaterialEx() VScript functions, works similar to (Get/Set)ScriptOverlayMaterial() but takes an extra argument for the index in an array of materials to set.

The default script overlay array limit is 8 as that would grant an even 10 overlays to use, including the regular script overlay and r_screenoverlay/env_screenoverlay combinied.


JimmyPLightning:
This is desperately needed in my opinion. Communicating gameplay functionality through the HUD is super important and VScript's only tool for that right now is basically plain text that can barely be formatted.

PR #1439, opened 2025-07-20
tf: Network player's handedness preference to spectators

copperpixel:
# Description
Flipping your viewmodel's position changes where projectiles are shot from, which can look silly from the perspective of a spectator with different settings. This PR makes spectators see the same viewmodel orientation as you by sending them the whether you use flipped viewmodels flag that the server already stores.

Before:

[tf_networked_handedness_preference_before.webm](https://github.com/user-attachments/assets/025dd43b-3a01-452c-ade3-e290fca53006)

After:

[tf_networked_handedness_preference_after.webm](https://github.com/user-attachments/assets/1bbfae10-11d1-46c3-a839-f261cbc8cf36)

### DIFF
--- a/src/game/client/c_baseviewmodel.cpp
+++ b/src/game/client/c_baseviewmodel.cpp
@@ -19,6 +19,7 @@
 #include <KeyValues.h>
 #include "hltvcamera.h"
 #ifdef TF_CLIENT_DLL
+	#include "c_tf_player.h"
 	#include "tf_weaponbase.h"
 #endif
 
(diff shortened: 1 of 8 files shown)


Generalisk:
Would be nice if this feature is optional via A console command, like `cl_spectator_use_target_viewmodel_prefs`.

copperpixel:
> Would be nice if this feature is optional via A console command, like `cl_spectator_use_target_viewmodel_prefs`.

Personally I don't really see the point of toggling this off.
IIRC Counter-Strike 2 doesn't allow you to turn it off either.

PR #1440, opened 2025-07-21
[TF2] Treat being in background map as not in-game

horiuchii:
# Description
This PR makes it so that being in a Background Map doesn't count as being in-game for the main menu.
This means the call vote, mute players, request coach and report players buttons won't appear and the disconnect button will be there instead be the quit game button.

The player can still type `disconnect` in console to leave the background map as usual.

### DIFF
--- a/src/game/client/tf/tf_hud_mainmenuoverride.cpp
+++ b/src/game/client/tf/tf_hud_mainmenuoverride.cpp
@@ -963,7 +963,7 @@ void CHudMainMenuOverride::OnUpdateMenu( void )
 	// So try and do the least amount of work if nothing has changed.
 
 	bool bSomethingChanged = false;
-	bool bInGame = engine->IsInGame();
+	bool bInGame = engine->IsInGame() && !engine->IsLevelMainMenuBackground();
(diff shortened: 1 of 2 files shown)


henke37:
This addresses the issue by adding conditions to where the value is used. Should the condition be put into the IsInGame function instead?

horiuchii:
> https://github.com/ValveSoftware/source-sdk-2013/pull/1440#issuecomment-3117658790

This isn't possible as we don't have access to whatever file it's defined in.

RileyKun:
> This addresses the issue by adding conditions to where the value is used. Should the condition be put into the IsInGame function instead?

`IsInGame()` from `IVEngineClient` only exists as a wrapper for `CClientState::IsActive()` in the engine dll. It does not do any other additional checks behind the scenes.

mastercoms:
Can also add to CTFDemoSupport::Update on the first line.

Also this doesn't hide the background, I suggest touching up the file to change some bIsGame checks to also include a check for bBackgroundLevel, while keep bIsGame = engine->IsInGame() && !bBackgroundLevel

PR #1448, opened 2025-07-25
Fix point_camera transmit state in multiplayer

ficool2:
The newly-added map `koth_boardwalk` features "mirrors" in the map, which is an effect created with the `point_camera` entity
However, in populated servers this doesn't work quite right. The mirrors will sometimes display fine, other times it will be skipping frames, and sometimes it won't refresh at all.

This is caused by an oversight in the visibility code for point_camera. The intention is to only transmit the camera to players who are in PVS of either the camera or its link, but this visibility state is calculated before the engine runs transmit checks for players. and there is only one visibility state. When the engine runs the transmit checks for all players later, it will use this singular visibility state and it's value will be whatever was calculated for the *last* player that joined the server, making it inconsistent for everyone else.

This PR fixes that by storing a visibility state for all players and using that in the transmit checks, so now the engine will correctly transmit on a per-player basis.

Tested with 23 bots with `sv_stressbots 1` (which enables networking to bots) on koth_boardwalk

Before:

image

After:

image

### DIFF
--- a/src/game/server/point_camera.cpp
+++ b/src/game/server/point_camera.cpp
@@ -52,6 +52,9 @@ CPointCamera::CPointCamera()
 	m_bFogEnable = false;
 	m_bFogRadial = false;
 
+	// By default, transmit to everyone
+	m_bitsTransmitPlayers.SetAll();
+
 	g_PointCameraList.Insert( this );
 }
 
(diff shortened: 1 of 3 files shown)


src/game/server/point_camera.cpp
```diff
@@ -74,34 +77,43 @@ void CPointCamera::Spawn( void )
...
+		m_bitsTransmitPlayers.Clear( nPlayerIndex );
+}
 
 //-----------------------------------------------------------------------------
 // Purpose: 
 //-----------------------------------------------------------------------------
 void CPointCamera::SetActive( bool bActive )
 {
-	// If the mapmaker's told the camera it's off, it enforces inactive state
-	if ( !m_bIsOn )
```
henke37, line 100: This appears to be straight up removing functionality?

ficool2: This statement never has any effects, each call to SetActive uses the value of `m_bIsOn`

PR #1450, opened 2025-07-25
Freezecam mat_viewportscale fix

celisej567:
# Description
When you change mat_viewportscale on something lower then 1 freezecam doesnt map on screen correctly.

Ive made a fix/hack that helps preventing it to happen. But i think i can be done better then that.

Thats how it looks before and after a fix/hack:
изображение

изображение


ewillerup:
Nice. Does it show correctly in a screenshot?

celisej567:
wdym on screenshot? 
if you mean f12, yes it works correctly.
if you mean about those 2 images, the first one is how it was before fix, second is after (how it visually should be when you use mat_viewportscale lower then 1 and have mat_viewportupscale 1)

PR #1451, opened 2025-07-25
Fix mult_reload_time_while_healed in tf_weaponbase.cpp

ChaseTownsend:
# Description

Fixed the `mult_reload_time_while_healed` not working if healed by 2 or more healers


ChaseTownsend:
accidentally changed something i was not supposed to,
reverted that change

PR #1456, opened 2025-07-27
[TF] Fixed Ullapool Caber explosions not counting as melee kills and ignoring damage/radius attributes.

Bitl:
# Description
This PR adds 2 fixes for the Ullapool Caber:

- Fixed the Ullapool Caber's explosion not being counted as melee damage (for kill_refills_meter)
- Fixed the Ullapool Caber's explosion ignoring mult_dmg and mult_explosion_radius attributes

This PR does NOT fix the Ullapool Caber's mini-crit bug. That's already handled by #1120.

PR #1457, opened 2025-07-27
[TF] Implement caber kill taunt

Bitl:
# Description
This PR implements a taunt kill onto the Ullapool Caber. Currently, the Caber's taunt consists of the Demoman showing off the Caber, then swinging it. However, it doesn't explode on impact or damage anyone, which I felt was the intended effect.

This taunt works when the Caber is detonated or not, and will do the same damage (500). If the Caber hasn't been detonated, the player will also be damaged as well, similar to the Equalizer taunt.

This PR is meant to be used in conjunction with #1456, as the "detonated" version of the kill taunt registers as a melee kill.

# Videos:
Un-detonated:

https://github.com/user-attachments/assets/4e6965d6-a0d1-4cc0-a53e-db7ea8830625

Detonated:

https://github.com/user-attachments/assets/4b6ab298-4ca5-4b8c-a5cd-e4b76bfc4fda


Bitl:
Updated PR to work with latest SDK code

Bitl:
Converted to draft to sync it with the repo.

Bitl:
This PR will be further updated when and if #1702 gets implemented, as it includes further alterations to the Caber that cannot be merged here without it getting added.

PR #1459, opened 2025-07-27
[TF] Allow TFBots to use the Wrap Assassin's alt-fire

Bitl:
# Description
This PR fixes an oversight where bots won't use the Wrap Assassin's secondary fire, despite it being similar to the Sandman's.

Slightly inspired by rafradek's change here: https://github.com/rafradek/sigsegv-mvm/blob/master/src/mod/ai/wrapassassin_altfire.cpp

### DIFF
--- a/src/game/server/tf/bot/tf_bot.cpp
+++ b/src/game/server/tf/bot/tf_bot.cpp
@@ -4597,7 +4597,7 @@ Action< CTFBot > *CTFBot::OpportunisticallyUseWeaponAbilities( void )
 				}
 			}
 		}
-		else if ( weapon->GetWeaponID() == TF_WEAPON_BAT_WOOD )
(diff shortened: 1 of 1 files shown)


src/game/server/tf/bot/tf_bot.cpp
```diff
@@ -4597,7 +4597,7 @@ Action< CTFBot > *CTFBot::OpportunisticallyUseWeaponAbilities( void )
 				}
 			}
 		}
-		else if ( weapon->GetWeaponID() == TF_WEAPON_BAT_WOOD )
+		else if ( weapon->GetWeaponID() == TF_WEAPON_BAT_WOOD || weapon->GetWeaponID() == TF_WEAPON_BAT_GIFTWRAP )
 		{
 			// sandman
```
HyperionCoding: Update this too (L:4602)

PR #1461, opened 2025-07-27
[TF] Allow TFBots to use the Shortstop's shove

Bitl:
# Description
This PR enables bots to use the Shortstop's shove, using the same calculations used for the Pyro bots' compression blast. In this instance though, the shove only works if the Scout detects a player that's too close to him (as that's how the shove works).

### DIFF
--- a/src/game/server/tf/bot/behavior/tf_bot_behavior.cpp
+++ b/src/game/server/tf/bot/behavior/tf_bot_behavior.cpp
@@ -16,6 +16,7 @@
 #include "tf_weapon_flamethrower.h"
 #include "tf_weapon_sniperrifle.h"
 #include "tf_weapon_compound_bow.h"
+#include "tf_weapon_pistol.h"
 #include "bot/tf_bot.h"
 #include "bot/tf_bot_manager.h"
 #include "bot/behavior/tf_bot_behavior.h"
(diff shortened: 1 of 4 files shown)


src/game/server/tf/bot/behavior/tf_bot_behavior.cpp
```diff
@@ -1369,6 +1370,18 @@ void CTFBotMainAction::FireWeaponAtEnemy( CTFBot *me )
 
 		return;
 	}
+	else if ( myWeapon->IsWeapon( TF_WEAPON_HANDGUN_SCOUT_PRIMARY ) )
+	{
+		CTFPistol_ScoutPrimary *pPistolPrimary = assert_cast<CTFPistol_ScoutPrimary*>( myWeapon );
+		// watch for enemy projectiles heading our way
+		if ( pPistolPrimary->CanUsePush() && me->ShouldFireCompressionBlast() )
+		{
+			// bounce missiles with compression blast
```
ZzZombo: Please, if you copy code, make sure to update all comments to the new reality they are in.

src/game/server/tf/bot/tf_bot.cpp
```diff
@@ -4114,7 +4115,22 @@ bool CTFBot::ShouldFireCompressionBlast( void )
 		{
```
ZzZombo, line 4115: I don't think it's correct to hijack a function like this. I get that it contains the most of the logic that you'd like to use, but it is literally named `ShouldFireCompressionBlast()` but you now add shove processing into it; it is something that I'd prohibit within our codebase over here.

Please, refactor the common code into a helper function or functions, leave the original function's behavior the same, and make a new function for your code.

src/game/server/tf/bot/behavior/tf_bot_behavior.cpp
```diff
@@ -1369,6 +1370,18 @@ void CTFBotMainAction::FireWeaponAtEnemy( CTFBot *me )
...
+			me->PressAltFireButton();
+		}
+
+		return;
```
TheBetaM: This causes the bot to not fire the Shortstop at all, move the return into the PressAltFireButton scope.

src/game/server/tf/bot/tf_bot.cpp
Bitl: i'm not going to do this, there's too little code added that warrants a new function, and it doesn't change the original logic.

Bitl:
Keeping this as a draft as the scout tries to push players, but the scout tries to push the player too soon before he approaches him.

src/game/server/tf/bot/tf_bot.cpp
ZzZombo: You are adding code to a public **SDK** repository. This means countless other users will have to work with your additions down the line. In your own fork of this code you are free to do whatever, but here, please, do everything the proper way rather than the lazy way.

PR #1462, opened 2025-07-29
[TF] Fixed Gunslinger being able to pause melee hit combos when holstering the weapon.

Bitl:
# Description
This PR attempts to fix a long-standing exploit where the player can "pause" hit combos when holstering the Gunslinger, especially when done in conjunction with the double melee swing glitch (described in https://github.com/ValveSoftware/Source-1-Games/issues/4039). This doesn't fix the glitch mentioned, but it makes it much more difficult to get a earlier-than-usual crit with the weapon.

This also includes a minor formatting fix.

### DIFF
--- a/src/game/shared/tf/tf_weapon_wrench.cpp
+++ b/src/game/shared/tf/tf_weapon_wrench.cpp
@@ -393,11 +393,25 @@ void CTFRobotArm::Precache()
 	PrecacheModel( g_HACK_GunslingerEngineerArmsOverride );
 }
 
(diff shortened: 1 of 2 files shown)


bakugo:
How does this work, exactly? `m_flLastComboHit` is just a timestamp, it doesn't get "paused" when the weapon is not active.

EDIT: Figured it out, this can *only* happen in conjunction with the swing storage exploit, it cannot happen on its own. I'll see if I can find a fix for the actual exploit.

Bitl:
Closed as #1463 is a better solution for this issue

PR #1466, opened 2025-07-31
MAX_MENU_STRING increase

mourningsickness:
Please let us have more characters in the menus. Things like Sourcemod plugins interface heavily with the menu system, and need to be redesigned, stripped down, or paginated when it shouldn't really be needed. See pics for an example that just 512 -> 1024 would fix.

### Current implementation:
Screenshot from 2025-07-31 13-18-07

### Adding more lines with better formatting, but not filled text fields:
Screenshot from 2025-07-31 12-42-30

### Formatted, with more text.. too long:
Screenshot from 2025-07-31 12-43-19


Bitl:
What would happen if the limit was set to 2048 or even 4096 characters

mourningsickness:
hi, hoping this can be merged before the coming css update is pushed.. since it may be the last ever.. :)

PR #1470, opened 2025-08-02
[TF2] Fix ownerless sentry rockets dealing no damage

horiuchii:
# Description
This PR fixes ownerless sentry rockets dealing no damage.
This is fixed by having the scorer not override the attacker if the scorer is null.


wgetJane:
isnt this supposed to prevent players from firing a rocket or something at their teammates and then pressing a disconnect/retry bind to hit them with the now-ownerless projectile

PR #1472, opened 2025-08-04
[TF2] Reimplament sv_showhitboxes for listen servers only

Dessensce:
this just comments some lines that restrict sv_showhitboxes to debug builds, so it works in release builds, restricted to listen servers only, and requires the developer cvar to be 1 or higher. main reason for reimplementing this if so Cosmetic Creators can Easily test their cosmetics in game, and see the hitboxes in real time

image

### DIFF
--- a/src/game/server/gameinterface.cpp
+++ b/src/game/server/gameinterface.cpp
@@ -129,174 +129,174 @@ extern ConVar tf_mm_servermode;
 #include "replay/ireplaysystem.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


kvog-git:
It looks like your editor automatically formatted the whitespace and pointer alignment in the entire file.

Dessensce:
Cheers for pointing that out, should be fixed now

wgetJane:
what's with all the extra stuff? just remove the `#ifdef _DEBUG` lines imo

PR #1473, opened 2025-08-05
[TF2] Fix Diamondback getting revenge crits from earlier kills when switched onto

HyperionCoding:
# Description

Fixes the following bug:

> If a Spy performs backstabs or saps buildings with a different secondary weapon equipped, the Crits are retroactively awarded as soon as the Spy switches to the Diamondback via a Resupply locker or by picking up a dropped Diamondback.

[Demonstration by SquimJim](https://youtu.be/ozhXCz0_bf4?si=Pd_0Z6wGlZdvN4WO&t=415)


seanmcgeehan:
I don't think this should be merged. Many weapons work similarly (eyelander when switching loadouts) and spy is very weak. IMO we should be fixing the bugs that make spy weaker.

PR #1476, opened 2025-08-06
[TF2] add convars for server owners and/or players to hide the freezepanel showing held item

Pandaptable:
# Description
Adds convars to hide the following panel
panel example

`sv_freezepanel_hide_item 0`
`sv_freezepanel_hide_rune 0`

Mannpower
mannpower

Regular Match
normal match
You can also hide *only* runes
`sv_freezepanel_hide_rune 1`
hiding rune only
or hiding everything
`sv_freezepanel_hide_item 1`
`sv_freezepanel_hide_rune 1` (Ignored if you have no runes)
hiding everything


Bitl:
Why couldn't this have an option to disable the panels with the client?

PR #1477, opened 2025-08-06
[TF] Restore functionality of preserve_ubercharge attribute

Bitl:
# Description
This PR reimplements the preserve_ubercharge attribute cut in the Jungle Inferno update in 2017. This attribute was used for the original version of the Vita-Saw. This does not require any items_game.txt modifications to implement.

The input for the attribute is an integer which represents the percent of Ubercharge the Medigun should save upon the player's death. For example, the original Vita-Saw used 20%, so the integer calculated must be 20. The code will then calculate the value as a percentage and pass it into m_flChargeLevelToPreserve, where it'll be passed to m_flChargeLevel on reset.

This implementation uses the m_flChargeLevelToPreserve variable used for the ubercharge_preserved_on_spawn_max attribute. As a result, both attributes shouldn't be used at the same time.

An additional note: when working on this I noticed that the MSVC compiler has a bit of a rounding/floating point error when calculating the saved Ubercharge (ex. 20% of uber / 100.0% = 0.199999988 = 19% of uber), so WIN32 divides by 99% rather than 100%. MSVC and GCC should be able to both calculate the correct values as a result.

### DIFF
--- a/src/game/shared/tf/tf_weapon_medigun.cpp
+++ b/src/game/shared/tf/tf_weapon_medigun.cpp
@@ -291,6 +291,17 @@ void CWeaponMedigun::WeaponReset( void )
 	if ( TFGameRules()->State_Get() == GR_STATE_RND_RUNNING )
 	{
 		// This is determined via an attribute in SetStoredChargeLevel()
+		int iPreserveUberCharge = 0;
+		CALL_ATTRIB_HOOK_INT_ON_OTHER( pOwner, iPreserveUberCharge, preserve_ubercharge );
+		if ( iPreserveUberCharge > 0 )
+		{
+#ifdef _WIN32
+			m_flChargeLevelToPreserve = ( iPreserveUberCharge / 99.f );
+#else
+			m_flChargeLevelToPreserve = ( iPreserveUberCharge / 100.f );
+#endif
+		}
+		
 		m_flChargeLevel = Min( (float)m_flChargeLevel, m_flChargeLevelToPreserve );
 	}
 	else


src/game/shared/tf/tf_weapon_medigun.cpp
```diff
@@ -291,6 +291,17 @@ void CWeaponMedigun::WeaponReset( void )
...
 	{
 		// This is determined via an attribute in SetStoredChargeLevel()
+		int iPreserveUberCharge = 0;
+		CALL_ATTRIB_HOOK_INT_ON_OTHER( pOwner, iPreserveUberCharge, preserve_ubercharge );
+		if ( iPreserveUberCharge > 0 )
+		{
+#ifdef _WIN32
+			m_flChargeLevelToPreserve = ( iPreserveUberCharge / 99.f );
+#else
+			m_flChargeLevelToPreserve = ( iPreserveUberCharge / 100.f );
```
FlaminSarge:
```suggestion
			m_flChargeLevelToPreserve = Max( m_flChargeLevelToPreserve, iPreserveUberCharge / 99.f );
#else
			m_flChargeLevelToPreserve = Max( m_flChargeLevelToPreserve, iPreserveUberCharge / 100.f );
```

This should allow it to work with the other attribute, making it so that it prefers whichever one provides more saved uber.

FlaminSarge:
It may be worth looking back at the old implementation to see if it looked at the attribute value on death or on respawn; if you are trying to get it as close to the original as possible, it would be good to match that.

Bitl:
it seems to be very close to the original, the medigun is the only place it checks for the value

src/game/shared/tf/tf_weapon_medigun.cpp
```diff
@@ -291,6 +291,17 @@ void CWeaponMedigun::WeaponReset( void )
 	if ( TFGameRules()->State_Get() == GR_STATE_RND_RUNNING )
 	{
 		// This is determined via an attribute in SetStoredChargeLevel()
+		int iPreserveUberCharge = 0;
+		CALL_ATTRIB_HOOK_INT_ON_OTHER( pOwner, iPreserveUberCharge, preserve_ubercharge );
+		if ( iPreserveUberCharge > 0 )
+		{
+#ifdef _WIN32
```
ZzZombo, line 298: I don't think it's correct to use this conditional for the purpose you're trying to use it for. Shouldn't you then rather check for MSVC?

FlaminSarge: Might also be worth dividing by 99 in all cases (or something like 99.9 or 99.5) instead of making it conditional.

PR #1485, opened 2025-08-09
[TF] Fixed a bug where Gas Passer gas can cover players with gas in spawn rooms.

Bitl:
# Description
Ever since Jungle Inferno released in 2017, the Gas Passer had a bug where gas could give players the gas condition even in spawn rooms. This PR aims to fix it. The gas effect will still pass through spawn rooms, but will not collide with players in the spawn room.

Should be merged with https://github.com/ValveSoftware/source-sdk-2013/pull/1334

### DIFF
--- a/src/game/shared/tf/tf_weapon_jar_gas.cpp
+++ b/src/game/shared/tf/tf_weapon_jar_gas.cpp
@@ -29,6 +29,7 @@
 #include "tf_flame.h"
 #include "dt_utlvector_send.h"
 #include "collisionutils.h"
+#include "func_respawnroom.h"
 #endif
 
 
@@ -363,6 +364,12 @@ bool CTFGasManager::ShouldCollide( CBaseEntity *pEnt ) const
 		return false;
 
 	if ( pEnt->GetTeamNumber() == GetTeamNumber() )
+		return false;
+	
+	if ( PointsCrossRespawnRoomVisualizer( GetInitialPosition(), pEnt->GetAbsOrigin() ) )
+		return false;
+
+	if ( PointInRespawnRoom( pEnt, pEnt->GetAbsOrigin() ) )
 		return false;
 
 	if ( TFGameRules() && TFGameRules()->IsTruceActive() )


jh34ghu43gu:
If this is a bug then mad milk and jarate would need to be unable to coat enemies in spawn too?

Might as well disable all damage and give the condition that robots in mvm get while they're in spawn to everyone.

Bitl:
Hi all

There has been a lot of confusion and frustration regarding this fix, so I wanted to explain it here.

What it was meant to do was prevent the gas effect itself from spreading into the spawn room and covering people with gas. It still covers people with gas when hit near it or directly with it and doesn't fix the gas passer throwing glitch, which has been fixed in other PRs.

It's extremely subtle and has some side effects, and it's clear that the community has been confused by this change, so I plan on making a pr that reverts it (unless Eric does this)

Bitl:
I posted a comment here earlier and i feel like that one didn't explain the purpose enough, so here's a better one:

This PR is meant to fix gas passer gas particle entities passing though respawn rooms. Respawn rooms in TF2 are considered as func_respawnroom, and don't include the respawn door entities or the "no access" part of the spawn door. This only includes properly set up spawn rooms. This means maps that have spawn rooms further away from the door itself or don't have this entity at all, will still cover people with gas, which is why this still happens on badwater or other maps when throwing the gas passer near the spawn door. This PR works in situations where the respawn room entity is close to a respawn door.

This pull request doesn't fix the gas passer being thrown through spawn doors, which has been fixed by other PRs and for some reason wasn't merged alongside this one.

The gas passer functions far differently to the mad milk and the jarate, as it creates an entity that spreads gas in an area rather than gives a player a condition on impact. Its a different weapon in how it functions, and as such must be handled differently from the other throwable jars.

Bitl:
This PR should also be implemented in order to properly fix the bigger issues

https://github.com/ValveSoftware/source-sdk-2013/pull/1334

Bitl:
I'll be working on a revert PR later today or tomorrow

jh34ghu43gu:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1485#issuecomment-3420084706

This makes more sense than the title and original comment, thank you for the explanation.

PR #1486, opened 2025-08-10
[TF] Fix and rework Pyro bot airblast

Bitl:
# Description
This PR overhauls Pyro bot airblast functionality to reflect more projectiles. The previous implementation only functioned with a few projectiles and ignores other projectiles such as:

- The Flying Guillotine cleavers
- Huntsman Arrows
- Pipebombs/Cannonbals
- Sandman balls
- Crusader's Crossbow healing bolts
- Flares
- Stickybombs
- and others

This implementation allows the Pyro bot to recognize each of the above as projectiles and deflect them. In easy and normal difficulties however, the Pyro bot only recognizes Cow Mangler Energy Balls and Rockets as deflectable projectiles.

In addition, this makes bot airblasts more responsive as the IsLineOfSightClear() seems to return false if there's a projectile in front of the Pyro. This also removes tf_bot_pyro_deflect_tolerance, as the dot product is always 0 when there's a projectile in front of the Pyro.

### DIFF
--- a/src/game/server/tf/bot/tf_bot.cpp
+++ b/src/game/server/tf/bot/tf_bot.cpp
@@ -42,7 +42,6 @@ ConVar tf_bot_force_class( "tf_bot_force_class", "", FCVAR_GAMEDLL, "If set to a
 ConVar tf_bot_notice_gunfire_range( "tf_bot_notice_gunfire_range", "3000", FCVAR_GAMEDLL );
 ConVar tf_bot_notice_quiet_gunfire_range( "tf_bot_notice_quiet_gunfire_range", "500", FCVAR_GAMEDLL );
 ConVar tf_bot_sniper_personal_space_range( "tf_bot_sniper_personal_space_range", "1000", FCVAR_CHEAT, "Enemies beyond this range don't worry the Sniper" );
-ConVar tf_bot_pyro_deflect_tolerance( "tf_bot_pyro_deflect_tolerance", "0.5", FCVAR_CHEAT );
 ConVar tf_bot_keep_class_after_death( "tf_bot_keep_class_after_death", "0", FCVAR_GAMEDLL );
 ConVar tf_bot_prefix_name_with_difficulty( "tf_bot_prefix_name_with_difficulty", "0", FCVAR_GAMEDLL, "Append the skill level of the bot to the bot's name" );
 ConVar tf_bot_near_point_travel_distance( "tf_bot_near_point_travel_distance", "750", FCVAR_CHEAT, "If within this travel distance to the current point, bot is 'near' it" );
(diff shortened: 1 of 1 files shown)


Bitl:
Forgot to upload videos of this in action

https://github.com/user-attachments/assets/3217bc39-159a-4b72-aa24-bc97182c56e3

https://github.com/user-attachments/assets/4e2ec050-fe2b-47ec-9910-bef633acdd16

Bitl:
Hi, the recent commit adds a timer to the Pyro's airblast, meaning that the bot will wait half-a second to a second (or a quater of a second to a half of a second on Expert) before he fires another blast. Unfortunately, this isn't fully compatible with #1461.

To fix this, change

```
else if ( myWeapon->IsWeapon( TF_WEAPON_HANDGUN_SCOUT_PRIMARY ) )
{
	CTFPistol_ScoutPrimary *pPistolPrimary = assert_cast<CTFPistol_ScoutPrimary*>( myWeapon );
	// watch for enemy projectiles heading our way
	if ( pPistolPrimary->CanUsePush() && me->ShouldFireCompressionBlast() )
	{
		// bounce missiles with compression blast
		me->PressAltFireButton();
	}
      
	return;
}
```
	
into

```
else if (myWeapon->IsWeapon(TF_WEAPON_HANDGUN_SCOUT_PRIMARY))
{
	CTFPistol_ScoutPrimary* pPistolPrimary = assert_cast<CTFPistol_ScoutPrimary*>(myWeapon);
	// watch for enemy projectiles heading our way
	if (pPistolPrimary->CanUsePush() && me->ShouldFireCompressionBlast())
	{
		// bounce missiles with compression blast
		me->PressAltFireButton();

		if ( !me->m_CompressionBlastTimer.HasStarted() )
		{
			if ( me->IsDifficulty( CTFBot::EXPERT ) )
			{
				me->m_CompressionBlastTimer.Start( RandomFloat( 0.25f, 0.5f ) );
			}
			else
			{
				me->m_CompressionBlastTimer.Start( RandomFloat( 0.5f, 1.0f ) );
			}
		}
	}

	return;
}
```

Bitl:
Made it so Expert bots will have reduced times between airblasts.

ZzZombo:
I think the code is overly confusing and, to be frank, poorly organized. First, the variable names. You have a `bool` variable to check _if we have a rocket or a grenade_. Then you do the difficulty check based on the value of this variable, when **we have not yet checked if the object is in the LOS or not**, but the name suggests that we did, when we only determined the type of the projectile, if any. Also please stick to "deflect(ion)" if the existing code uses it.

And why even bother with all this if the projectile can't be deflected in the first place? Leave the original check as is, the earlier we bail the better. And even better, then we do not need to do this check in two places, simplifying the code a lot. It is overly verbose as is. I'd hate to have such code in our codebase, for example. Condense all the checks into a single `if` where possible and doesn't hurt readability, like https://github.com/ValveSoftware/source-sdk-2013/pull/1486/files#diff-f1cb3e5a7257263fab99711431ea3c01b4506abc9bd5c689bbdd00c66be49a70R4187-R4190, although restoring the deflection check would made this particular occurrence moot. Also does the SDK really not have an `isInstanceOf` macro?

And lastly, did you check that this code won't erroneously deflect projectiles behind solid geometry and objects?

PR #1487, opened 2025-08-10
[TF] Implement a way for users to toggle paint on items

Bitl:
# Description
As of recent, community made cosmetics implemented in TF2 have had various issues where the majority of the cosmetic can be recolored, causing team recognition issues (and potentially accessibility issues). Most of the time, the problem can be fixed by the community member creating the item, but sometimes cosmetics can be unchanged for a long period of time without changes.

This PR adds the client-side ConVar "econ_allow_paint_tint", which allows users to toggle whether or not they want to see paint on items. Having this turned off resets the item to the normal item color. This solution doesn't fix the team recognition issue, but allows users to have a temporary workaround until the effected items are fixed.

### DIFF
--- a/src/game/shared/econ/econ_wearable.cpp
+++ b/src/game/shared/econ/econ_wearable.cpp
@@ -459,6 +459,8 @@ RenderGroup_t CEconWearable::GetRenderGroup()
 	return BaseClass::GetRenderGroup();
 }
 
+ConVar econ_allow_painted_items( "econ_allow_painted_items", "1", FCVAR_ARCHIVE );
+
 //-----------------------------------------------------------------------------
 // Purpose: Wearable tint colors
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Bitl:
Demonstration:
https://github.com/user-attachments/assets/0c76250c-a788-42a7-8941-c995e702778d

Bitl:
Demonstration:

https://github.com/user-attachments/assets/9f94bad8-0a88-4e57-8241-0cb63c028174

JoriKos:
It's a neat command but there is only really one cosmetic item that's problematic (and it still is pretty obvious which team it is), changing that cosmetic is far better than completely disabling all paints

Bitl:
> It's a neat command but there is only really one cosmetic item that's problematic (and it still is pretty obvious which team it is), changing that cosmetic is far better than completely disabling all paints

It's true that there are only a few cosmetics that are problematic. However, some players might still see certain painted loadouts as being a problem in a couple ways:

- When it comes to art style consistency (as in cosmetics that don't seem to fit in with whatever team color, such as fully green cosmetics on the BLU team and orange or pink cosmetics on the RED team).
- Loadouts with non team-colored paints (such as lime green scouts on RED or pink pyros on BLU)

While most cosmetics tend to fall under the official TF2 color guide (as seen here https://www.teamfortress.com/contribute/images/tips_palette.gif) when unpainted, applying paint changes the original appearance of the item and indirectly changes how that item looks to an enemy player, which could effect team recognition. 

Paints and customization isn't a bad thing, that's why I wanted this to be an option and why paint is turned on by default. For people who have an issue with team recognition with painted items, giving them a way to disable it is a good thing for accessibility.

PR #1490, opened 2025-08-12
[TF2] Fix Thermal Thruster passives not being removed if it's unequipped during use (or player is a Medic)

bakugo:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/4731

The `TF_COND_ROCKETPACK` cond (stomp, lower fall damage, etc) from the Thermal Thruster is supposed to be removed as soon as the player touches the ground, but the code responsible does not do this if the player does not have a Thermal Thruster equipped.

As a result, players can stop the cond from being removed and keep it until they die by unequipping the Thermal Thruster after using it but before touching the ground (such as by flying into a resupply cabinet after changing their loadout). Additionally, Medics with the Quick Fix will always keep the cond until they die if they heal a Pyro as he uses the Thermal Thruster, since they inherit the cond from the Pyro but don't have the weapon themselves, so the cond is never removed from them.

Fixed by removing `TF_COND_ROCKETPACK` if a player touches the ground without the Thermal Thruster equipped.

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -3135,6 +3135,11 @@ void CTFPlayerShared::ConditionThink( void )
 #endif
 				}
 			}
+			else
+			{
+				// RocketPack not found, may have been unequipped, or we may be a Medic who inherited the cond
+				RemoveCond( TF_COND_ROCKETPACK );
+			}
 		}
 	}
 


AzureWoof:
Definitely should be fixed, but at the same time I'd like to see the stomp return to the Detonator as a balance change, without the decreased fall damage of course. This bug being in the game for so long has transformed the Detonator into being a much more interesting weapon to play with, and I'd find it relatively harmless.

Matiaz00Baby:
Made an account just to say I'll never forgive you for this.

AbelStekel:
this should be reverted
this 'fix' just removes a funny quirk from the game that was not relevant in 99% of cases and sets a rly bad precedent of just removing any bug without looking at if there is actually relevant gains from it (kinda similar to that recent [now-reverted] gas passer fix, sure ur fixing a 'bug' [players taking dmg in spawn], but was this bug actually relevant?)
this is kind of akin to if they were to remove speedshots 8 years (yes, it has been 8 years since the jetpack was added) after the original launch of the game , it wouldnt make any sense. ppl are used to it working like this and removing this functionality just removes potential amusement from the game

AbelStekel:
i am going to quietly retract what i said after finding members of pyro team (which distorto should know a lot about) harrasing the guy who patched out the stomp bug
https://steamcommunity.com/id/bakugo
in case ur curious
this is really awful behaviour

wgetJane:
i personally dont think this bug should've been fixed since it seems like a funny mechanic even if it's unintended

however, i dont see how sending people to harass bakugo is exactly going to get you what you want, it's just embarrassing and childish

instead, if you want to do something productive, scroll up and click this to make a new pull request that reverts this change:

image

in that pr, explain why you want the change to be reverted, and send people there to comment why they agree and express support for the revert

that would have an actual chance of getting what you want vs throwing a collective temper tantrum

bakugo:
Something that would make everyone happy would be to just add the `boots_falling_stomp` attribute to the Detonator and call it a day, but I guess that might count as an actual balance change so it might be off the table...

old702:
It seems the community’s reaction to the fix has been quite mixed. It might be worth considering reverting it, or at least re-evaluating the idea — since there are valid arguments for why this “bug” might actually have a positive impact on the game and relatively harmless overall.

The intention is understandable, but it seems to target something harmless and even beneficial, while the actual goal was to fix a completely different bug (the Medic stomp).

The Thermal Thruster bug has an additional hidden features:

-  It increases knockback from all sources of damage (which can make mobility and positioning more engaging).

-  It also amplifies effects such as direct-hit minicrits, creating a natural risk–reward dynamic.

- Additionally, it causes a persistent looping sound that follows the player and reveals their position, which can be considered a significant drawback too.

Despite being a bug, It actually feels quite balanced on its own and comes with several drawbacks that effectively act as built-in penalties for using it

Adding stomp-related attributes to the Detonator, however, would be a major balance change. @bakugo Perhaps increased knockback could serve as a more elegant alternative solution for the Detonator, but that's just my speculation without deeply thinking into it.

And if the goal is to completely remove the Thermal Thruster bug, the remaining infinite sound loop issue also needs to be addressed separately.

megascatterbomb:
We should not be leaving bugs in the game based purely on the enjoyment gained from exploiting said bugs. You can argue that the bug is minor and has barely any impact, but the truth remains that this is an exploit which allows players to gain an unfair advantage over others.

If you want the behaviour to remain, it should be implemented properly in a separate PR (e.g. as a passive buff for the Thermal Thruster) where we can debate the specifics of its implementation. However, that would count as a balance change and it's widely understood that balance changes are outside the scope of this repo's intended purpose.

And before people ask "but what about bhop-gardening?" or any other bugs left in the game that affect gameplay; obviously these should be addressed on a case-by-case basis (and I would make an exception for bhop gardening), but the mere presence of enjoyment on the part of the exploiter is not enough for me to accept it remaining.

I personally believe this fix should be reinstated, and the like-dislike ratio on the original PR proves that there is interest in having this fixed.

Matiaz00Baby:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1490#issuecomment-3470932504

I somewhat understand your point but if were talking based on the generals opinion of the change more people wanted it back then people wanted it gone. Also saying it provides an unfair advantage is a bit of a reach I believe, theres a fair tradeoff which is almost every scenario. Truth is this doesn't affect anyone but the people who want to use it because up until the post had been made there was no complaints about it nor did it give an unfair advantage. It just doesn't make alot of sense considering that this game is barley being worked on and  most other bugs like b-hop gardening will never be fixed because it's integrated itself so much into the game that it's practically a feature. not all bugs are bad.

Luxzhv:
By keeping this exploit, we keep a play style players want for a totally different weapon, while having another bug tied to another class aswell.

Fix the medic stomp, and let's be done with this.

SolarLightTF2:
as the world's biggest pyro hater let it be known that nobody was seriously complaining about this bug

once upon a time valve would look at cool bugs and actually implement them into the game (charge turn bugs became the tide turner)

let detonator stomp or leave the bug in the game. the bug requires so much setup that it'll practically never occur by accident

Luxzhv:
Requesting this PR to be locked and we all move to [#7654](https://github.com/ValveSoftware/Source-1-Games/issues/7654#issue-3570594766) to have a more constructive discussion about this.

PR #1491, opened 2025-08-13
[All] add cl_no_texture_stream to toggle texture streaming

Bitl:
# Description
Closes https://github.com/ValveSoftware/Source-1-Games/issues/4884*

Since the Tough Break update in 2015, Source 2013 got texture streaming, which aims to improve performance when loading certain textures in. However, this causes some issues:

- Decreased texture quality when first loading a model
- Slower loading times

-no_texture_stream is the primary option to disable this feature, however it isn't in ConVar form. I added cl_no_texture_stream to add this option automatically on bootup before the client dll loads other modules, enabling users to toggle this setting without messing around with launch options.

*However, this isn't a replacement for -no_texture_stream, as it requires a restart when enabling it. Adding a ConVar that toggles texture steaming dynamically is the best option, but requires access to the materialsystem module code to do so. This implementation is the best possible way to implement this kind of option with the SDK codebase.

### DIFF
--- a/src/game/client/cdll_client_int.cpp
+++ b/src/game/client/cdll_client_int.cpp
@@ -331,12 +331,14 @@ class IMoveHelper;
 
 void DispatchHudText( const char *pszName );
 
-static ConVar s_CV_ShowParticleCounts("showparticlecounts", "0", 0, "Display number of particles drawn per frame");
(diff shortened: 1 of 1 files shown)


Bitl:
Closing this as this method seems ineffective through testing

PR #1499, opened 2025-08-18
[Audio] Fix ordering of hitsound calls in CBaseHLBludgeonWeapon::Swing() and play melee hitsound in CBaseHLBludgeonWeapon::Hit()

pivotman319-owo:
# Description
This PR properly fixes melee weapon hit sounds in _Half-Life 2_ and derivative mods by adding a missing `WeaponSound( MELEE_HIT )` call at the end of the hit target entity check in `CBaseHLBludgeonWeapon::Hit()`, and changes the ordering of the swing `WeaponSound` call in `CBaseHLBludgeonWeapon::Swing()` to be placed within the miss conditional block.

- Original, unfixed codepath in live Half-Life 2 and Team Fortress 2 SDK (for reference):

https://github.com/user-attachments/assets/7582d0f1-f39e-4510-81b1-54f28d34da3a

- Fix in PR:

https://github.com/user-attachments/assets/1d1a54b8-f088-466c-995f-0b6a84eed29f

## Note
This exact hitsound fix is only _partially_ present in live HL2 exclusively for the player stunstick ent (`weapon_stunstick_player`). The regular stunstick (`weapon_stunstick`) and crowbar/mattpipe (`weapon_crowbar`) do not have this change, because the fix that's included is in the wrong place!

https://github.com/user-attachments/assets/951f3663-6d4b-46c2-95e2-e05010dbab0b

### DIFF
--- a/src/game/server/basebludgeonweapon.cpp
+++ b/src/game/server/basebludgeonweapon.cpp
@@ -178,10 +178,6 @@ void CBaseHLBludgeonWeapon::Hit( trace_t &traceHit, Activity nHitActivity, bool 
 		// Play hit sound
 		WeaponSound( MELEE_HIT );
 	}
-	else {
-		// Play hit world sound
-		WeaponSound( MELEE_HIT_WORLD );
-	}
 
 	// Apply an impact effect
 	ImpactEffect( traceHit );


dimhotepus:
Just two notes.

1. Crowbar and stunstick both have melee hit / miss sounds in animation events callbacks

See https://github.com/ValveSoftware/source-sdk-2013/blob/a36ead80b3ede9f269314c08edd3ecc23de4b160/src/game/shared/hl2mp/weapon_crowbar.cpp#L123, https://github.com/ValveSoftware/source-sdk-2013/blob/a36ead80b3ede9f269314c08edd3ecc23de4b160/src/game/shared/hl2mp/weapon_crowbar.cpp#L132 and https://github.com/ValveSoftware/source-sdk-2013/blob/a36ead80b3ede9f269314c08edd3ecc23de4b160/src/game/shared/hl2mp/weapon_stunstick.cpp#L358, https://github.com/ValveSoftware/source-sdk-2013/blob/a36ead80b3ede9f269314c08edd3ecc23de4b160/src/game/shared/hl2mp/weapon_stunstick.cpp#L406

so melee hit sound you have added is redudant.

2. Current implementation still plays swing sound when water is hit. Need to check nothing hit first and only then play swing sound, like:

```
// -------------------------
//	Miss
// -------------------------
if ( traceHit.fraction == 1.0f )
{
	nHitActivity = bIsSecondary ? ACT_VM_MISSCENTER2 : ACT_VM_MISSCENTER;

	// We want to test the first swing again
	Vector testEnd = swingStart + forward * GetRange();
	
	// See if we happened to hit water
	if ( !ImpactWater( swingStart, testEnd ) )
	{
		// Play swing sound only if nothing impacted.
		WeaponSound( SINGLE );
	}
}
```

Otherwise, great MR, thank you!

Bitl:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1499#issuecomment-3445903322

 WeaponSound( MELEE_HIT ); seems to only be needed in MP, maybe the one added in this PR should use #ifndef HL2MP

PR #1503, opened 2025-08-23
[TF2] VScript `-noscripting` crash 

satanskitty:
Currently `-noscripting` launch option crashes due to some pointers being `NULL`. I added a check to every single instance where these variables are accessed, probably overkill but tracking down each and every individual crash would be too much of a pain


Bitl:
Wasn't this fixed in #1347?

PR #1506, opened 2025-08-24
[TF2] Make Killfeed use the Assisters team color if applicable instead of always using the Killers team color

gidi30:
# Description
Currently the Killfeed will always color both the Killer and Assister in the team color of the Killer even if the Assister is on a different team.
This PR aims to change that by storing the Assisters name and team seperately from the Killer and drawing the + in a neutral color if the teams are different.

# Before:

https://github.com/user-attachments/assets/bedddd64-83d0-4c7c-a5df-e91cbcb3dfc1

# After:

https://github.com/user-attachments/assets/90a8c91a-0be5-4697-bedb-98c81e8f6d6d

### DIFF
--- a/src/game/client/tf/hud_basedeathnotice.cpp
+++ b/src/game/client/tf/hud_basedeathnotice.cpp
@@ -151,17 +151,22 @@ void CHudBaseDeathNotice::Paint()
 
 		wchar_t victim[256]=L"";
 		wchar_t killer[256]=L"";
+		wchar_t	assister[256]=L"";
 
+		const wchar_t *plus = L"+";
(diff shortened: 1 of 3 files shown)


Professor-Heavy:
[Oof. Looks like I was a few months too late to this one.](https://github.com/Professor-Heavy/source-sdk-2013/tree/assister-killfeed-info) Remember to always check for existing pull requests and issues, kids.

To be honest, I do personally wonder if we can take more from the assister. Displaying their Mannpower powerup and killstreak (Only used/relevant in `player_killed` if the assister is a Medic) is a possibility, basically putting them on the same level of detail as the killer themselves.

PR #1507, opened 2025-08-24
[TF] Fix melee hit registration

Bitl:
# Description
One commonly reported TF2 issue is one related to melee hit registration. If you swing a bat and it hit a teammate, intending to hit the enemy behind the teammate, it will hit the teammate instead of the enemy, which makes melee weapons feel unresponsive. There was a fix for this issue introduced for MvM bots, but not for players. This code changes this line of code to ONLY hit teammates if friendly fire (mp_friendlyfire) is enabled.

### DIFF
--- a/src/game/shared/tf/tf_weaponbase_melee.cpp
+++ b/src/game/shared/tf/tf_weaponbase_melee.cpp
@@ -424,9 +424,8 @@ bool CTFWeaponBaseMelee::DoSwingTraceInternal( trace_t &trace, bool bCleave, CUt
 	Vector vecSwingStart = pPlayer->Weapon_ShootPosition();
 	Vector vecSwingEnd = vecSwingStart + vecForward * fSwingRange;
 
-	// In MvM, melee hits from the robot team wont hit teammates to ensure mobs of melee bots don't 
-	// swarm so tightly they hit each other and no-one else
-	bool bDontHitTeammates = pPlayer->GetTeamNumber() == TF_TEAM_PVE_INVADERS && TFGameRules()->IsMannVsMachineMode();
+	// only hit teammates if friendly fire is on.
+	bool bDontHitTeammates = ( !friendlyfire.GetBool() );
 	CTraceFilterIgnoreTeammates ignoreTeammatesFilter( pPlayer, COLLISION_GROUP_NONE, pPlayer->GetTeamNumber() );
 
 	if ( bCleave )
--- a/src/game/shared/tf/tf_weaponbase_melee.cpp
+++ b/src/game/shared/tf/tf_weaponbase_melee.cpp
@@ -425,7 +425,7 @@ bool CTFWeaponBaseMelee::DoSwingTraceInternal( trace_t &trace, bool bCleave, CUt
 	Vector vecSwingEnd = vecSwingStart + vecForward * fSwingRange;
 
 	// only hit teammates if friendly fire is on.
-	bool bDontHitTeammates = ( !friendlyfire.GetBool() );
+	bool bDontHitTeammates = !friendlyfire.GetBool();
 	CTraceFilterIgnoreTeammates ignoreTeammatesFilter( pPlayer, COLLISION_GROUP_NONE, pPlayer->GetTeamNumber() );
 
 	if ( bCleave )


Bitl:
note that this is a basic implementation, and doesn't consider other teammate related on hit events, so this is marked as a draft until this is fixed.

Bitl:
This PR now functions as expected. Here's a demo video about it!

https://www.youtube.com/watch?v=J-2n-fbAOq8
[revised melee hit fix](https://www.youtube.com/watch?v=J-2n-fbAOq8)

Bitl:
Implemented a HL2MP version

Gavitro:
`// In MvM, melee hits from the robot team wont hit teammates to ensure mobs of melee bots don't swarm so tightly they hit each other and no-one else`
`bool bDontHitTeammates = pPlayer->GetTeamNumber() == TF_TEAM_PVE_INVADERS && TFGameRules()->IsMannVsMachineMode();`
`CTraceFilterIgnoreTeammates ignoreTeammatesFilter( pPlayer, COLLISION_GROUP_NONE, pPlayer->GetTeamNumber() );`

I would not remove or replace this code, because some plugins/vscripts abuse it in order to bypass teammates.

Bitl:
This code is no longer needed with this PR, however I'll implement a Vscript function that re-enables something similar to this

Bitl:
that code is no longer needed since the majority of melee weapons don't have the listed exceptions (since very few melee weapons have on hit attributes for teammates) However, I'll possibly make it so that you could switch back to the original behavior using vscript

Bitl:
There are now 3 ways to set the old melee behavior:

1. The tf_melee_old_trace_behavior ConVar (Global)
2. The SetOldMeleeTrace VScript function in CTFPlayer, intended to be used on the owner of the melee weapon (Player)
3. The set_old_melee_trace_behavior static attribute (Player/Weapon)

SourceMod plugins could also hook into the CTFPlayerShared::SetOldMeleeTrace function as well.

JoriKos:
This is a balance decision, not a bug. Also being able to not hit teammates anymore would be very very badly received by the community as well as make Spychecking even easier (making Spy much weaker when he already is very weak)

Bitl:
this absolutely isn't a balance decision. it's a bug. even if it isn't one, it's genuinely annoying to hit your teammates during combat. for example: if you're playing as demoknight and a teammate goes in front of you while you're charging into an enemy, the teammate will tank the sword damage, which frustrates the demoknight player. in contrast: many hitscan and projectile weapons pass through teammates, and it only makes sense that melee weapons should also be treated the same way. spychecking can already be done by shooting your teammates or using a flamethrower. this change only lines up with how teammates should react when someone's shooting projectiles.

this also wouldn't be negatively received by the community, as a bunch of larger figures have expressed the same problems. this is a reason i added a toggle convar to it, for server owners that prefer the older melee hit registration.

this PR also does not prevent every instance of teammate hits, as this PR makes it so the Disciplinary Action and PASS time mechanics will function normally, as well as the unused add_give_health_to_teammate_on_hit attribute.

JoriKos:
> This absolutely isn't a balance decision

I don't see why it wouldn't be. 

> it's genuinely annoying to hit your teammates during combat.

It's annoying but necessary, the same applies to autobalance and projectiles being consumed by teammates. It would be stupid to suddenly be hit by a projectile that you physically couldn't see coming, similar to being hit by melee you can't even see. It's just something necessary, but annoying.

> spychecking can already be done by shooting your teammates or using a flamethrower

Exactly, so why make it even easier to check? I've had times with people have hit me once and then because I didn't react they didn't think I was suspicious. Now it would be really obvious because you hear the melee sound that otherwise doesn't play.

> a bunch of larger figures have expressed the same problems

Doesn't mean they're right though, they just say it's annoying but removing hits on teammates would 100% be a very unpopular decision that gets removed within a week. Some community figures also call for Quickplay back but they're objectively wrong when they say it's an improvement. Just because someone has a large subscriber/view count doesn't mean they are more correct.

> this PR also does not prevent every instance of teammate hits

Those aren't the hits I'm talking about. Not having teammate hits actually removes a lot of silly interactions.

Bitl:
With some re-evaluation, yeah this is a balance change. I'll mark this PR as a modding resource for those who want to implement this functionality in their mods

Bitl:
I decided to revert the modding resource tag and decided to stick with my original opinion and it won't be changed any longer

hotpockette:
Considering the problem this is trying to solve, your solution is far too drastic.
It would make more sense to only ignore friendly players if the attack is in range of a valid enemy target.

I imagine this would involve two traces:
One trace for friendlies, one trace for enemies. If both are detected in one swing, ignore friendly players for that attack.

Teammates would still be allowed to hit eachother for fun, and I can't imagine it would change anything in regards to spychecking. It would only apply in situations where you truly would have hit your enemy had you not been blocked by a teammate.

PR #1508, opened 2025-08-25
Add `C_BaseEntity::CanGlow` to opt-out of glowing for Sentry Gun shield

drunderscore:
This fix was requested from me by a community member reporting bugs for [CastingEssentials Next](https://github.com/drunderscore/CastingEssentialsNext).

Before (caused by 57a8b64)
image
After
image

### DIFF
--- a/src/game/client/c_baseentity.h
+++ b/src/game/client/c_baseentity.h
@@ -1449,6 +1449,10 @@ public:
 
 	void TrackAngRotation( bool bTrack );
 
+#ifdef GLOWS_ENABLE
+	virtual bool CanGlow() const { return true; }
+#endif // GLOWS_ENABLE
+
 private:
 	friend void OnRenderStart();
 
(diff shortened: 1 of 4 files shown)


mastercoms:
Though an unintended side effect of a change, it isn't necessarily a bad one I feel. Is there a particular reason the original behavior should be maintained?

copperpixel:
> Though an unintended side effect of a change, it isn't necessarily a bad one I feel. Is there a particular reason the original behavior should be maintained?

Since the shield masks the sentry model's glow outline, the player loses information on what level the sentry is, which IMO is more important than knowing whether it is wrangled or not.

JoriKos:
Is there a chance there could be a toggle, or perhaps a way to display both Wrangler shield and Sentry Gun? I quite like the big circle, but I do see how not having the level visible is an issue.

PR #1509, opened 2025-08-25
[TF2] Fix reanimator fast revive exploit

HyperionCoding:
# Description

Fixes the following reanimator exploit:

> Constantly pressing the primary attack button ([default key](https://wiki.teamfortress.com/wiki/List_of_default_keys): MOUSE1) revives teammates considerably faster than the regular way. This happens because each time the reviver presses their primary attack button, a small percentage of health is recovered.

The fix is not in the reanimator code (`AddMarkerHealth()`) because it would break multiple Medics healing one marker and Vaccinator instant(ish) revive, and allow Medic changing between different markers to negate the fix.


Generalisk:
I wonder how many MvM players will complain about it if this gets merged? And how long it'll take before Valve reverts the fix?

wgetJane:
unless im misunderstanding how this works, wouldnt it be better to just move the m_flReviveMarkerNextHeal check to here:

```cpp
if ( gpGlobals->curtime >= m_flReviveMarkerNextHeal )
{
	pReviveMarker->AddMarkerHealth( flReviveRate );
}
m_flReviveMarkerNextHeal = gpGlobals->curtime + ROUND_TO_TICKS( 0.2f );
```

PR #1510, opened 2025-08-26
Added AppID mounting to non-engine mods

rawra89:
# Description
There doesn't appear to be a way to use AppID based mounting on non-engine Source Engine projects, with the original code giving an error. This is very annoying even though the ``gameinfo.txt`` files that this SDK comes with uses AppID mounting. I've noticed that the ``ENGINE_DLL`` define had code to retrieve an application install directory through ``SteamGames()``, but trying to use that code outside that context causes a linker error. Instead, I've implemented a way to get the AppID's install directory without having to use the Steam API. This method checks the Steam installation path, and uses ``libraryfolders.vdf`` and ``app_manifest_appid.acf`` to look for the proper AppID.


rawra89:
**Note:** There's also an issue with Steam where occasionally, if you first installed a game without launching it, it doesn't update the ``libraryfolders.vdf`` file. It seems to be a bit inconsistent, because I've had this issue with Portal when testing this code. But now, when I test it, it seems to update the entry immediately after installing it. You may have to launch the game once for Steam to add the queried app to ``libraryfolders.vdf`` so that this code can detect the game.

Zeldaboy14:
I should add on here that this issue doesn't exist in the old sdk at all. I recently had to revert this, as compiling HLMV from the 2006 EP1 source does not work with your own mod in terms of launching, thanks to this. Not sure why there was a whole weird check added into the latest sdk

rawra89:
I accidentally closed this pull request, I accidentally committed something else to this and will create another PR for this at some point.

PR #1514, opened 2025-08-27
[TF2] Crusader's Crossbow arrow healing is now affected by mult_health_fromhealers (used by Back Scratcher)

JoriKos:
Added an extra `CALL_ATTRIB_HOOK_FLOAT_ON_OTHER` for the specified attrib. I don't really know if this is the best approach, but there is pretty much no reason why you would have both attributes. One reduces healing from Medics and one reduces it from Medic + other healing sources, so if both effects are applied they would affect each other.

From testing, it reduces healing just fine.

Intended to fix https://github.com/ValveSoftware/Source-1-Games/issues/7521


JoriKos:
> I don't really know if this is the best approach, but there is pretty much no reason why you would have both attributes.

I found some similar instances of this being done in other places (for example, tf_weapon_rocketlauncher.cpp at line 682-684), so it should be finee.

PR #1515, opened 2025-08-27
Add "ParticleEffectStopAndDestroy" Client Effect and "DestroyImmediately" Input for info_particle_system

horiuchii:
# Description

This PR adds the `DestroyImmediately` Input for `info_particle_system`, this destroys the particles instantly instead of stopping the emission like with the `Stop` Input.

https://github.com/user-attachments/assets/beffc06b-4d25-4489-b02d-356d317f6f80

It also adds the client effect `ParticleEffectDestroyImmediately`, which is the same situation as before but compared to the `ParticleEffectStop` client effect.

### DIFF
--- a/src/game/client/c_particle_system.cpp
+++ b/src/game/client/c_particle_system.cpp
@@ -33,6 +33,7 @@ protected:
 	int			m_iEffectIndex;
 	bool		m_bActive;
 	bool		m_bOldActive;
+	bool		m_bDestroyImmediately;
 	float		m_flStartTime;	// Time at which the effect started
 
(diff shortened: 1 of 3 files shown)


rabscootle:
Would this fix resolves weapon unusual effects persisting in player view after death?

https://github.com/ValveSoftware/Source-1-Games/issues/3926

horiuchii:
> https://github.com/ValveSoftware/source-sdk-2013/pull/1515#issuecomment-3230058254

No, this PR only adds inputs. No changes have been made to other systems.

PR #1520, opened 2025-09-03
[TF2] Add ForceDifficultyBasedAim CTFBot flag to allow MvM missions to let Spies aim better

FlaminSarge:
# Description
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/7538

This adds a flag that popfiles can use to let spies use default aiming behavior instead of handicapped aiming behavior.


HyperionCoding:
I think a better way would be to make it depend on the bot skill and use the imprecise aim for all skill levels except expert. it doesn't need a whole new attribute.

FlaminSarge:
> I think a better way would be to make it depend on the bot skill and use the imprecise aim for all skill levels except expert. it doesn't need a whole new attribute.

Because bot difficulty controls more than just this aim setting, we have to maintain compatibility with the potential existing use case of someone intentionally using Expert Spy _with_ the handicapped aim behavior.

PR #1528, opened 2025-09-13
[TF2] Add A Non full charge sound Check to tf_weapon_sniperrifle.cpp

ChaseTownsend:
This PR fixes any sniper rifle that is Not the Machina playing an empty sound if "sniper_full_charge_damage_bonus" is not 1


Bitl:
i feel like restricting the sniper charge sound to only the machina would cause issues for modders who are trying to implement sniper rifles with this attribute.

a better change would be the following:

```	
		// Prevents Non-Machina rifles with "sniper_full_charge_damage_bonus" 
		// from playing "SPECIAL3", which usually is Undefined
		const CEconItemView* pItem = GetAttributeContainer()->GetItem();

		if ( flDamageBonus > 1.0f && ( pItem && pItem->GetStaticData()->GetWeaponReplacementSound( GetTeamNumber(), SPECIAL3 ) ) )
```

this fixes the issue, while also allowing modders to implement their own sound effects for their sniper rifles.

ChaseTownsend:
Yeah that looks much better than my version.
I didn't even know that you could get the sound replacement.

PR #1529, opened 2025-09-13
[ALL]? Add DevMsg's for returns in CBaseCombatWeapon::WeaponSound

ChaseTownsend:
Adds some DevMsg's to Warn Players/Developers about missing shoot sounds
an example of this is in TF2 when adding "sniper_full_charge_damage_bonus" to a non Machina Rifle. Code tried to play "Special3" But for other Rifles, they dont have this sound defined, leading to a null sound in CBaseCombatWeapon::WeaponSound.


ChaseTownsend:
Closed due to many prints from spawning.

PR #1530, opened 2025-09-14
[TF2] Allow null caller on team_control_point's SetOwner input

megascatterbomb:
`team_control_point` has a SetOwner input for manually changing a point's ownership to a specified team. This input will credit a player for capturing the point if that player is the caller. If the calling entity is not a player, no credit is given.

However, there is an additional guard and assert against null callers which unexpectedly prevents the point from changing ownership at all. Given that nothing happens with the caller unless that caller is a player, these checks don't appear to serve any purpose beyond what the IsPlayer check already achieves.

This PR removes the guard and assert, incorporating the null-check alongside the IsPlayer check.

An easy way to test this PR is to use the following command in console on a koth map. Do not use `ent_fire` as that sets you as the caller.

`script Entities.FindByClassname(null, "team_control_point").AcceptInput("SetOwner", "3", null, null)`

### DIFF
--- a/src/game/server/team_control_point.cpp
+++ b/src/game/server/team_control_point.cpp
@@ -336,11 +336,6 @@ void CTeamControlPoint::InputSetOwner( inputdata_t &input )
 
 	Assert( iCapTeam >= 0 && iCapTeam < GetNumberOfTeams() );
 
-	Assert( input.pCaller );
-
-	if ( !input.pCaller )
-		return;
-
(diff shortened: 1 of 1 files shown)


SirYodaJedi:
Shouldn't the Activator be used instead of the Caller? That way if the initiator of the I/O chain is a player, they still get credit (as the caller is never going to be a player outside of ent_fire, but the activator of an I/O chain can be).

PR #1532, opened 2025-09-15
[TF2] Fix team skins for tf_projectile_rocket

DudeTheNinja:
`tf_projectile_rocket` has no proper full support for team-based skingroups. When a rocket is fired, it will always be skin 0, the RED team skin. This would be considered fully-intended behaviour since no rocket model never uses team skins in the vanilla game (outside of clientside content mods and potentially custom maps changing the model with `SetModel`). However, this would only be the case were it not for the method `CTFProjectile_Rocket::Deflected`, which changes the team skin when the rocket is airblasted to the skingroup of the team that the deflector belongs to. 
As a result, the lack of team skin setting when the rocket is spawned is possibly erroneous.

This pull request fixes this discrepancy by adding the team skin support when the rocket is spawned.

### DIFF
--- a/src/game/server/tf/tf_projectile_rocket.cpp
+++ b/src/game/server/tf/tf_projectile_rocket.cpp
@@ -52,6 +52,8 @@ void CTFProjectile_Rocket::Spawn()
 {
 	SetModel( ROCKET_MODEL );
 	BaseClass::Spawn();
+	CTFPlayer* pTFOwner = ToTFPlayer(GetOwnerPlayer());
+	m_nSkin = ( pTFOwner -> GetTeamNumber() == TF_TEAM_BLUE ) ? 1 : 0;
 }
 
 //-----------------------------------------------------------------------------


DudeTheNinja:
This is my first time making a PR for _any_ project, so I apologise if anything's messy with this. Please let me know if my formatting or methodology is not up-to-speed and I'll happily make amendments if necessary.

bakugo:
This may cause a crash if the rocket isn't owned by a player. You don't even have to check for players specifically, all entities have `GetTeamNumber()` so you can just call `GetOwnerEntity()`, check if the result is null, and then call `GetTeamNumber()` on that.

DudeTheNinja:
> This may cause a crash if the rocket isn't owned by a player. You don't even have to check for players specifically, all entities have `GetTeamNumber()` so you can just call `GetOwnerEntity()`, check if the result is null, and then call `GetTeamNumber()` on that.

Thank you for pointing that out, I've updated that bit from `GetOwnerPlayer()` to `GetOwnerEntity()` . 

(I presume the commit automatically gets added to the PR?)

bakugo:
Your code no longer compiles at all, you didn't change the type of the variable. You should at least try compiling before committing.

DudeTheNinja:
Sorry. I'm doing this late at night and only used the in-browser editor to do that quick fix and glossed over that. Thank you for being patient with me, and I apologise for wasting your time.

PR #1534, opened 2025-09-16
[TF2] fix: cleanup collect on rest logic

mastercoms:
Use the provided utility functions for trigger_hurt and respawn room to determine the relevance of the trigger, rather than doing a broader, incorrect bounds check. This takes into consideration all the relevant conditions for each respective trigger to be valid, and checks if the currency's point is truly within the trigger.

This fixes an error where some maps with more complex geometry for their triggers will cause a currency pickup, which was exacerbated by #618, which added respawn room to the logic.

### DIFF
--- a/src/game/server/tf/entity_currencypack.cpp
+++ b/src/game/server/tf/entity_currencypack.cpp
@@ -189,41 +189,25 @@ void CCurrencyPack::ComeToRest( void )
 		TFGameRules()->DistributeCurrencyAmount( m_nAmount );
 		m_bTouched = true;
 		UTIL_Remove( this );
-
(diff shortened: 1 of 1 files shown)


HyperionCoding:
#1307 duplicate

mastercoms:
> #1307 duplicate

That PR differs with this one, this PR uses the shared utility functions and does not do other changes.

mtxfellen:
~~Outside of the other miscellaneous differences, I wonder if using the utility functions here is ideal as it precludes checking whether the triggers are actually enabled before auto-collecting the contained packs.~~ This isn't true!

PR #1542, opened 2025-09-23
[ALL] Add Convar to Remove print from having cheats on for Achievements

ChaseTownsend:
Adds a Convar that disables the `"Achievements disabled: cheats turned on in this app session."` message from showing up. useful for people messing with the SDK or server owners who have to turn cheats on and have their consoles being spammed by this message.


JoriKos:
Also useful when you do anything BASE Jumper-related on an `sv_cheats 1` server considering every check for the achievement gives this output.

PR #1543, opened 2025-09-24
Fix mounting of custom content in gameinfo

ficool2:
There is two issues with how the `custom` folder is mounted in the SDK currently:
1. The custom folder is using the hardcoded mod's directory name. Not only is this less maintainable, external tools cannot resolve the path correctly. This was an issue observed in Hammer.
2. TF2 uses `custom_mod` pathid to only mount custom HUDs are that are installed in a folder with this path id. Right now custom HUDs do not work as there is no `custom_mod` pathid unlike live TF

The PR addresses both of these issues for TF2 and HL2MP

### DIFF
--- a/game/mod_hl2mp/gameinfo.txt
+++ b/game/mod_hl2mp/gameinfo.txt
@@ -42,7 +42,7 @@
 			// folder.
 			//
 			// Note that this folder is scanned only when the game is booted.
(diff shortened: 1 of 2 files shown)


mastercoms:
Missing `vgui` pathid as found in #1187

PR #1551, opened 2025-10-11
Stop a crash when prop_ragdoll has NULL modelname

BreakinBenny:
This already exists for prop_physics and even prop_dynamic, but not prop_ragdoll.

# Description
**CBaseProp**s already default to "error.mdl" if we didn't provide a model name, while CPhysicsProp outright halts if you leave it undefined. **CRagdollProp** however, is an exception if you just `ent_create prop_ragdoll`.

### DIFF
--- a/src/game/server/baseentity.cpp
+++ b/src/game/server/baseentity.cpp
@@ -8510,7 +8510,7 @@ void CBaseEntity::SetCollisionBoundsFromModel()
 
 
 //------------------------------------------------------------------------------
-// Purpose: Create an NPC of the given type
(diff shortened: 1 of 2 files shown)


BreakinBenny:
I should mention that I haven't been able to compile "Server (TF)" to check whether this works, even though I have the required VS2022, Python and Windows 10 stuff. It gives me the C2382 error.

BreakinBenny:
Closed because I need to think about how to do this one. It's a good idea, but there's more to it than just spawning CRagdollProp like with CBaseProp.

PR #1553, opened 2025-10-14
[TF] Restore cut MvM Upgrade Station music.

Bitl:
# Description
This PR restores the cut MvM Upgrade Station music with a clientside implementation. This implementation is enabled by default.

NOTE: This may conflict with custom maps that play the song through external entities. I am trying to figure out a solution for this, so at the moment this will be marked as a draft.

The tf_music_upgrade_machine.wav file must be edited to support looping. The tf_music_upgrade_machine.wav provided with this PR supports looping.

https://github.com/user-attachments/assets/0d3ad5f2-8649-4256-9e60-3d909b60d821

### DIFF
--- a/src/game/client/tf/player_vs_environment/c_tf_upgrades.cpp
+++ b/src/game/client/tf/player_vs_environment/c_tf_upgrades.cpp
@@ -42,7 +42,7 @@
 extern CAchievementMgr g_AchievementMgrTF;
 
 ConVar tf_mvm_tabs_discovered( "tf_mvm_tabs_discovered", "0", FCVAR_ARCHIVE, "Remember how many times players have clicked tabs." );
-ConVar tf_mvm_play_music( "tf_mvm_play_music", "1", FCVAR_ARCHIVE, "Play the cut MvM upgrade station music." );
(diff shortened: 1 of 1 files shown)


Dessensce:
i personally feel like it would be better as a map edit, where it plays as an ambient_generic in a radius, instead of just when the UI is open

Bitl:
> i personally feel like it would be better as a map edit, where it plays as an ambient_generic in a radius, instead of just when the UI is open

This isn't really the point of it though. This implementation intends to directly integrate the upgrade machine music completely clientside, which helps mappers stay under the entity limit and prevents other users from hearing it when you're in the upgrade station zone. The user can also enable or disable the music if they wish, which they can't do in other map implementations.

gidi30:
I feel like the description of the CVar calling the music cut doesn't really feel right as if this gets merged the music would no longer be cut.
Something like "Play music while in the mvm upgrade menu." would be better.

Bitl:
That has been implemented

Tkain:
It's worth noting that [Freaky Fair](https://wiki.teamfortress.com/wiki/Freaky_Fair) has `ambient_generic` entities playing music at its upgrade stations, so this PR's upgrade station music should probably be disabled by default on that map to prevent overlap. I'm not sure what the most sensible way to disable it there would be, though.

PR #1557, opened 2025-10-19
[TF] Prevent Dragon's Fury projectile from getting reflected

Bitl:
# Description
This PR disables deflections/reflections of the Dragon's Fury flame ball. For whatever reason, this wasn't implemented into the Dragon's Fury despite the spell fireball being unable to be reflected. This change was added to ensure consistency with the spell fireball, and isn't intended to be a balance change.

### DIFF
--- a/src/game/shared/tf/tf_projectile_dragons_fury.cpp
+++ b/src/game/shared/tf/tf_projectile_dragons_fury.cpp
@@ -168,6 +168,7 @@ public:
 
 	virtual float		GetDamageRadius()	const			{ return tf_fireball_radius.GetFloat(); }
 	virtual int			GetCustomDamageType() const OVERRIDE { Assert( false ); return TF_DMG_CUSTOM_DRAGONS_FURY_IGNITE; }
+	virtual bool		IsDeflectable() OVERRIDE { return false; }
 
 	virtual void RocketTouch( CBaseEntity *pOther ) OVERRIDE
 	{


JoriKos:
I'm not sure the fireball spell and the Dragon's Fury's fireball are similar in any way though. I quite enjoy trying to airblast them, and it seems odd to make it consistent with a completely different and unrelated projectile.

Bitl:
It seems that valve didn't originally intend the dragon's fury to be reflected, seeing that there's no reflected kill icon and the projectile acts irrationally when reflected 

Seeing that they're both fireballs, it makes sense to make them work similarly.

JoriKos:
That makes more sense, in what ways does it act irrationally?

Brain-dawg:
Reflected fireballs don't even render correctly from what I've seen, a great MvM bugfix too for making DF not useless against pyro bots.

AJMickey:
> It seems that valve didn't originally intend the dragon's fury to be reflected, seeing that there's no reflected kill icon and the projectile acts irrationally when reflected
> 
> Seeing that they're both fireballs, it makes sense to make them work similarly.

I know I'm a bit late, but this makes sense. However, this is such a huge change that it has heavy balancing implications. This isn't like fixing the Iron Bomber's hitbox, or fixing the Neon Annihilator not dealing crits to players coated with gas - those were relatively small - this would very heavily change balance.

For example, right now, airblast is a very effective counter to the Dragon's Fury when used right (dare I say to the point of countering the DF). If the projectile suddenly was unreflectable, this would suddenly be spun on its head - decent Dragon's Fury pyros would mop the floor with regular pyros with very little chance for counterplay. Bottom line, it would be a very risky change to make due to the implications it could have not only on the balance of the DF and its interaction with Pyros, but the view of the playerbase on the potential for balance changes, which might not be what Valve wants.

Personally, I think it would make more sense to try and fix the rendering bugs in regard to the DF's projectiles being reflected. The lack of a kill icon is notable, but this isn't exclusive to the Dragon's Fury, and is more on Valve's end than anything.

Brain-dawg:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1557#issuecomment-3442140577

This post is extremely disingenuous and fundamentally misunderstands the problem.  This is exactly like fixing gas/neon.

In 13k hours I have almost never seen an enemy pyro deliberately counter DF by timing reflects, and the few pyros that have this timing mastered are so frustrating and obnoxious to fight with zero counterplay from the DF user, it tips the balance way too far AGAINST the DF if anything.  

The reflect timing is incredibly awkward, has almost no feedback, and the clear intended balance philosophy of the DF is exactly what you imply is a "problem" (stronger dps at the cost of airblast).  

The rendering "bugs" come from the fact that the particle effect ITSELF was designed to be a one-and-done effect, and "fixing" this involves completely replacing the particle... the particle that valve clearly intended to be used on something UN-reflectable.

DF is a det/scorch combo weapon, and classic shotgun combo is a pretty bad loadout compared to other flamers. There is still a perfectly balanced pyro vs. pyro dichotomy without reflectable fireballs.  It is not the DF users' fault that you see every problem as a nail, and your hammer (flamethrower) doesn't work in this particular situation.  If someone in the lobby is particularly obnoxious, and you insist on not switching to another class to counter it for a bit, run shotgun.  If you think this is a bad balance philosophy, I have bad news for you about the rest of the game.

This is like complaining that a Shortstop scout can kill you from further range than a scattergun. It doesn't make any sense in real gameplay and completely ignores the trade-offs that Shortstop users need to make.

AJMickey:
I see it as different because the gas/neon change was something that from what I'd seen, everyone thought and agreed was a bug for as long as I've been playing, but I've _almost_ never seen anyone consider the DF fireball being reflectable to be a bug (not to say it can't be one, but imo the _reception_ would be a very stark contrast, and I do think it impacts balance a lot more than the combo)

I don't disagree that it's annoying and awkward, for the record. DF is my most used pyro primary by far, and fighting a good pyro that knows how to airblast is **pain**. I wouldn't say there's no counterplay, though - I've found that because they expect you to hold m1, they'll airblast with a predictable timing, so delaying your shot by a split second baits the airblast and lets you land a combo hit, which tends to throw them off (still very annoying to fight muscle memory to do this though)

Sorry if I was too hardline in my initial post, I probably should've been a bit more open to the possibility. I get it feels pretty dumb considering the litany of bugs to do with the reflection, but I'm ultimately still very skeptical about changing it for aforementioned reasons, and that's why I said what I said in my initial comment - I did personally feel that the visuals not matching reflects was just an odd quirk for most of the time I've been using the weapon, since it's had quite a lot of those through its life and I just mentally chalked it up to strange coding, which probably biased me against seeing reflection as a bug. (IDK about it these days. Sorry again, I could've made my initial comment a lot better in hindsight)

Brain-dawg:
I was a bit harsh in my reply, and I apologize, PR comment sections aren't the right place for opinionated ranting like that.  The point I wanted to make was it seems to me that most of the clues point to reflecting itself as the bug rather than the bugs associated with reflecting. The heated tone was uncalled for.

It was a bit too sweeping for me to say zero counterplay on the DF side, but I do firmly believe that fighting a DF pyro is much less obnoxious and has more counterplay than fighting a non-DF pyro who's reflecting your fireballs.  Most of the time, it happens to me accidentally when they're spamming airblast or trying to reflect a rocket coming from somewhere else, maybe a fraction of a % of pyros do it intentionally.

PR #1558, opened 2025-10-20
[TF2] Fix spelled color cosmetics not changing colors in backpack after initialization of panel

ThatBapGuy:
As the title explains, this PR adds the functionality mentioned within the attached issue.

In current TF2, it grabs the paint color at the frame the panel is loaded. Which results in a static image that doesn’t associate correctly for spelled items that keep changing colors.

This PR corrects that, making spelled cosmetics that cycle with colors now animate within the backpack. The new code only runs on items that have the spell every tick whilst running only once for everything else as originally intended.

Closes https://github.com/ValveSoftware/Source-1-Games/issues/7587

SpelledCosmetics

### DIFF
--- a/src/game/client/econ/item_model_panel.cpp
+++ b/src/game/client/econ/item_model_panel.cpp
@@ -3149,7 +3149,7 @@ void CItemModelPanel::UpdatePanels( void )
 
 			if ( !bIsEconTool )
 			{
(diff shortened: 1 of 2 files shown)


FrozenDragon0:
Okay, now this will be an awesome change to have, thanks for doing this, just submitted my first pull request also.

PR #1563, opened 2025-10-22
Allow exceeding t-junction limit in VBSP

ficool2:
A problem that many modern maps have to deal with is the dreaded `Too many t-junctions to fix up!` error. T-junctions are formed to avoid visual seams in touching geometry. The limit is typically hit by heavy usage of `func_detail` geometry touching the world, or water brushes. Many modern TF2 maps tend to hit this error due to higher density of brush detail. 

In VBSP, the limit is dictated by `MAX_MAP_PRIMITIVES` and `MAX_MAP_PRIMINDICES` which are both 65536. ("Primitives" is what the fixed up geometry is called internally). Unfortunately, the BSP format stores primitive indexes as unsigned 16-bit integers, which means that these limits cannot be raised without modifying the BSP format, and therefore breaking backwards compatibility.

Rather than a fatal error when hitting the limit, this PR modifies the behavior to issue a warning and continue compiling. In practice, this doesn't appear to have significant side effects, and it's certainly better than having to reduce `func_detail` usage or convert them to entities etc. This is also better than using `-notjunc` or `-nodetailcracks` parameters, as this allows most of the seams to remain fixed.

Note: [VBSP++](https://ficool2.github.io/HammerPlusPlus-Website/tools.html) implements this change, and it has been battle-tested publicly on many maps.

### DIFF
--- a/src/utils/vbsp/faces.cpp
+++ b/src/utils/vbsp/faces.cpp
@@ -584,6 +584,9 @@ FixFaceEdges
 
 ==================
 */
+int g_numprimitives_overflow = 0;
+int g_numprimindices_overflow = 0;
+
 void FixFaceEdges (face_t **pList, face_t *f)
 {
 	int		p1, p2;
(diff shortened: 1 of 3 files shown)


misyltoad:
Hm... I think I did a different solution for the Half-Life 2 Anniversary update which involved changing how detail brushes deal with tjuncs...

Will have to see if I can dig that out.

misyltoad:
So what I did there was add an option that essentially just did:
```cpp

	if ( g_bNoTJuncsDetail )
	{
		if ( f->contents & CONTENTS_DETAIL )
			return;
	}
```
in `FixFaceEdges`

PR #1565, opened 2025-10-23
[TF2] Fix the Pumpkin Bombs Halloween spell missing an underwater version of it's rocket trail particle.

FrozenDragon0:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/7619

Adds a unique particle for the pumpkin bombs spell when a rocket launcher with the spell is fired underwater.
The modified .PCF file with this new particle effect added can be found in the .zip below
[Rocket fix.zip](https://github.com/user-attachments/files/23094801/Rocket.fix.zip)


kvog-git:
I would recommend configuring your editor to not convert tabs to spaces. The whitespace changes in your commits make the diff hard to read.

PR #1567, opened 2025-10-23
[TF2] Fixed Pomson colliding with teammates at close range

JoriKos:
The Pomson 6000 currently immediately collides with teammates, which makes it absolutely unusable near any teammate. This fix simply makes it behave like other projectiles, such as rockets, which pass through teammates for a short duration after being fired. Does not affect the behaviour on enemy players, nor does it affect the Bison.

Demonstration of how it works with this fix: https://www.youtube.com/watch?v=NR5eppSOFgk

Partially fixes https://github.com/ValveSoftware/Source-1-Games/issues/2592

### DIFF
--- a/src/game/shared/tf/tf_projectile_energy_ring.cpp
+++ b/src/game/shared/tf/tf_projectile_energy_ring.cpp
@@ -256,11 +256,11 @@ void CTFProjectile_EnergyRing::ProjectileTouch( CBaseEntity *pOther )
 
 	if ( bCombatEntity )
 	{
-		// Bison projectiles shouldn't collide with friendly things
-		if ((ShouldPenetrate() && (pOther->InSameTeam(this) || (gpGlobals->curtime - m_flLastHitTime) < tf_bison_tick_time.GetFloat())) || // Bison
-			(!ShouldPenetrate() && pOther->InSameTeam(this) && !CanCollideWithTeammates())) // Pomson
+		// Bison projectiles shouldn't collide with friendly things, Pomson should only collide after default no-collision period is over
+		if ((ShouldPenetrate() && (pOther->InSameTeam(this) || (gpGlobals->curtime - m_flLastHitTime) < tf_bison_tick_time.GetFloat())) ||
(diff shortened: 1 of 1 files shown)


JoriKos:
I'd also like to point out this PR was created at the most unfortunate time, as I initially started writing it mere minutes after the update dropped.

FlaminSarge:
Does this cover any cases that https://github.com/ValveSoftware/source-sdk-2013/pull/1232 does not?

JoriKos:
> Does this cover any cases that #1232 does not?

Weird, that one didn't show up when I looked for it when I first looked into this. It does seem much less efficient, spliting up the single if statement into 4 separate ones (which seems a bit excessive) and (also seemingly unnecessarily) checks for it being a player. 

Other than that it seems like they both achieve the same objective, not sure what the standard is for a situation like this.

FlaminSarge:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1567#issuecomment-3442252877

I would argue that the way the other PR is laid out makes the code far more readable and understandable than a one-liner, despite the capability to make it a one-liner.

The player check might need to be looked at, though; do these things need to still collide with buildings/etc? From a logical perspective, checking for player before checking CanCollideWithTeammates makes sense, since Teammates wouldn't include buildings.

juliarose:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1567#issuecomment-3445521088

Without the IsPlayer check the projectile will pass through buildings,  the Engineer is not normally able to shoot through his or friendly buildings (I am pretty sure.) It's also identical in behavior to Jarate/Mad Milk. https://github.com/ValveSoftware/source-sdk-2013/blob/a36ead80b3ede9f269314c08edd3ecc23de4b160/src/game/shared/tf/tf_weapon_jar.cpp#L507

I rewrote the if statement because it was hard to read. This also avoids calling ShouldPenetrate or InSameTeam more than once.

JoriKos:
> Without the IsPlayer check the projectile will pass through buildings

That makes a lot of sense actually. I've closed the PR in favour of yours, though I do still prefer the formatting to be one line :P

PR #1574, opened 2025-10-24
[TF2] Fix high melee firing speed not working correctly

naydef:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/6223

I'm a bit concerned if this fix will be working correctly if `m_flNextPrimaryAttack` becomes too high (loss of resolution).
Maybe it does not matter...

Suggestions welcomed

### DIFF
--- a/src/game/shared/tf/tf_weaponbase_melee.cpp
+++ b/src/game/shared/tf/tf_weaponbase_melee.cpp
@@ -271,7 +271,8 @@ void CTFWeaponBaseMelee::PlaySwingSound( void )
 //-----------------------------------------------------------------------------
 void CTFWeaponBaseMelee::Swing( CTFPlayer *pPlayer )
 {
-	if ( m_flSmackTime>=m_flNextPrimaryAttack )
+	// A previous smack could be pending, so do the smack before we forget about it
+	if ( m_flSmackTime >= m_flNextPrimaryAttack )
 	{
 		Smack();
 	}


src/game/shared/tf/tf_weaponbase_melee.cpp
```diff
@@ -320,6 +320,10 @@ void CTFWeaponBaseMelee::Swing( CTFPlayer *pPlayer )
 #endif
 
 	m_flSmackTime = GetSmackTime( m_iWeaponMode );
+	if(m_flSmackTime >= m_flNextPrimaryAttack)
```
ficool2: Match the style format of Valve code (space between parentheses), space after `if`

src/game/shared/tf/tf_weaponbase_melee.cpp
```diff
@@ -320,6 +320,10 @@ void CTFWeaponBaseMelee::Swing( CTFPlayer *pPlayer )
...
+	{
+		m_flSmackTime = m_flNextPrimaryAttack - 0.01;
```
ficool2: `0.01`f to avoid double to float conversion

ficool2: Not only can this be problematic with float resolution, it won't work correctly on servers running higher tickrates (like 100)

An alternative approach to this whole PR which I suggest you try, is to run Smack() if smacktime >= primaryattacktime when the player tries attacking again

naydef: Yea, I knew this is not the best approach. I'll try your idea and report back.

naydef: Applied suggestion

naydef:
I tested ficool's suggestion and it seems to be working just fine.

Tested with the following SM plugin: https://gist.github.com/naydef/2dc40f8c881b09c8d375d7ebeffe2ee8

src/game/shared/tf/tf_weaponbase_melee.cpp
```diff
@@ -271,6 +271,11 @@ void CTFWeaponBaseMelee::PlaySwingSound( void )
 //-----------------------------------------------------------------------------
 void CTFWeaponBaseMelee::Swing( CTFPlayer *pPlayer )
 {
+	if ( m_flSmackTime>=m_flNextPrimaryAttack )
```
ficool2: Add spaces around the operator, and also add a comment mentioning why this exists

PR #1575, opened 2025-10-24
[TF] Fix crash in CEntityBird::SpawnRandomBirds caused by recent bird fix

Bitl:
# Description
This PR fixes a crash in #1315 when calling the CEntityBird::SpawnRandomBirds function. The original PR changes the KeyValues definition to a KeyValuesAD, meaning the KeyValues file is deleted from memory before the pFileKV->deleteThis(); at the end of the file is called.

### DIFF
--- a/src/game/server/tf/entity_bird.cpp
+++ b/src/game/server/tf/entity_bird.cpp
@@ -139,8 +139,6 @@ void CEntityBird::SpawnRandomBirds( void )
 			}
 		}
 	}
-
-	pFileKV->deleteThis();
 }
 
 


CosminPerRam:
Deeply sorry about it, not sure how I missed it.

bakugo:
Where is this `SpawnRandomBirds` function even called from? Maybe I'm just blind, but I can't find it.

gidi30:
> Where is this `SpawnRandomBirds` function even called from? Maybe I'm just blind, but I can't find it.

It's no longer called at any point since the function was part of the teasers for the Über update.

kvog-git:
It's also completely optimized away in release builds. I'm curious how this change (or the original PR, for that matter) was tested?

CosminPerRam:
> It's also completely optimized away in release builds. I'm curious how this change (or the original PR, for that matter) was tested?

I saw that it is not used in the sdk at all and github search also yielded no usages across forks, unfortunately due to my ignorance I indeed did not test its behavior at all, which I should have done.

PR #1580, opened 2025-10-25
CTFClassMenu: add key binds for selecting the random class

CosminPerRam:
Adds keybinds for selecting the random class, this was requested in the [subreddit](https://www.reddit.com/r/tf2/comments/1oerhjr/comment/nl3n5az/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button).

**EDIT**: After revisions, this PR adds this via the `backspace` key now, not `0`, see first point of this [comment](https://github.com/ValveSoftware/source-sdk-2013/pull/1580#issuecomment-3447861908).

Video (can't really show it but I'm pressing `0`). 

https://github.com/user-attachments/assets/6e31a30d-a180-4d2a-84c6-22496e99db68

Note that I've also edited the `tf2_misc` vpk in `tf2_misc > resource > ui > classselection.res`: "Resource/UI/ClassSelection.res" > "random" to include:
```
		"labelText"			"0"	[$WIN32]
		"labelText"			""		[$X360]
```
for proper rendering, but this feature works without it.

### DIFF
--- a/src/game/client/tf/vgui/tf_classmenu.cpp
+++ b/src/game/client/tf/vgui/tf_classmenu.cpp
@@ -875,7 +875,7 @@ void CTFClassMenu::OnKeyCodePressed( KeyCode code )
 {
 	m_KeyRepeat.KeyDown( code );
 
-	if ( code == KEY_0 || code == KEY_PAD_0 ) {
+	if ( code == KEY_BACKSPACE ) {
 		SelectClass( TF_CLASS_RANDOM  );
 		Go();
 	}


src/game/client/tf/vgui/tf_classmenu.cpp
```diff
@@ -875,7 +875,11 @@ void CTFClassMenu::OnKeyCodePressed( KeyCode code )
 {
 	m_KeyRepeat.KeyDown( code );
 
-	if ( code > KEY_0 && code <= KEY_9 )
+	if ( code == KEY_0 || code == KEY_PAD_0 ) {
+		SelectClass( iRemapIndexToClass[ TF_CLASS_RANDOM ] );
```
kvog-git:
```suggestion
		SelectClass( TF_CLASS_RANDOM  );
```

`iRemapIndexToClass` maps the menu index to the TF_CLASS enum. No reason to use it if you're passing `TF_CLASS_RANDOM` directly.

Tkain:
The 0 key already has an action for the class selection menu; it cancels the class selection menu when the player has already chosen a class. The label for the 'cancel' button at the bottom of the menu also indicates that the 0 key is a shortcut for it.

This functionality is handled by the following branch, which this PR would make unreachable with `KEY_0` and `KEY_PAD_0`:

https://github.com/ValveSoftware/source-sdk-2013/blob/a36ead80b3ede9f269314c08edd3ecc23de4b160/src/game/client/tf/vgui/tf_classmenu.cpp#L896-L909

CosminPerRam:
I haven't played MvM (or any gamemodes that require ready-up), totally went over my head.

Here are some potential candidates to implement this feature:
1. Change `0` to something like `Backspace`: (and use the symbol (`←` or `⌫`) because the word is too long for the ui element (or widen the ui element for the word))
2. Change the Cancel keybind to Backspace: This is quite disruptive as it will break muscle memory for players.

Thoughts @Tkain?

Tkain:
The issue I mentioned doesn't have to do with MvM or readying up - the cancel button shows up whenever you open the class selection menu after having chosen a class. You can see it in the video you posted about 10 seconds in.

In any case, both candidates would solve the issue; I would probably recommend the former over the latter to avoid breaking players' muscle memory.

CosminPerRam:
> You can see it in the video you posted about 10 seconds in.

I made these changes very late and my head was in another place, thanks for the thoughts.

PR #845, opened 2025-02-26
[TF2] Fix Minigun barrel movement being updated multiple times per tick

Marioiscool246:
Fixes the minigun viewmodel having crazy high barrel spin velocities on high latency servers.

Unfixed behavior:

https://github.com/user-attachments/assets/af1d8046-454d-455a-a353-174e339dd80f

Fixed behavior:

https://github.com/user-attachments/assets/a7b333b4-16ab-43c5-a072-4c428819d6ac

PR #1581, opened 2025-10-25
[TF2] fix miniguns not spinning or spinning very slowly outside of first-person

wgetJane:
fixes a bug introduced by a recent patch: https://github.com/ValveSoftware/source-sdk-2013/pull/845#issuecomment-3445127597

fixes ValveSoftware/Source-1-Games/issues/7636

(i tested behaviours before that patch, after that patch, and after this patch)

note to other contributors submitting prediction-related fixes:

`prediction->IsFirstTimePredicted()` should only be used alone if you're sure that it's always called in prediction

because it always returns `false` outside of prediction

otherwise check for `prediction->InPrediction()` first

### DIFF
--- a/src/game/shared/tf/tf_weapon_minigun.cpp
+++ b/src/game/shared/tf/tf_weapon_minigun.cpp
@@ -1095,7 +1095,7 @@ void CTFMinigun::StandardBlendingRules( CStudioHdr *hdr, Vector pos[], Quaternio
 //-----------------------------------------------------------------------------
 void CTFMinigun::UpdateBarrelMovement()
 {
-	if ( !prediction->IsFirstTimePredicted() )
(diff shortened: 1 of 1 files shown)


wgetJane:
to be clear, "third-person" here means the miniguns on other players, not on your own player (using the thirdperson command)

Marioiscool246:
I originally thought this was only related to the barrel movement on the viewmodel, since it didn't have the spin speed stuff going on in thirdperson. I am still new to source engine stuff so thanks for pointing this out :)

wgetJane:
> I originally thought this was only related to the barrel movement on the viewmodel, since it didn't have the spin speed stuff going on in thirdperson. I am still new to source engine stuff so thanks for pointing this out :)

yeah thats fair, it's not exactly well-documented at all

PR #1582, opened 2025-10-25
Fix Chromatic Corruption's animation not looping smoothly like all the other paint spells do.

FrozenDragon0:
Fix Chromatic Corruption's animation not looping smoothly.

Fixes: https://github.com/ValveSoftware/Source-1-Games/issues/7640

This makes the spell's animation end smooth like all the other paint spells. Looks a bit faster but it's better than having a pretty obvious cut off.
This commit also adds some commented out text next to each of the paint spells showing their actual name to not get confused with their internal code names.

Old:
https://wiki.teamfortress.com/wiki/Chromatic_Corruption_(halloween_spell)#/media/File:Chromatic_Corruption_Brigade_Helm.gif

New:

https://github.com/user-attachments/assets/b6ce98a1-1ed4-4029-b3d3-486f023ef62f

### DIFF
--- a/src/game/shared/econ/econ_item_view.cpp
+++ b/src/game/shared/econ/econ_item_view.cpp
@@ -1249,7 +1249,7 @@ static int RemapOverridePaintIndexToRGB( uint32 unIndex, uint32 unTeamIndex )
 {
 	enum { kSamplePoints = 256, };
 
-	static uint32 k_unWitchYellow[] = {
+	static uint32 k_unWitchYellow[] = { // Die Job
 		5328971, 5328971, 5328971, 5328971, 5328971, 5328971, 5328971, 5328971, 
 		5328971, 5328971, 5395018, 5591625, 5723465, 5855050, 5921095, 6052679, 
 		6315591, 6447429, 6513222, 6710852, 7039299, 7170884, 7433793, 7631425, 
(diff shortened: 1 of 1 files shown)


isthatted-tf2:
Please don't add this, it makes the spell look worse. People love the spell as is.


Edit: After seeing the updated reference, I like this change it is a legitimate bug fix. Should get added

PimpyStrussel:
Hey, see what your trying to do. I think most people would prefer the spell stay as is. Its been the same for 12 years and looks great, It is considered the best paint spell in the game, I would strongly suggest just leaving chromatic as is.

FrozenDragon0:
> Please don't add this, it makes the spell look worse. People love the spell as is.

The example I showed was crap as the .gif was shorter than the spell's full animation time, so getting a better one.

FrozenDragon0:
New video of the updated changes, my apologies for the previous reference being crap.

https://github.com/user-attachments/assets/daafd1ec-8a3e-4deb-8a96-0b77f51dba6e

KimmyTF2:
> Hey, see what your trying to do. I think most people would prefer the spell stay as is. Its been the same for 12 years and looks great, It is considered the best paint spell in the game, I would strongly suggest just leaving chromatic as is.

https://github.com/ValveSoftware/source-sdk-2013/pull/1559
This was changed just 3 days ago after not being touched for 13 years, and I own 2 items with this spell applied. I don't complain, as this looks much better now, but this only adds confusion with how Valve handles old items. Spells are super old and very expensive now, but touching an all-year visible item that messes up with team recognition is apparently not possible because "people already paid for it and might like it."

PR #1583, opened 2025-10-25
[TF2] Fix being able to pick up tier 1 magic spells without a spellbook equipped.

FrozenDragon0:
Fixes: https://github.com/ValveSoftware/Source-1-Games/issues/7641

This PR prevents players from being able to pick up tier 1 magic spells if they don't have a spellbook equipped, this behavior already exists for tier 0 magic spells but not tier 1 currently.

With this fix, attempting to pick up a tier 1 magic spell without a spellbook equipped will notify you to equip a spellbook in your action slot before picking it up.

### DIFF
--- a/src/game/server/tf/halloween/spell/tf_spell_pickup.cpp
+++ b/src/game/server/tf/halloween/spell/tf_spell_pickup.cpp
@@ -69,10 +69,10 @@ bool CSpellPickup::ItemCanBeTouchedByPlayer( CBasePlayer *pPlayer )
 	if ( IsDisabled() )
 		return false;
 
-	// Dont let them pick up new spells if they already have a spell unless its a tier 1 spell
 	CTFPlayer *pTFPlayer = ToTFPlayer( pPlayer );
-	if ( pTFPlayer && m_nTier == 0 )
+	if ( pTFPlayer )
 	{
+		// Check for spellbook regardless of tier
 		CTFSpellBook *pSpellBook = dynamic_cast< CTFSpellBook* >( pTFPlayer->GetEntityForLoadoutSlot( LOADOUT_POSITION_ACTION ) );
 		if ( !pSpellBook )
 		{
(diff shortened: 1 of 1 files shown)


Kenzzer:
The fact there's a comment stating the following
`// Dont let them pick up new spells if they already have a spell **unless its a tier 1 spell**`
means it's not a bug and doesn't require a fix

This is likely this way, to allow any players in denying high-tier spells

FrozenDragon0:
> The fact there's a comment stating the following `// Dont let them pick up new spells if they already have a spell **unless its a tier 1 spell**` means it's not a bug and doesn't require a fix
> 
> This is likely this way, to allow any players in denying high-tier spells

I think you may be getting a bit confused, that line is mentioning the existing behavior for tier 1 spells which allow you to pick them up even if you already have a tier 0 spell. 
The bug this PR is meant to fix is being able to pick up tier 1 spells without a spellbook equipped in your action slot, doing so in the base game causes the tier 1 spellbook to disappear as if it was picked up even though the player has no spellbook equipped, whereas if you attempt to do the same for a tier 0 spell, the game just provides you the message about equipping a spellbook to pick up the powerup.

The fix here is to make it so if you attempt to pick up a tier 1 spell without a spellbook equipped in your action slot then the game denies it and gives you the message to equip one before picking it up.

Kenzzer:
Like you say
> The fix here is to make it so if you attempt to pick up a tier 1 spell without a spellbook equipped in your action slot then the game denies it and gives you the message to equip one before picking it up

This still changes the game behaviour, the ability to pick up tier 1 spells without a spellbook was done on purpose (as per valve comment). Your changes would make the tier1 spells undeniable by players that dont have the spellbook equipped, while it seems it was done purpose by valve.

It's a good idea to notify players that they need to equip a spell book, but they should still be able to deny tier 1 spells regardless or whether or not they've the spell book equipped.

FlaminSarge:
> (as per valve comment)

The valve comment only mentions the case where the player already has a spell, and does not mention the case where the player does not have a spellbook at all. I'm not sure I follow this logic.

I don't have an opinion on whether players should be able to eat t1 spellbooks without having a spellbook equipped, just that I don't agree with you referencing the valve comment as them having this specific behavior 'on purpose'.

FrozenDragon0:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1583#issuecomment-3470883730

That's what I'm trying to understand, I highly doubt that Valve allowing tier 1 spellbooks to be picked up without a spellbook was intentional behavior and more of an accidental oversight.

PR #1585, opened 2025-10-25
matsys_controls: Fix model panels not encoding pose parameter values

copperpixel:
# Description
This issue caused Mutated Milk and Self-Aware Beauty Mark to not render properly in the TF loadout screen and 3D class portrait on the HUD.

These weapons override the player model's pose parameters to values declared in the econ item schema. 

`CTFPlayerModelPanel::SwitchHeldItemTo` would attempt to apply these values by calling `CMDLPanel::SetPoseParameterByName`, however that function wasn't setting them correctly since it would use the passed in ranged values directly without prior conversion to an encoded 0..1 value.

This PR addresses this by calling `Studio_SetPoseParameter` to encode the value, then using that to set the pose parameter.

| Before | After |
|--------|--------|
| scout_before | scout_after |
| sniper_before | sniper_after |

Partially resolves https://github.com/ValveSoftware/source-1-games/issues/3518


copperpixel:
Also can be observed with Scout lunch box items, The Gas Passer and The Hot Hand although I could only spot a noticeable difference with the latter.

| Before | After |
|--------|--------|
| hothand_before | hothand_after |

PR #1498, opened 2025-08-17
[TF2] Updated armory controls to mimic December 2021 Workshop dialog QOL update

bobatealee:
This pull request adds support for the armory (Mann Co. Catalog) mimicking the following features from the Workshop dialog, which gained them in the [December 17, 2021 patch](https://www.teamfortress.com/post.php?id=102740):

- Previous/next buttons now have support for wrapping around to the first/last page
- Added skip previous/next buttons, which skip several pages at once (commands: `prevpageskip`, `nextpageskip`)
  - Controlled by `tf_armory_page_skip`; default value is 10
- Added skip to start/end buttons, which bring you to the first/last page (commands: `skiptostart`, `skiptoend`)

Additionally, I've modified the .res file responsible for this menu with support for the new button layout and minor consistency fixes, [expanded_armory_controls_res.zip](https://github.com/user-attachments/files/21825210/expanded_armory_controls_res.zip).

expanded_armory_controls

PR #1586, opened 2025-10-26
CTFStorePreviewItemPanel: use https instead http to avoid redirect

CosminPerRam:
Accessing the link auto redirects (via 301) to https, but this can be avoided completely.
The similar has been done in https://github.com/ValveSoftware/source-sdk-2013/pull/1498/commits/12c68fb1653011d617fc48838c2f58a6f36cb32c.


CosminPerRam:
Looking again into PR's that make such changes as this one, I will close this in favor of [#905](https://github.com/ValveSoftware/source-sdk-2013/pull/905) as it contains this change amongst many others of this kind.

@Bitl tagging you because of the reference, as you might be interested.

PR #1591, opened 2025-10-28
Fix teleporter_build_rate_multiplier applying to Dispenser

Drgregs-Marcus:
Fixes a typo where the dispenser case was calling the teleporter_build_rate_multiplier; now it calls "dispenser_build_rate_multiplier".

### DIFF
--- a/src/game/client/tf/vgui/tf_controls.cpp
+++ b/src/game/client/tf/vgui/tf_controls.cpp
@@ -2215,6 +2215,7 @@ void CTFLogoPanel::PaintTFLogo( float flAngle, const Color& color ) const
 void CTFLogoPanel::Paint()
 {
 	m_flOffsetAngle += gpGlobals->frametime * m_flVelocity;
+	m_flOffsetAngle = fmodf( m_flOffsetAngle, 360.f );
 	PaintTFLogo( m_flOffsetAngle, GetFgColor() );
 	BaseClass::Paint();
 }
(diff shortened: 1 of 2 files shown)


Churkinator:
You probably should state that "teleporter_build_rate_multiplier" was replaced with "dispenser_build_rate_multiplier".

Drgregs-Marcus:
> You probably should state that "teleporter_build_rate_multiplier" was replaced with "dispenser_build_rate_multiplier".

It doesn't "replace" teleporter_build_rate_multiplier; instead, it changes what attribute is referenced for OBJ_Dispenser in "CALL_ATTRIB_HOOK_FLOAT". Essentially, splitting "teleporter_build_rate_multiplier" the attribute into two, one for the dispenser and one for the teleporter.

Nonetheless, I will edit my comment to clarify what it does.

mastercoms:
`dispenser_build_rate_multiplier` doesn't exist as an attribute, it would need to be defined in items_game.txt probably to have good defaults / a correct type definition. Also, the `teleporter_build_rate_multiplier` could have effectively been used as "non-sentry" build rate since people would be aware of this behavior.

Drgregs-Marcus:
> `dispenser_build_rate_multiplier` doesn't exist as an attribute, it would need to be defined in items_game.txt probably to have good defaults / a correct type definition. Also, the `teleporter_build_rate_multiplier` could have effectively been used as "non-sentry" build rate since people would be aware of this behavior.

I'm well aware that it would need to be defined in items_game.txt, and I planned to email a Valve employee to add the attribute definition and translations if this pull request were to be merged. However, I suppose I could also include those here.
As for it being used as a "non-sentry" build attribute, I am doubtful anything actually uses it as such. and if so, it's a very simple fix.

PR #1592, opened 2025-10-28
[TF2] Remove the restriction to inspect upgrade on your teammates during MVM

num-get:
I don't see any good reason why the restriction is there.


HyperionCoding:
They specifically added that because MvM, I doubt they are removing it.

Ashe-tf:
There's a pr that adds a cvar
https://github.com/ValveSoftware/source-sdk-2013/pull/1364

FlaminSarge:
As Ashe said, this should use the cvar implementation.

PR #1595, opened 2025-10-28
[TF2] Fix airblast projectile deflection trajectory colliding with projectiles

### LINKED ISSUE #5528: [TF2] [Bug] Airblasting while aiming at a projectile causes the airblasted projectile to redirect towards the player who airblasts
This is a bug that has persisted for years, but nobody has really done a deep-dive until recently about why this has occurred.

Basically, airblasting a projectile attempts to redirect that projectile towards wherever you're currently aiming. This erroneously includes other projectiles, including the projectile you're trying to airblast. So, aiming directly at the projectile you're trying to airblast will make the game think that the direction you're trying to airblast the projectile in is towards yourself, since your crosshair collides with the very side of the projectile's bounding box, which will always be closer to you than the center of the projectile, leading to the projectile to be airblasted towards you.

[This video](https://www.youtube.com/watch?v=Sz9nDnYfYOk) visually shows the bug in action and what causes it to occur.

The ideal way to fix this would be to ignore projectiles themselves and ensure that it only considers map geometry when trying to decide where to redirect the projectile.

bakugo:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/5528

Whenever a Pyro airblasts a projectile, the game does a trace from the Pyro's view to determine what surface he's aiming at, and redirects the projectile there. However, this trace can actually collide with projectiles, including the same projectile that is being deflected, which can result in the deflect sending the projectile in an unexpected direction instead of where the Pyro is actually aiming at. [Video demonstration of the bug here.](https://www.youtube.com/watch?v=Sz9nDnYfYOk)

Fixed by filtering out projectiles in `CTraceFilterDeflection`.

### DIFF
--- a/src/game/shared/tf/tf_weaponbase.cpp
+++ b/src/game/shared/tf/tf_weaponbase.cpp
@@ -5904,6 +5904,9 @@ public:
 		if ( pEntity->IsBaseObject() )
 			return false;
 
+		if ( pEntity->IsBaseProjectile() )
+			return false;
+
 		if ( pEntity->IsCombatItem() )
 			return false;
 


bakugo:
Nevermind, apparently the numerous Pyro bugs were actually intended features all along.

KimmyTF2:
Uh, can this be implemented by someone else? It would be a shame to let this fix be lost.

blackletum:
> Uh, can this be implemented by someone else? It would be a shame to let this fix be lost.

I gotchu homie:

https://github.com/ValveSoftware/source-sdk-2013/pull/1911

PR #1596, opened 2025-10-29
[TF2] Prevent cyoa_pda_open 1 from forcing Sniper to unzoom

### LINKED ISSUE #7350: [TF2] ConTracker menu exploit bypasses delay after scoped shot
After the Sniper fires a scoped shot, there's a slight delay before he's able to unscope. However, quickly opening and closing the ConTracker menu while he's scoped in will force him to immediately unscope. Performing this method right after he fires a scoped shot will essentially allow him to bypass that delay.

https://github.com/user-attachments/assets/60d20827-43f1-4df3-bb8f-baeff30a546a

liminalbloc:
Resolves [ValveSoftware/Source-1-Games#7350](https://github.com/ValveSoftware/Source-1-Games/issues/7350)

While the Sniper is zoomed in with his sniper rifle, setting  `cyoa_pda_open` to 1 will immediately force him to unzoom. This can be abused to skip the short unzoom delay and other movement penalties that occur after firing a zoomed-in shot. This change will prevent this behavior from happening.

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -12113,7 +12113,7 @@ bool CTFPlayer::CanAttack( int iCanAttackFlags )
 
 	Assert( pRules );
 
-	if ( IsViewingCYOAPDA() )
+	if ( IsViewingCYOAPDA()  && !m_Shared.InCond( TF_COND_ZOOMED ) )
 		return false;
 
 	if ( m_Shared.HasPasstimeBall() ) 


bakugo:
Delfy just uploaded a video showing this off, so it might be a good idea to merge a fix for it sooner rather than later.

That said, I'm not too sure about this specific fix. Doesn't this allow snipers to shoot while in the PDA animation?

liminalbloc:
> That said, I'm not too sure about this specific fix. Doesn't this allow snipers to shoot while in the PDA animation?

This fix shouldn't cause that and a different bug is already making it possible to do anyways just by quickly toggling `cyoa_pda_open` with the ConTracker equipped ([ValveSoftware/Source-1-Games/issues/7348](https://github.com/ValveSoftware/Source-1-Games/issues/7348)).

bakugo:
You're right, my bad. Having the ConTracker open prevents all button inputs in `CTFGameMovement::StunMove()`, but either way, that doesn't stop someone from closing it and shooting immediately after to get the same effect.

Edit: Actually, there's still a problem: this patch makes it so you don't unzoom when opening the ConTracker anymore, so you can stay zoomed and charge a headshot, then immediately shoot it when closing the ConTracker.

liminalbloc:
> Actually, there's still a problem: this patch makes it so you don't unzoom when opening the ConTracker anymore, so you can stay zoomed and charge a headshot, then immediately shoot it when closing the ConTracker.

I assumed this wasn't much of a concern in this case since you can pretty much already do the same thing with other menus like the main menu, loadout, and inventory. Opening those menus still keeps you zoomed in and doesn't stop charging the shot. I think figuring out how to handle charging while in a menu would have to be its own discussion.

PR #1597, opened 2025-10-30
[TF2] Fix `dmg_from_sentry_reduced` ignoring Sentry rockets

ChaseTownsend:
Fixed sentry rockets getting ignored by the `dmg_from_sentry_reduced`, as code used to only check if inflictor was sentrygun.
And as and extra check, only apply attribute if the owner of the rocket is a sentry.
This works because the scorer is set to players instead


FlaminSarge:
As a heads up, it looks like Eric merges PRs using the original commit text (rather than the PR title); for the sake of clarity in the commit log, it might be worth updating/force-pushing this branch with a better commit message (1st line) instead of "Implement Fix"

FlaminSarge:
(For future reference:
```
git commit --amend -m "New commit message"
git push -f
```
should handle renaming the commit locally without needing to open a new PR; it'll always take the current state from whatever branch you're working on even if you overwrite it with a force-push; force pushes are normally dangerous but it's probably fine to edit a commit message this way)

PR #1600, opened 2025-10-30
[TF2] Improved huntsman arrow hitbox selection (no raycast approach)

wgetJane:
# Description

fixes ValveSoftware/Source-1-Games/issues/5904

supersedes #634

alternative: #1593 (simpler solution)

this pr's solution:
\- find point on the arrow's trajectory line that's closest to the hitbox centre
\- from that trajectory point, find distance to closest point on the hitbox's surface
\- no intersection tests (a lot cheaper)
image

solution in #1593 (cross product):
\- from the arrow's impact point, raycast to hitbox origin
\- from that raycast's intersection point, find perpendicular distance to arrow's trajectory line
image

solution in a4f7b4f (projection):
\- finds point on the trajectory line that's closest to the hitbox origin
\- from that trajectory point, raycast to hitbox origin (to get raycast distance)
image

original behaviour:
\- shift arrow's impact point 16 units forward
\- from that point, raycast to hitbox origin (to get raycast distance)
\- this can easily give unintuitive results (frustrating to both the user and the victim)
\- bug: hitboxes are treated as axis-aligned boxes rather than oriented boxes!!!
image

### DIFF
--- a/src/game/server/tf/tf_projectile_arrow.cpp
+++ b/src/game/server/tf/tf_projectile_arrow.cpp
@@ -816,26 +816,38 @@ void CTFProjectile_Arrow::ArrowTouch( CBaseEntity *pOther )
 	if ( !closest_box )
 	{
 		// Locate the hitbox closest to our point of impact on the collision box.
-		Vector position, start, forward, origin;
-		QAngle angles;
-		float closest_dist = 99999;
+		Vector forward, impact_pos;
+		float closest_dist = FLT_MAX;
 
 		// Intense, but extremely accurate:
 		AngleVectors( GetAbsAngles(), &forward );
-		origin = GetAbsOrigin();
+		impact_pos = GetAbsOrigin();
+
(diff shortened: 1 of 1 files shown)


wgetJane:
some extra considerations:

https://github.com/ValveSoftware/source-sdk-2013/blob/ae9287b1a4788aa5302652fe06373c113dfdb789/src/game/server/tf/tf_projectile_arrow.cpp#L799-L806

i think `start + vel * gpGlobals->frametime` here is weird, like why is it even based on frametime?

the trace distance here would typically be around ~30 units, which could fall short

imo something like `start + forward * 48` makes more sense, or maybe even longer (for giant targets)

\---
another thing: using dot product instead of distance for selecting a hitbox

this would have similar results but would be slightly biased towards farther hitboxes
(because angular changes are less pronounced from farther away)

however, that might actually be desirable, because of how perspective works
(like how farther objects are closer to your crosshair when projected onto your 2d screen)

im not too sure though since it feels subjective, idk which makes more sense perceptually

PR #1602, opened 2025-10-31
[TF2] Add visual timer for vote duration

horiuchii:
# Description
Closes [#ValveSoftware/Source-1-Games/5209](https://github.com/ValveSoftware/Source-1-Games/issues/5209#issue-1916377368)

This PR adds a timer to the vote panel, displaying how much time is left before the vote is concluded.
It also fixes a bug where `m_bIsYesNoVote` was reading a byte in the user message when it should've been reading a bit and removes the 1 second variance as it seems counter intuitive to this PR.

The modified votehud.res file can be downloaded [here](https://github.com/user-attachments/files/23249107/modified_vote_panel.zip).

https://github.com/user-attachments/assets/53c47510-abda-4c6b-85a5-d3ffb37279a4

### DIFF
--- a/src/game/client/hud_vote.cpp
+++ b/src/game/client/hud_vote.cpp
@@ -1472,9 +1472,19 @@ void CHudVote::MsgFunc_VoteStart( bf_read &msg )
 	char szParam1[k_MAX_VOTE_NAME_LENGTH] = { 0 };
 	msg.ReadString( szParam1, sizeof(szParam1) );
 
-	pVotePanel->m_bIsYesNoVote = msg.ReadByte();
+	pVotePanel->m_bIsYesNoVote = msg.ReadOneBit();
 	int iTargetEntIndex = msg.ReadByte();
 
+	// Duration of the vote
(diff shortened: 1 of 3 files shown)


wgetJane:
i'm curious why the voting window even has a 1-second variance

seems bizarre

Squid-Eevee:
If I could suggest a tiny detail, votes that only pertain to a specific team could use the team coloured capture progress icons, while server-wide votes such as voting for the next map could use the neutral one.

FlaminSarge:
> i'm curious why the voting window even has a 1-second variance
> 
> seems bizarre

Tinfoil theory, back when it was first implemented, if bots knew the exact vote duration they could be set up to leave right as the vote went through and break something, so with the variance that wouldn't necessarily be possible if it weren't sent to the client; if we want to preserve the spirit of that behavior, maybe keep the variance but always send the non-variance value in the msg to clients so bots can't use that for nefarious purposes.

Primary-Ad2848:
Its a very good quality of life change, I hope they will merge it.

PR #1241, opened 2025-05-05
tf: Add filter/search option in advanced options menu

copperpixel:
# Description
This PR adds the ability to filter settings in TF's advanced options menu. In order to make this work it also adds a check in `PanelListPanel::ComputeVPixelsNeeded()` and `PanelListPanel::PerformLayout()` to prevent allocating space for invisible panels.

This change should make navigation around the menu better, since I think the amount of settings makes it a bit too cluttered.

Demonstration video:

https://github.com/user-attachments/assets/f3cb4e5d-5f88-4749-84b3-de0c4221877f

# Changes in runtime files needed
Following modifications to `tf/resource/ui/tfadvancedoptionsdialog.res` need to be made, as well as localization strings for `#TF_AdvancedOptions_Filter`:

<details>
  <summary>Click to expand</summary>
  
```diff
 "Resource\UI\TFAdvancedOptionsDialog.res"
 {
 	"TFAdvancedOptionsDialog"
 	{
 		"ControlName"		"EditablePanel"
 		"fieldName"		"TFAdvancedOptionsDialog"
 		"xpos"		"c-250"
 		"ypos"		"90"
 		"wide"		"500"
 		"tall"		"350"
 		"autoResize"		"0"
 		"pinCorner"		"0"
 		"visible"		"1"
 		"enabled"		"1"
 		"tabPosition"		"0"
 		"bgcolor_override"		"46 43 42 255"
 		"paintbackgroundtype"	"2"
 		"settitlebarvisible"	""
 		"paintborder"			"1"
 		"paintbackground"		"0"
 		"border"				"GrayDialogBorder"
 		
 		"control_w"			"500"
 		"control_h"			"25"
 		"slider_w"			"500"
 		"slider_h"			"25"
 	}
 	"TitleLabel"
 	{
 		"ControlName"	"CExLabel"
 		"fieldName"		"TitleLabel"
 		"font"			"HudFontMediumBold"
 		"labelText"		"#TF_AdvancedOptions"
 		"textAlignment"	"center"
 		"xpos"			"0"
 		"ypos"			"0"
 		"zpos"			"1"
 		"wide"			"500"
 		"tall"			"40"
 		"autoResize"	"0"
 		"pinCorner"		"0"
 		"visible"		"1"
 		"enabled"		"1"
 		"fgcolor_override" "200 80 60 255"
 	}
+	"FilterLabel"
+	{
+		"ControlName"	"CExLabel"
+		"fieldName"		"FilterLabel"
[...]

PR #1605, opened 2025-10-31
vgui_controls: Restore default locale initialization in VGui_InitInterfacesList

copperpixel:
# Description
Moved from https://github.com/ValveSoftware/source-sdk-2013/pull/1241

Some time between SDK update in 2015 and 2025 this piece of code was removed which caused some Unicode functions to not work properly in non-English alphabet languages.

An example of this being an issue is in the PR linked above; it contains code which will find a match case-insensitivily in two wide strings (`wszFilterBuf` and `wszDisplayText`):
```cpp
			// Find a match
			bool bMatch = false;
			const wchar_t *pwchLetter = wszDisplayText;
			while ( *pwchLetter != L'\0' )
			{
				const wchar_t *pwchSearch = pwchLetter;
				const wchar_t *pwchTest = wszFilterBuf;
				while ( *pwchTest != L'\0' )
				{
					if ( *pwchSearch == L'\0' )
						break;
					if ( towlower( *pwchSearch ) != towlower( *pwchTest ) )
						break;
					++pwchSearch; ++pwchTest;
				}
				if ( *pwchTest == L'\0' )
				{
					bMatch = true;
					break;
				}
				++pwchLetter;
			}
```
Without setting the locales, passing a Polish accent character like `Ł`  into `towlower` fails to convert it to lowercase version `ł`, returning uppercase `Ł` instead.

This makes the  ``if ( towlower( *pwchSearch ) != towlower( *pwchTest ) )`` statement incorrectly fail when comparing `Ł` and `ł`, `Ą` and `ą`, etc.

Restoring behavior from old SDK versions fixes this.

### DIFF
--- a/src/vgui2/vgui_controls/controls.cpp
+++ b/src/vgui2/vgui_controls/controls.cpp
@@ -40,6 +40,13 @@ bool VGui_InitInterfacesList( const char *moduleName, CreateInterfaceFn *factory
 	strncpy(g_szControlsModuleName, moduleName, sizeof(g_szControlsModuleName));
 	g_szControlsModuleName[sizeof(g_szControlsModuleName) - 1] = 0;
 
+	// initialize our locale (must be done for every vgui dll/exe)
+	// "" makes it use the default locale, required to make iswprint() work correctly in different languages
+	setlocale( LC_CTYPE, "" );
+	setlocale( LC_TIME, "" );
+	setlocale( LC_COLLATE, "" );
+	setlocale( LC_MONETARY, "" );
+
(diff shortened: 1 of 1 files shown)


copperpixel:
This issue can also be observed in Team Fortress 2 here:

https://github.com/ValveSoftware/source-sdk-2013/blob/2d3a6efb50bba856a44e73d4f0098ed4a726699c/src/game/shared/econ/econ_item_view.cpp#L1917-L1949

`DoesItemPassSearchFilter` is used for the filter field in the backpack menu and Mann Co. Store.

PR #1606, opened 2025-11-01
[TF2] Fix cloak and rage meters being mispredicted

ficool2:
Spy's cloak, and items that use the rage meter (such as the Buff Banner) would throw a cascade of prediction errors every frame. This is because the client is processing `CTFPlayerShared` thinks before receiving network updates, while the server updates it before the player entity thinks. The client would end up deviating the values in networked meter properties like `m_flCloakMeter` by up to 1 unit .

This PR changes the client to run the thinks in the same order as the server. Meter draining and refilling no longer throw any prediction errors. Tested with `net_fakelag 100`:

**Edit**: as noted by wgetJane, this also fixes Cloak and Dagger noise spam (https://github.com/ValveSoftware/Source-1-Games/issues/4136)

Before:

https://github.com/user-attachments/assets/93894ee1-226b-4ad9-bc72-282ccbdbbde9

After:

https://github.com/user-attachments/assets/d387ed30-2804-4a9c-80eb-98e871741fa6

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -1406,6 +1406,12 @@ void CTFPlayerShared::OnPreDataChanged( void )
 	m_nOldDisguiseClass = GetDisguiseClass();
 	m_nOldDisguiseTeam = GetDisguiseTeam();
 	m_iOldMovementStunParity = m_iMovementStunParity;
+
+	// Local player will run this in PreThink
+	if ( !prediction->InPrediction() )
+	{
+		SharedThink();
+	}
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


wgetJane:
does this fix ValveSoftware/Source-1-Games/issues/4136?

ficool2:
> does this fix [ValveSoftware/Source-1-Games/issues/4136](https://github.com/ValveSoftware/Source-1-Games/issues/4136)?

Yes this fixes Cloak and Dagger specifically. There's still another loudness issue with the stock watches but I believe that's caused by the offhand viewmodel being mispredicted (I'm making another PR to resolve that one)

figardo:
This seems to break spy cloaking from the perspective of other players. Testing with and without #1608 didn't seem to make a difference to this.


https://github.com/user-attachments/assets/ee82bb99-fab7-4f1d-a3d7-a3ec417daee7


https://github.com/user-attachments/assets/c9069687-e501-4d08-a4c7-243bd7ad04bb

bakugo:
I believe the problem here is that `C_BasePlayer::PreThink()` is only ever called for the local player, during prediction. The other calls to `SharedThink()`/`InvisibilityThink()` would need to be preserved for non-local players.

ficool2:
> This seems to break spy cloaking from the perspective of other players. Testing with and without #1608 didn't seem to make a difference to this.
> with1606.mp4
> without1606.mp4

This is fixed now

PR #1609, opened 2025-11-01
vgui_controls: Fix TextEntry horizontal scrolling when removing selected text

copperpixel:
# Description
This PR makes horizontal scroll update if needed when deleting selected text in text entry fields. This caused input to register off-screen until the cursor position catched up with the scroll start index.

Can be reproduced in developer console's command input field, however GameUI needs to be recompiled for changes to take effect.

Before:

[vgui_TextEntry_horiz_scroll_fix_before.webm](https://github.com/user-attachments/assets/2c61579a-4be0-4db7-b3ad-51bcffdc881f)

After:

[vgui_TextEntry_horiz_scroll_fix_after.webm](https://github.com/user-attachments/assets/4f10d635-0e8b-47d0-9d13-5ed373f9d15f)

### DIFF
--- a/src/vgui2/vgui_controls/TextEntry.cpp
+++ b/src/vgui2/vgui_controls/TextEntry.cpp
@@ -3140,7 +3140,7 @@ void TextEntry::DeleteSelected()
 	_cursorPos = x0;
 
 	// scroll left if we need to
-	if ( _horizScrollingAllowed && ( _cursorPos < _currentStartIndex ) )
+	if ( _cursorPos < _currentStartIndex )
 	{
 		_currentStartIndex = _cursorPos;
 	}


PiMoNFeeD:
Do you think `_horizScrollingAllowed` check is actually required? `KEY_BACKSPACE` doesn't check for it, see `TextEntry::Backspace`, it just moves the horizontal scroll regardless

copperpixel:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1609#issuecomment-4086299785

You're right. Looking back `_currentStartIndex` shouldn't ever exceed `_cursorPos` when horizontal scrolling is disabled.

PR #1611, opened 2025-11-02
[TF2] Fix players going into the reference pose when stomping

gidi30:
# Description
Currently if a Soldier Pyro or Medic meet the criteria for the stomping animation they may go into a reference pose if they use certain weapons due to missing entries in the acttables and missing activities in the player models.

For Soldier if your active weapon uses Primary2 or MeleeAllClass animations it will try to use the untranslated ACT_MP_FALLING_STOMP activity which doesnt exist for Soldier.

For Pyro if your active weapon translates the activity to ACT_MP_FALLING_STOMP_PRIMARY or ACT_MP_FALLING_STOMP_MELEE, which most of them do, it will cause the player to go into a reference pose due to Pyro not having those activities.

For Medic it will always cause a reference pose if the game decides to do the stomping activity due to him not having any stomping activities.

This PR tries to fix it by adding the missing entries to the Primary2 and MeleeAllClass acttables and adding a check to see if the model can actually use the activity before applying it.

### DIFF
--- a/src/game/shared/tf/tf_playeranimstate.cpp
+++ b/src/game/shared/tf/tf_playeranimstate.cpp
@@ -1461,7 +1461,7 @@ bool CTFPlayerAnimState::HandleJumping( Activity &idealActivity )
 		else if ( ( GetBasePlayer()->GetFlags() & FL_ONGROUND ) == 0 )
 		{
 			// In an air walk.
-			if ( m_pTFPlayer->m_Local.m_flFallVelocity > PLAYER_MAX_SAFE_FALL_SPEED && m_pTFPlayer->m_Shared.CanFallStomp() && SelectWeightedSequence( TranslateActivity( ACT_MP_FALLING_STOMP ) ) < 0 )
+			if ( m_pTFPlayer->m_Local.m_flFallVelocity > PLAYER_MAX_SAFE_FALL_SPEED && m_pTFPlayer->m_Shared.CanFallStomp() && !( SelectWeightedSequence( TranslateActivity( ACT_MP_FALLING_STOMP ) ) < 0 ) )
 			{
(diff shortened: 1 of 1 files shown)


FlaminSarge:
There are a few other cases in https://github.com/ValveSoftware/Source-1-Games/issues/4197 that this should probably handle

src/game/shared/tf/tf_playeranimstate.cpp
```diff
@@ -1461,7 +1461,7 @@ bool CTFPlayerAnimState::HandleJumping( Activity &idealActivity )
 		else if ( ( GetBasePlayer()->GetFlags() & FL_ONGROUND ) == 0 )
 		{
 			// In an air walk.
-			if ( m_pTFPlayer->m_Local.m_flFallVelocity > PLAYER_MAX_SAFE_FALL_SPEED && m_pTFPlayer->m_Shared.CanFallStomp() )
+			if ( m_pTFPlayer->m_Local.m_flFallVelocity > PLAYER_MAX_SAFE_FALL_SPEED && m_pTFPlayer->m_Shared.CanFallStomp() && !( SelectWeightedSequence( TranslateActivity( ACT_MP_FALLING_STOMP ) ) < 0 ) )
```
ZzZombo, line 1464: Sorry, why did you write the condition in such a convulted way? Just check the result is greater than or equals to 0.

gidi30: The condition was mostly copied from `HandleDucking` where it checks for `SelectWeightedSequence( TranslateActivity( ACT_MP_CROUCHWALK ) ) < 0` to early out of the function.
However you are correct I could have done `SelectWeightedSequence( TranslateActivity( ACT_MP_FALLING_STOMP ) ) >= 0`

PR #1622, opened 2025-11-05
tf: Update recent rifles tracer rounds fix to check for spectated client's cl_flipviewmodels

copperpixel:
# Description
This PR updates the fix from f3d6634f452ee1a6ca569444335589aa2397398e (https://github.com/ValveSoftware/source-sdk-2013/pull/1244) so in observer mode the tracers origin gets reoriented based on spectated client's flipped viewmodels setting by checking for `TeamFortress_ShouldFlipClientViewModel`  from 20877e40d8dc7ead02a126ecfd20e4ba33f80456 (https://github.com/ValveSoftware/source-sdk-2013/pull/1439) instead of getting the ConVar value directly.

Without this the tracers would come from the wrong side if the local client's and spectator target's flipped viewmodels setting doesn't match.

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -37,6 +37,7 @@
 // Client specific.
 #ifdef CLIENT_DLL
 #include "c_tf_player.h"
+#include "c_baseviewmodel.h"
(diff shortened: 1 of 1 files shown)


birchish:
Oh, good catch. Thank you for making a fix for this.

PR #1623, opened 2025-11-05
[TF2] Fix that allows chat filters to properly load on level start

VermilionCosmos:
This is a one file fix for the in-game chat filters. Currently they "work" if you open the chat filter window at least once. This is reset however on game close and you will need to do this every single time you open the game.

This fix allows the CVAR cl_chatfilters to be read on level start instead of only being read when the chatfilter panel renders the checkboxes.

This fix also removes the CVAR cl_chatfilter_version as it is dead code and has not done anything since its implementation (except cause chat filters to not work).

Please consider this fix as Chat Filters currently are functionally broken unless you know the work around (which I would like to stop doing). Most people think the chat filters are broken due to how they are currently not read on level start.

### DIFF
--- a/src/game/client/hud_basechat.cpp
+++ b/src/game/client/hud_basechat.cpp
@@ -37,11 +37,8 @@
 ConVar hud_saytext_time( "hud_saytext_time", "12", 0 );
 ConVar cl_showtextmsg( "cl_showtextmsg", "1", 0, "Enable/disable text messages printing on the screen." );
 ConVar cl_chatfilters( "cl_chatfilters", "63", FCVAR_CLIENTDLL | FCVAR_ARCHIVE, "Stores the chat filter settings " );
-ConVar cl_chatfilter_version( "cl_chatfilter_version", "0", FCVAR_CLIENTDLL | FCVAR_ARCHIVE | FCVAR_HIDDEN, "Stores the chat filter version" );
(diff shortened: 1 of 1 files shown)


ghost:
Why did you delete all that chatfilter version shit here? It allows newly added filters to be set to their default value without resetting any of the filters the player already set. It doesn't even break anything. I see no reason to remove it

VermilionCosmos:
> Why did you delete all that chatfilter version shit here? It allows newly added filters to be set to their default value without resetting any of the filters the player already set. It doesn't even break anything. I see no reason to remove it

Because that addition is what caused the chatfilters to not load in the first place. There also has never been, nor will there ever be at this rate, a different chat filter version. It was unused code.

ghost:
> > Why did you delete all that chatfilter version shit here? It allows newly added filters to be set to their default value without resetting any of the filters the player already set. It doesn't even break anything. I see no reason to remove it
> 
> Because that addition is what caused the chatfilters to not load in the first place. There also has never been, nor will there ever be at this rate, a different chat filter version. It was unused code.

Bruh you're completely wrong here. The problem is that `m_iFilterFlags` is incorrectly initialized to the default value when you start the game. The filterversion code doesn't even run every session, only the very first time you load a map. You should leave that shit in in case they ever add a filter for voice commands (which I made a PR for)

VermilionCosmos:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1623#issuecomment-4231213498

Then they can take your PR if that is the case. They did not merge this, so I do not understand the hostility. Be more professional please.

PR #1626, opened 2025-11-06
[TF2] Remove redundant child particle precaching

cueki:
Removes 3 redundant `PrecacheParticleSystem()` calls that attempt to precache child particles directly.

Child particles are automatically precached recursively when their parent is cached.

Changes:
- `zombie.cpp` - Removed `bomibomicon_ring` (parent `bombinomicon_burningdebris` cached in tf_weapon_spellbook.cpp)
- `tf_projectile_flare.cpp` - Removed `Explosions_MA_FlyingEmbers` (parent `ExplosionCore_MidAir` cached in tf_weapon_spellbook.cpp)
- `tf_weapon_spellbook.cpp` - Removed `merasmus_tp_bits` & `bomibomicon_ring` (parent `merasmus_tp` cached in merasmus.cpp)

### DIFF
--- a/src/game/client/vgui_netgraphpanel.cpp
+++ b/src/game/client/vgui_netgraphpanel.cpp
@@ -768,10 +768,10 @@ void CNetGraphPanel::DrawTextFields( int graphvalue, int x, int y, int w, netban
 
 	g_pMatSystemSurface->DrawColoredText( font, x, y, GRAPH_RED, GRAPH_GREEN, GRAPH_BLUE, 255, "%s", sz );
 
-	Q_snprintf( sz, sizeof( sz ), "lerp: %5.1f ms", GetClientInterpAmount() * 1000.0f );
+	Q_snprintf( sz, sizeof( sz ), "lerp: %5.1f ms", ROUND_TO_TICKS( GetClientInterpAmount() ) * 1000.0f );
 
 	int interpcolor[ 3 ] = { (int)GRAPH_RED, (int)GRAPH_GREEN, (int)GRAPH_BLUE }; 
-	float flInterp = GetClientInterpAmount();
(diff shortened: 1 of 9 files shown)


JoriKos:
Should probably have a better commit message than 'stuff'

ficool2:
I do not like this change as it violates separation of concerns, for example, if a mod removes the spellbook, the particles for the zombie would break. Entities should precache all particles they reference and not make any assumptions that other entities have precached it for them, nor should they make assumptions about how the data is structured externally.

Also redundant particle precache calls are harmless as internally it uses a hash lookup to prevent duplicate precaching

cueki:
This fix addresses what appears to be an oversight in the codebase.

- 10,102 `DmeParticleSystemDefinition` elements exist as valid precache candidates
- 332 `PrecacheParticleSystem()` calls exist in the code
- Of those 332, only 6 are `DmeParticleChild` elements
- Of those 6, 3 have their parents directly cached elsewhere

This is confusing and inconsistent. If we're precaching the parent system explicitly (which automatically precaches all children), why also attempt to precache the child directly? It adds confusion to the codebase about what actually needs to be precached.

Perhaps a better solution would be to replace these DmeParticleChild calls with their respective parents to ensure compatibility?

ficool2:
The point about separation of concerns still counts. Each entity should precache the particles it will be using and not rely on other entities. Like I mentioned, this also relies on knowledge of how the external data is structured. If I was writing this code I would not care about such details, i.e I see this particle in the particle browser and I just use it. It doesn't matter if those particles are children, parents etc, the particles in isolation should be precached and not rely on implementation details behind the scenes

cueki:
I still think the argument for consistency stands, but I agree with your last point about decoupling from implementation knowledge, thanks @ficool2.

PR #1627, opened 2025-11-06
[TF2] Allow inactive weapons to modify the 3d model in the hud

gidi30:
# Description
Currently weapons like the Gunslinger or the B.A.S.E jumper (and some others) don't appear correctly in the playerstatus and scoreboard 3d model due to only the active weapon being added to it.

This pr fixes this by adding all weapons to the hud model.

# Before:

# After:


seanmcgeehan:
Does this work for Spy disguises?

gidi30:
> Does this work for Spy disguises?

Sadly no since inactive Spy disguise weapons are not networked meaning they can't be accessed on the client.
Making them networked would require some rewrites that are best kept for their own pull request.

PR #1629, opened 2025-11-06
tier1: Fix bad thread locking code in CUtlSymbolTableMT

RoyaleNoir:
# Description

`CUtlSymbolTableMT::Find()` currently performs a read lock, despite calling `CUtlSymbolTable::Find()` which writes to the member variable `m_pUserSearchString` and relies on that value remaining unchanged for the remainder of the function.

This was actually the cause of a nasty recurring bug in the original retail version of [G String](https://store.steampowered.com/app/1224600/G_String/), where if one thread looked up `$basetexture` while another thread was performing a specific material var lookup, all subsequently drawn model materials would have their base texture set to null, causing them to fall back to the default white texture. Strictly speaking that bug is probably also possible in TF2, but does not appear to happen nearly as often, if at all.

### DIFF
--- a/src/public/tier1/utlsymbol.h
+++ b/src/public/tier1/utlsymbol.h
@@ -194,9 +194,11 @@ public:
 
 	CUtlSymbol Find( const char* pString ) const
 	{
-		m_lock.LockForRead();
+		// CUtlSymbolTable::Find() writes to m_pUserSearchString,
+		// so this needs to be a write lock.
+		m_lock.LockForWrite();
 		CUtlSymbol result = CUtlSymbolTable::Find( pString );
-		m_lock.UnlockRead();
+		m_lock.UnlockWrite();
 		return result;
 	}
 


ficool2:
The PR in its current state is likely a performance degradation, instead `m_pUserSearchString` should be stored on the stack rather than poorly fixing it like this. You will only need to modify ` CUtlSymbolTable::CLess::operator`

RoyaleNoir:
> The PR in its current state is likely a performance degradation, instead `m_pUserSearchString` should be stored on the stack rather than poorly fixing it like this. You will only need to modify ` CUtlSymbolTable::CLess::operator`

I think doing that would require completely restructuring the class, since the way `Find()` works is by passing `INVALID_STRING_INDEX` to `CUtlRBTree::Find` and then having the LessFunc use that as a flag to instead look for the member variable (and even that's implemented in a nasty way):
https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/tier1/utlsymbol.cpp#L188
https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/tier1/utlsymbol.cpp#L145

The other proper solution of modifying `CUtlRBTree` to accept a comparison function that can take extra parameters would break a lot of things because it (and by extension LessFuncs) are used all over the place.

In terms of performance impact, I can't imagine it's super high given that Source usually only has about two active threads at any given time (main thread and the queued materialsystem thread) and every other version of Source seemingly gets away fine just using write locks here (the ep1, orangebox, and swarm SDKs all use write locks).

Though in general a proper solution would be nice, yeah.

ficool2:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1629#issuecomment-3499968258

The class doesn't need a restructure, you can make a stack class like this

```cpp
	class CStringPoolIndexSearch : public CStringPoolIndex
	{
	public:
		CStringPoolIndexSearch( const char* pString ) : INVALID_STRING_INDEX
		{
			m_pUserSearchString = pString;
		}
		const char* m_pUserSearchString;
	};
```
Pass it like so

```cpp
	CStringPoolIndexSearch search( pString );
	
	UtlSymId_t idx = m_Lookup.Find( search );
```

Then in `CUtlSymbolTable::CLess::operator()` you can reinterpret the index class into this search one if its == `INVALID_STRING_INDEX`

RoyaleNoir:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1629#issuecomment-3499986835

Oh yeah that'd probably work, though it makes me a bit uneasy because I'm not sure if there are other cases where the LessFunc is called with either parameter being `INVALID_STRING_INDEX`, in which case that'd end up accessing out of bounds.

ficool2:
The only entrypoint to `CUtlSymbolTable::CLess::operator()` is from `CUtlSymbolTable::Find`, therefore it will never access out of bounds. The `CTree` class is protected so theres is no way external code would be accessing this either

ficool2:
I also just checked and CS:GO ended up using this same kind of fix of storing the pointer on the stack

PR #1631, opened 2025-11-07
Fix tf_point_weapon_mimic not using Fire Sound

WereTech:
This simply adds an EmitSound to Fire that uses `m_pzsFireSound` (which is defined, but not used anywhere) before the switch case for picking the right weapon type.

I have read and agree to the terms for making this pull request.

### DIFF
--- a/src/game/server/tf/player_vs_environment/tf_point_weapon_mimic.cpp
+++ b/src/game/server/tf/player_vs_environment/tf_point_weapon_mimic.cpp
@@ -14,6 +14,7 @@ public:
 	CTFPointWeaponMimic();
 	~CTFPointWeaponMimic();
 	virtual void Spawn();
+	virtual void Precache();
 
 	void InputFireOnce( inputdata_t& inputdata );
 	void InputFireMultiple( inputdata_t& inputdata );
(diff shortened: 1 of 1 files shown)


src/game/server/tf/player_vs_environment/tf_point_weapon_mimic.cpp
```diff
@@ -149,6 +149,8 @@ void CTFPointWeaponMimic::Fire()
 {
 	Assert( m_nWeaponType >= 0 && m_nWeaponType < WEAPON_TYPES );
 
+	PrecacheScriptSound( m_pzsFireSound );
```
ficool2: Should be precaching in `CTFPointWeaponMimic::Precache`

WereTech: I don't see this function for the entity, am I supposed to add this and use PrecacheScriptSound there?

ficool2: yes

WereTech: changed to your suggestion, was it done correctly?

src/game/server/tf/player_vs_environment/tf_point_weapon_mimic.cpp
```diff
@@ -99,13 +100,21 @@ CTFPointWeaponMimic::~CTFPointWeaponMimic()
 void CTFPointWeaponMimic::Spawn()
 {
 	BaseClass::Spawn();
+	Precache();
```
ficool2, line 103: Precache is called by the game, you don't call it yourself

WereTech: It doesn't seem to precache without the explicit call in Spawn, as removing the call will now produce "\<sound\> not precached" and not play.

PR #1632, opened 2025-11-07
Fix CancelPending input causing a crash when a logic_relay fires this to itself

WereTech:
Check that the input caller is not the logic_relay itself as this will cause I/O issues that will lead to a crash.

### DIFF
--- a/src/game/server/logicrelay.cpp
+++ b/src/game/server/logicrelay.cpp
@@ -115,6 +115,10 @@ void CLogicRelay::InputCancelPending( inputdata_t &inputdata )
 	{
 		g_EventQueue.CancelEvents( this );
 	}
+	else
+	{
+		Warning("Entity %s - (%s) is attempting to CancelPending itself. This is not allowed!\n", this->GetEntityName(), this->GetClassname());
+	}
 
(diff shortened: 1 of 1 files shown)


Brain-dawg:
Might be useful to have this print a Warning to let a mapper know they have bad I/O, rather than silently rejecting it.

WereTech:
Added the warning message. Wasn't sure how to style it though as I just looked through other warning messages and picked whatever seemed ok enough.

TeamSpen210:
It'd probably be better to see if you can make `CancelEvents` just ignore the `CancelPending` input - I think `nOutputID` would let you uniquely identify the `EventQueuePrioritizedEvent_t` which matches the  `inputdata_t`.

WereTech:
I did try to look into adjusting that function, but it seems that CancelPending only belongs to a logic_relay, so I opted into simply changing the function that calls for cancelevents to just not allow the relay to call that on itself instead of changing a function that many other things rely on which could introduce other issues if done incorrectly.

TeamSpen210:
I was imagining adding an overload, or parameter to the function.

PR #1634, opened 2025-11-08
Fix point_viewcontrol locking the player's view

WereTech:
For the sake of keeping compatibility, the camera will be reset on a new round. This also fixes the camera when point_viewcontrol is killed.

Additional changes are to remove the IsAlive check in Disable as that doesn't seem to be necessary and must be worked around with VScript to disable the point_viewcontrol on a player while they are dead.

### DIFF
--- a/src/game/shared/teamplayroundbased_gamerules.cpp
+++ b/src/game/shared/teamplayroundbased_gamerules.cpp
@@ -2879,6 +2879,7 @@ void CTeamplayRoundBasedRules::RoundRespawn( void )
 
 		if ( pPlayer )
 		{
+			pPlayer->SetViewEntity( pPlayer ); // Additionally reset the view entity for the player
 			pPlayer->ResetPerRoundStats();
 		}
 	}
--- a/src/game/shared/tf/tf_gamerules.cpp
+++ b/src/game/shared/tf/tf_gamerules.cpp
@@ -14949,17 +14949,6 @@ void CTFGameRules::RoundRespawn( void )
 	{
 		m_bHasSpawnedSoccerBall[i] = false;
 	}
-	
-	// Reset the player's view entity as they would no longer be using a camera at this point
-	for ( int i = 0; i < MAX_PLAYERS; i++ )
-	{
-		CBasePlayer *pPlayer = UTIL_PlayerByIndex(i);
-		
-		if ( pPlayer )
-		{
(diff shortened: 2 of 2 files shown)


src/game/server/player.cpp
```diff
@@ -5037,6 +5037,7 @@ void CBasePlayer::Spawn( void )
 
 	// Shared spawning code..
 	SharedSpawn();
+	SetViewEntity( this );
```
ficool2: Some custom maps intentionally respawn the player under the influence of a point_viewcontrol while expecting it to remain active, so this will break them

ficool2: Remove this and then remove the IsAlive check in `CTriggerCamera::Disable`. You might also want to add a new spawnflag to control the invulnerability (i.e how it sets `m_takedamage` on the player), because that's usually not desirable too

WereTech: > You might also want to add a new spawnflag to control the invulnerability (i.e how it sets `m_takedamage` on the player), because that's usually not desirable too

This PR is just for fixing the player's view, so that will be in a separate PR if no others exist.

WereTech: > Remove this and then remove the IsAlive check in `CTriggerCamera::Disable`.

This does not fix the issue as the player's camera will still be stuck to the viewcontrol.
Do you suggest that I move where it resets the camera? It still needs to happen somewhere, whether through SetViewEntity or calling Disable on the viewcontrol.
There is also the opportunity to set the player's viewcontrol after respawn as it sets the state before the `player_spawn` event is called.

ficool2: The concern is compatibility, some maps rely on the camera persisting when the player is dead and respawns. It should only disable on new round

I suggested removing IsAlive check because an infamous problem with cameras is not being able to disable them on suicides, usually vscript is needed to workaround it

WereTech: I moved it over to tf_gamerules RoundRespawn and removed the IsAlive check. Is this sufficient?

ficool2: This affects other games, not just TF2, so it should not be in tf_gamerules

WereTech: So should I move it to the BaseClass in teamplayroundbased_gamerules?
edit: Moved it to there

PR #1637, opened 2025-11-08
Add a new spawnflag to point_viewcontrol to toggle invulnerability of the viewer

WereTech:
This will allow creators to decide whether or not the player is allowed to take damage while using a point_viewcontrol. However, this still creates the issue of the player's current view entity not being reset by default and needs an explicit call for Disable or resetting the view entity back to the player. If pulled in conjunction with my other pull request at #1634, this handling can be done a bit better as it removes the IsAlive check from the Disable function so it could be disabled immediately on death instead of needing to wait until the player has respawned (if they are allowed to).

### DIFF
--- a/src/game/server/triggers.cpp
+++ b/src/game/server/triggers.cpp
@@ -3242,8 +3242,11 @@ void CTriggerCamera::Disable( void )
 		{
 			((CBasePlayer*)m_hPlayer.Get())->GetActiveWeapon()->RemoveEffects( EF_NODRAW );
 		}
-		//return the player to previous takedamage state
-		m_hPlayer->m_takedamage = m_nOldTakeDamage;
+		// return the player to previous takedamage state if the camera has enabled invulnerability
+		if ( !HasSpawnFlags( SF_CAMERA_PLAYER_NO_INVULN ) )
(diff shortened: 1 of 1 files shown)


src/game/server/triggers.cpp
```diff
@@ -3095,10 +3096,12 @@ void CTriggerCamera::Enable( void )
 
 	m_nPlayerButtons = pPlayer->m_nButtons;
 
-	
-	// Make the player invulnerable while under control of the camera.  This will prevent situations where the player dies while under camera control but cannot restart their game due to disabled player inputs.
 	m_nOldTakeDamage = m_hPlayer->m_takedamage;
-	m_hPlayer->m_takedamage = DAMAGE_NO;
+	if ( HasSpawnFlags( SF_CAMERA_PLAYER_INVULN ) )
```
ficool2: This condition needs to be inverted for backwards compatibility (and rename the flag accodingly). The previous behavior should exist for old maps.
It also shouldn't revert the m_takeDamage when Disabling in this case

WereTech: Made the changes

PR #1643, opened 2025-11-09
[TF2] Uncomment attribute code for every commented out attribute

Drgregs-Marcus:
Various attributes in TF2's source code are disabled for seemingly no reason, and with additions such as VScript, the SDK itself, and ``trigger_add_or_remove_tf_player_attributes``, there is no reason to keep them disabled.
The full list of attributes this would re-enable:
``addperc_ondmgdone_tmpbuff``
``override_hand_model_index``
``mini_crit_airborne_deploy``
``stickybomb_fizzle_time``
``ammo_becomes_health``
``mult_zoom_fov``
``panic_attack``

Additionally, there are three other attributes: ``disguise_speed_penalty``, ``rj_air_bombardment``, and ``set_charged_airblast``. However, they no longer function and cause compilation errors, so those will remain disabled.

This would need to be accompanied by this [pull request](https://github.com/ValveSoftware/source-sdk-2013/pull/1488)

### DIFF
--- a/src/game/shared/tf/tf_weapon_flamethrower.cpp
+++ b/src/game/shared/tf/tf_weapon_flamethrower.cpp
@@ -1232,16 +1232,16 @@ float CTFFlameThrower::GetDeflectionRadius() const
 {
 	float fMultiplier = 1.0f;
 
-	//int iChargedAirblast = 0;
-	//CALL_ATTRIB_HOOK_INT( iChargedAirblast, set_charged_airblast );
-	//if ( iChargedAirblast != 0 )
-	//{
+	// int iChargedAirblast = 0;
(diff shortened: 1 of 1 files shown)


JoriKos:
Seems like changes to tf_weapon_flamethrower.cpp only remove a bunch of spaces, and don't actually uncomment the code.

Drgregs-Marcus:
> Seems like changes to tf_weapon_flamethrower.cpp only remove a bunch of spaces, and don't actually uncomment the code.

The spaces removal was an accident, a small oversight when re-commenting out set_charged_airblast. 
I had to comment out set_charged_airblast again because it would cause compile errors and would potentially require more polish, such as HUD elements.
If you want, you can make a separate pull request to fix set_charged_airblast, and I'll link it here. It would be a great help.

PR #1645, opened 2025-11-11
[TF2] Fix to prevent "es_419" from being truncated due to buffer size

gaelcoral:
# Description
I'm not sure if this is actually used anywhere, but I noticed that the `rchCleanedCode` char buffer was set to 6, which works fine for standard ICUs.
However, for Latin American Spanish (es_419), the string contains 6 characters + `'\0'`, which makes 7, exceeding the buffer limit. As a result, it gets truncated to `es_41`.

This fix increases the buffer size to 8 to prevent truncation for es_419.

### DIFF
--- a/src/common/language.cpp
+++ b/src/common/language.cpp
@@ -105,14 +105,11 @@ ELanguage PchLanguageICUCodeToELanguage( const char *pchICUCode, ELanguage eDefa
 	if ( !pchICUCode )
 		return eDefault;
 
-	// Match to no more than the param length so either a short 'en' or
-	// full 'zh-Hant' can match
-	int nLen = Q_strlen( pchICUCode );
-
-	// we only have 5 character ICU codes so this should be enough room
-	char rchCleanedCode[ 6 ];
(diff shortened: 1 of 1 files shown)


FlaminSarge:
Looks like the max length can be 11, I know TF2 doesn't use/support any of these but for future proofing may be worth setting this to 12 or 16
https://www.localeplanet.com/icu/

ficool2:
Fyi this function is not used in any engine/game version

gaelcoral:
> Fyi this function is not used in any engine/game version

Yeah, that's what I thought too but I believe it might be used for localization in the Steam Web API? I'm not entirely sure though. 
My main intention was to fix this [issue](https://github.com/ValveSoftware/Source-1-Games/issues/6757), where TF2 and Dota 2 item names appear in English on Steam when using Latin American Spanish. However, I just noticed that in Dota 2, when calling the `GetStoreMetaData` API and appending `&language=es_41`, it correctly retrieves the Latin American Spanish strings. But if you use `&language=es_419`, it fails and only returns English strings.

So, I suppose this PR could potentially fix that issue (at least for Dota 2).

image

PR #1647, opened 2025-11-12
[HL2MP] Re-add npc_fastzombie and prop_vehicle_jeep

WonderlandWar:
On 6/14/2006, the entities **npc_fastzombie** and **prop_vehicle_jeep** were removed from the game. It has been a mystery for a lot of old Half-Life 2: Deathmatch players for years now. Many old maps used these entities, but they don't play the same with them missing.

It's only beneficial to bring them back, it restores the functionality to old maps and newer mappers can work with these entities.


hl2mp:
I support this, I would really like to have this.

Generalisk:
I think they were most likely removed by accident. But then again, this is Valve we're talking about, they work in mysterious ways.

PR #1649, opened 2025-11-13
[TF2] Implement "Keep party members on the same team" in matchmaking settings.

MaximumADHD:
This pull request attempts to implement a solution for keeping party members on the same team in casual matchmaking.

As of now, the auto-balancing system actively disregards parties and will often yank members of your party over to the opposite team. This was especially frustrating during the halloween event when we were trying to complete quests as a group and suddenly our contract points were not being counted as a group.

# What was changed

I added a bool argument `bKeepPartiesOnSameTeam` to `CTFAutobalance::FindNextCandidate`, which disqualifies players in a party of two or more from being selected as a candidate for auto-balancing when `CTFParty::GetKeepPartyOnSameTeam()` is true. I also added a new bool to the protodef message `CTFGroupMatchCriteriaProto` called `keep_party_on_same_team`, which is controlled by the corresponding checkbox in the matchmaking settings. It is also hooked up to the existing ConVar `tf_party_keep_on_same_team`, which now functions with a default value of `1`.

The logic in `CTFAutobalance::FindCandidates` now runs `FindNextCandidate` up to two time each pass: 
* Once to find someone who is either not in a party, or a member in a party that is okay with being put on the other team.
* If the criteria above fails, auto-balance will have no choice but to force a party to be split.

# Potential Issues

This change _may_ conflict with the skill-based design intent of auto-balance, but I think it would provide a better user experience for groups of friends playing TF2. Ultimately it's up to the discretion of the TF team, but I hope the concern I've brought up to necessitate this change is considered.

I also don't have a good testing suite to see if this change is effective, but it does compile and I believe it should work as intended.

### DIFF
--- a/src/game/server/tf/tf_autobalance.cpp
+++ b/src/game/server/tf/tf_autobalance.cpp
@@ -387,12 +387,12 @@ bool CTFAutobalance::FindCandidates()
 	while ( nNumFound < nTotal )
 	{
 		// find a candidate that is not currently in a party
-		CTFPlayer *pTFPlayer = FindNextCandidate(true);
+		CTFPlayer *pTFPlayer = FindNextCandidate( true );
 
 		if ( !pTFPlayer )
 		{
 			// everyone is in a party, we have no choice but to split (?)
-			pTFPlayer = FindNextCandidate(false);
+			pTFPlayer = FindNextCandidate( false );
 		}
 
 		if ( pTFPlayer )
--- a/src/game/shared/tf/tf_matchcriteria.cpp
+++ b/src/game/shared/tf/tf_matchcriteria.cpp
@@ -125,7 +125,7 @@ void ITFGroupMatchCriteria::SetCustomPingTolerance( uint32_t unCustomPingToleran
 }
 
 //-----------------------------------------------------------------------------
-void ITFGroupMatchCriteria::SetKeepPartyOnSameTeam(bool bKeepPartyOnSameTeam)
+void ITFGroupMatchCriteria::SetKeepPartyOnSameTeam( bool bKeepPartyOnSameTeam )
 {
 	if ( Proto().keep_party_on_same_team() != bKeepPartyOnSameTeam )
 	{


Platina6978:
Exactly this right here!

Even outside of halloween it is very frustrating to party up with your friends, queue for a game and then have your party broken up as soon as there is team inbalance. There's no ad-hoc connection possible so either requeueing (spending more time in the menu rather than playing together) or waiting for the map to end are your only options.

Even if it is against the skill-based design of autobalance, I think it is fair to say that system has long stopped working properly anyway, as most people will run into completely one-sided stomps 95% of the time they play the game regardless.

JoriKos:
Would this be fully implemented with the checkbox in the matchmaking settings? It seems weird to have this as a setting labeled 'coming soon' (and also it's preferable as a setting you can toggle) while also implementing it as a base thing already.

MaximumADHD:
> Would this be fully implemented with the checkbox in the matchmaking settings? It seems weird to have this as a setting labeled 'coming soon' (and also it's preferable as a setting you can toggle) while also implementing it as a base thing already.

Which checkbox are you referring to? I haven’t seen that before.

JoriKos:
Press on a party member slot > Matchmaking Settings > this menu
afbeelding

MaximumADHD:
I see. I think this change falls outside the scope of that unimplemented feature, unless you are suggesting that parties should have the option to specify whether they are okay with being split up by auto-balance? The scope of the pull request is just targeting the auto-balance behavior during live matches.

JoriKos:
Yeah IIRC this feature is generally interpreted as autobalance not splitting up parties. To me that means this PR is close to implementing this behaviour, but without the checkbox. Alternatively the checkbox could simply be removed, but I think it is preferred if it stays as an option to toggle that's on by default.

MaximumADHD:
Ah okay hang on, I was focusing too much on the bottom checkbox because I thought that’s what you were hovering over, I see what you’re referring to now. I’m blind.

I wonder if there’s a reason they didn’t implement it? Clearly it was considered, and it doesn’t seem to be too difficult to support. I think I just need to integrate the boolean flag into the CTFParty class somehow.

JoriKos:
> I wonder if there’s a reason they didn’t implement it?

I imagine it has something to do with the major lack (pun not intended) of major updates, so they likely don't see it as a priority.

MaximumADHD:
Alright, I added a bool flag to `CTFGroupMatchCriteriaProto` called `keep_party_on_same_team` which is synchronized with the server through the existing party settings logic. I've updated the original title and description to reflect what this pull request is now.

gaelcoral:
To be honest, I see the "Keep party members on the same team" option more as something that applies when queueing as a party. For example, if your party has six members and the option is disabled, the matchmaker can place some of your party members on the opposite team. This PR isn't related to that behavior at all. Still, I think having "Keep party members on the same team" implemented (the way I described) could help reduce queue times though who knows if people would find a way to abuse it.

My hot take: I'm against disabling autobalance for party members. I've noticed that matches become extremely unbalanced when there are parties involved, and in theory autobalance is supposed to help with that. (Autobalance looks for a player with a compatible MMR; it's not random, I've verified this.) But then there's the exploit where players use the "retry" command to avoid being switched, which leads to parties steamrolling matches. In my opinion, Valve should fix the retry exploit to prevent this.

I also understand that players don't want to be switched because they won't earn contract points on the opposite team. The only real solution would be for the Friendly-Fire system to still work even when a party member gets switched to the other team. Obviously without duplicating contract points (if your friend kills you to "help" it just wouldn't count).

gaelcoral:
> Even if it is against the skill-based design of autobalance, I think it is fair to say that system has long stopped working properly anyway, as most people will run into completely one-sided stomps 95% of the time they play the game regardless.

To be honest, the game modes that use VScript and do team scrambling (ZI & VSH) end up messing up a lot of players MMR. There's basically a 50/50 chance of gaining or losing MMR for no real reason, just because a band-aid fix was added so those modes would work.

This became really noticeable during Scream Fortress 2024 and 2025, where ZI is very popular (btw, it's a good game mode) and players don't even notice that their MMR is being affected + with MMR decay, because many players return to play, it makes the problem worse.

Generalisk:
it's... beautiful 🥲

SireAnster:
To whoever from Valve that may read this post.

Please not push this change into the main game. It would cause more problems than solutions. Team Stacking is already an issue and this is gonna make it more noticeable.

Don't make it a toggle or anything similar, just don't add it.

JoriKos:
This has been a requested feature, and an upcoming feature, for years. Having one team steamroll the other is a matchmaking issue, not an issue with this feature.

Sky-rym:
This is kinda unrelated tbh but similar to this it'd be nice to have a way to queue into a game as a party and actively want your party to be split from the get go so yall can play against each other, I honestly find that really fun lol.

PR #1658, opened 2025-11-16
[TF2] Fix an assertion error from trying to remove a nonexistent particle effect

Tkain:
# Description

The following function call causes an assertion error whenever a Spy begins to cloak in TF2:

https://github.com/ValveSoftware/source-sdk-2013/blob/50d5de34e2ae116230fd51683104e2ac9f201565/src/game/shared/tf/tf_player_shared.cpp#L7033

This call causes an assertion error because TF2 does not have any particle definition named `balloontoss_drip`; in fact, this is the only line in the SDK that references such a particle effect. Since this particle effect has no definition and is never created, this function call effectively does nothing. (This was likely related to a scrapped [water balloon weapon](https://wiki.teamfortress.com/wiki/Paint_Bomb) that was never fully implemented.)

This PR prevents this assertion error by commenting out the offending function call.


src/game/shared/tf/tf_player_shared.cpp
```diff
@@ -7000,7 +7000,7 @@ void CTFPlayerShared::OnAddStealthed( void )
 
 #ifdef CLIENT_DLL
 	// Remove water balloon effect if it on player
-	m_pOuter->ParticleProp()->StopParticlesNamed( "balloontoss_drip", true );
+	// m_pOuter->ParticleProp()->StopParticlesNamed( "balloontoss_drip", true );
```
copperpixel: Should probably be removed all together instead of commented out.

Tkain: Most of the code for the water balloon in TF2 is commented out rather than deleted, so I did the same to be consistent.

copperpixel: > Most of the code for the water balloon in TF2 is commented out rather than deleted, so I did the same to be consistent.

Ah, fair enough then.

PR #1661, opened 2025-11-18
[TF2] Don't discard ammo on the Force-a-Nature if the clip is full

num-get:
Basically https://github.com/ValveSoftware/source-sdk-2013/pull/1398 but for Scout's Force-a-Nature.

Before:

https://github.com/user-attachments/assets/0f22b0ac-3f92-4e48-a4e0-516da1a83b42

After:

https://github.com/user-attachments/assets/ef83ba6f-72d7-4c9c-871f-e3da3c89ade3

### DIFF
--- a/src/game/shared/tf/tf_weapon_shotgun.cpp
+++ b/src/game/shared/tf/tf_weapon_shotgun.cpp
@@ -417,7 +417,7 @@ void CTFScatterGun::FinishReload( void )
 	if ( !pOwner )
 		return;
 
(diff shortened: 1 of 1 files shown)


wgetJane:
> I'm only going to like this idea if Valve adds a 1-shell only reload animation

why would that be relevant, what this pr fixes is the FaN reload removing reserve ammo if you touch a resupply locker before finishing the reload, which is what #1398 does for single-reload weapons

PR #1663, opened 2025-11-19
[TF2] Allow HWM models and VCDs to be enabled

Tkain:
# Description

Alongside its set of usual playermodels, TF2 includes a set of *hardware morph (HWM)* playermodels that use a more advanced set of facial flexes than their regular counterparts. Previously, the use of HWM models and their facial flexes could be controlled with the variables `mp_usehwmmodels` and `mp_usehwmvcds` respectively. However, at some point in TF2's lifetime, the code for these commands was commented out, forcibly disabling HWM on all PCs regardless of these variables' values.

As this PR's existence likely suggests, I don't know why this change was made.

This PR uncomments the code for the commands `mp_usehwmmodels` and `mp_usehwmvcds`, once again allowing players to use HWM by setting their values to 1. In order to preserve the current behavior of not using HWM by default, this PR also changes both variables' default values to -1.

This resolves https://github.com/ValveSoftware/Source-1-Games/issues/4751.

### DIFF
--- a/src/game/client/c_sceneentity.cpp
+++ b/src/game/client/c_sceneentity.cpp
@@ -80,7 +80,7 @@ char const *C_SceneEntity::GetSceneFileName()
 	return g_pStringTableClientSideChoreoScenes->GetString( m_nSceneStringIndex );
 }
 
-ConVar mp_usehwmvcds( "mp_usehwmvcds", "0", NULL, "Enable the use of the hw morph vcd(s). (-1 = never, 1 = always, 0 = based upon GPU)" ); // -1 = never, 0 = if hasfastvertextextures, 1 = always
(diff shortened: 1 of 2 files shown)


num-get:
It's disabled because Valve ceased to make a separate set of VCD for HWM models to support newer taunts so it's broken anyway.

ficool2:
The efficient shaders for HWM are disabled in the engine as support was dropped (`HasFastVertexTextures` is forced off)

The HWM materials are also missing properties like to handle Jarate

I do not see any reason why you should use HWM ingame, the higher quality morphing is not noticeable. The only significant difference would be wrinklemaps but those can be made to work on non HWM too

JoriKos:
An additional reason is that skin cosmetic items don't go well with HWM models, as demonstrated with the Burly Beast (HWM on the left, regular on the right taken from the TF Wiki):

Bad_Image_hwm_model_clipping

PR #1664, opened 2025-11-19
vpc: Remove Python dependency for embedding VScripts at compile-time

copperpixel:
# Description
This Python script is used as a custom build step to embed VScripts into the C++ code to run in the VScript VM.

This can be easily done in native scripting solutions (PowerShell and Bourne shell on Windows and Linux respectively) and I don't think requires introducing a new dependency for building the sources.

### DIFF
--- a/src/devtools/bin/texttoarray.ps1
+++ b/src/devtools/bin/texttoarray.ps1
@@ -1,25 +1,31 @@
 param
 (
-    [ Parameter( Mandatory=$true ) ][ string ]$sFileName,
-    [ Parameter( Mandatory=$true ) ][ string ]$sObjName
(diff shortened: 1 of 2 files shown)


Generalisk:
I get errors while building on Windows
image

copperpixel:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1664#issuecomment-3558440840

Can you show the build logs and does it still occur with the latest version?

copperpixel:
<DETAILS>
<SUMMARY></SUMMARY>

~~Drafted for now since this can sometimes fail with multi-threaded building.~~

</DETAILS>

PR #1665, opened 2025-11-20
[TF2] Fix Spectral Spectrum BLU team coloring

SeanCole02:
This PR updates the final color to better match the design decisions behind the recently updated team spirit footprint color (#1559),  aligning it with the actual team color (instead of fading to white) while keeping the animation speed and loop unchanged.

Quick static preview of the final tick color changes (left is before, right is after):
image

Full animation:

https://github.com/user-attachments/assets/fb3a2999-fcdc-4eeb-a57d-3dccd9d8a1df


FrozenDragon0:
Not 100% sure how I feel on this, I can see the intent behind this but this would also upset a lot of people I feel.

Edit: Wouldn't mind this change now that I think about it.

PR #1670, opened 2025-11-23
all games: Fix bad default pose parameters in model panels

copperpixel:
# Description

`Studio_CalcDefaultPoseParameters` is used in `CMDLPanel::SetPoseParameters` to reset pose parameters to default values.

If a zero-state for a pose parameter doesn't exist, it sets the value to middle of the pose parameter range instead of the start contrary to `CBaseAnimating::InitBoneControllers`. 

Some animations might've been expecting defaulting to the start such as Pyro's all-class melee sequence:

| Before | After |
| ------------- | ------------- |
| matsys_controls_bad_pose_defaults_before | matsys_controls_bad_pose_defaults_after |

This PR fixes that by making `Studio_CalcDefaultPoseParameters`  set pose parameters to 0 by default.

Also it updates `CTFPlayerModelPanel::SwitchHeldItemTo` to reset all pose parameters before overriding them to values declared in the econ item schema. Previous behavior for instance caused Pyro's arm to break when switching from The Hot Hand to The Gas Passer.

Closes https://github.com/ValveSoftware/Source-1-Games/issues/4639

### DIFF
--- a/src/game/client/tf/vgui/tf_playermodelpanel.cpp
+++ b/src/game/client/tf/vgui/tf_playermodelpanel.cpp
@@ -715,6 +715,7 @@ void CTFPlayerModelPanel::SwitchHeldItemTo( CEconItemView *pItem )
 	}
 
 	// update poseparam
+	SetPoseParameters( NULL, 0 );
 	if ( pItem->GetStaticData()->GetNumPlayerPoseParameters( m_iTeam ) > 0 )
 	{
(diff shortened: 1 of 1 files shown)


copperpixel:
<details>

~~Alternate solution would be to update Studio_CalcDefaultPoseParameters in bonesetup.cpp to default to start of pose param range but I don't know if there would be any side effects in tools.~~

</details>

PR #1672, opened 2025-11-24
[TF2] Implement unused arrow-related kill icons

Tkain:
# Description

This PR implements the following unused TF2 kill icons for their respective scenarios (see [here](https://wiki.teamfortress.com/wiki/Kill_icon#Unused_content) for their images):

- Burning arrow headshot kills
- Reflected arrow kills
- Reflected arrow headshot kills
- Burning reflected arrow headshot kills

The first of these four icons is implemented with a new special damage type (`TF_DMG_CUSTOM_HEADSHOT_FLYINGBURN`) since headshots and flaming arrows are tracked in their own special damage types that cannot be combined as-is. The other three are implemented simply by checking for whether a projectile has been deflected in the process of choosing a kill icon.

### DIFF
--- a/src/game/client/tf/hud_basedeathnotice.h
+++ b/src/game/client/tf/hud_basedeathnotice.h
@@ -54,7 +54,7 @@ struct DeathNoticeItem
 
 	DeathNoticePlayer	Killer;
 	DeathNoticePlayer   Victim;
-	char		szIcon[32];		// name of icon to display
+	char		szIcon[64];		// name of icon to display
 	wchar_t		wzInfoText[32];	// any additional text to display next to icon
 	wchar_t		wzInfoTextEnd[32];	// any additional text to display next to victim name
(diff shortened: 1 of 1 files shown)


wgetJane:
have you tested these icons? i think i remember trying to implement these many many years ago and it turns out that there's a character limit that a few of the unused icon names don't fit within

Tkain:
Admittedly, I've only verified that the burning arrow headshot and reflect headshot icons work. Getting a bot to perform reflected arrow kills is not easy, unfortunately.

My latest commit should fix the buffer issue you mentioned, so the other two icons should work in theory, but I would appreciate if anyone else could verify this (especially if they have an easier way to test them).

wgetJane:
managed to find it: mastercomfig/tf2-patches-old/pull/299

i think the only thing missing is the cleaver reflect

PR #1673, opened 2025-11-25
game_controls: Fix model panels erroneously setting mouse cursor position after camera manipulation

copperpixel:
# Description
This PR adds a check to `CBaseModelPanel::OnMouseReleased` to only restore cursor position when the panel is capturing the mouse.

This caused an issue where if the mouse was released when the press wasn't started on the model panel it caused the mouse cursor to warp to a wrong position, that either being:
&nbsp;&nbsp;&nbsp;&nbsp;a) garbage memory caused by m_nClickStartX and m_nClickStartY not being initialized
&nbsp;&nbsp;&nbsp;&nbsp;b) the position of where the panel was clicked on the last time

Demonstration video of the issue (had to make the cursor very large because otherwise Shadowplay wouldn't render it into the video (?)):

[pottery_wheel_cursor_wrong_warp_before.webm](https://github.com/user-attachments/assets/f150aeb1-b828-4ae7-9153-cbd029b0a11c)

### DIFF
--- a/src/game/client/game_controls/basemodel_panel.cpp
+++ b/src/game/client/game_controls/basemodel_panel.cpp
@@ -516,6 +516,10 @@ void CBaseModelPanel::OnMouseReleased( vgui::MouseCode code )
 	if ( !m_bAllowRotation && !m_bAllowPitch )
 		return;
 
+	// Only accept manipulation if we were capturing the mouse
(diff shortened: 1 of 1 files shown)


bobatealee:
You are on fire with these!!

PR #1679, opened 2025-11-28
Allow all games to use detail prop shapes & update detail prop cvar defaults

horiuchii:
# Description
This PR simply enables detail shapes for all games, which also includes detail prop swaying and avoidance.
I'm unsure why this was never enabled for all games, nothing seems to be noticeably incorrect or bad after some basic testing in TF2.

I also updated the defaults for the CVars added to match the ones found in Day of Defeat: Source so that the new features can take effect.

### DIFF
--- a/src/game/client/detailobjectsystem.cpp
+++ b/src/game/client/detailobjectsystem.cpp
@@ -27,9 +27,7 @@
 #include <algorithm>
 #include "tier0/valve_minmax_on.h"
 
-#if defined(DOD_DLL) || defined(CSTRIKE_DLL) || defined(TF_CLIENT_DLL)
 #define USE_DETAIL_SHAPES
-#endif
 
 #ifdef USE_DETAIL_SHAPES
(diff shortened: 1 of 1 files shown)


src/game/client/detailobjectsystem.cpp
```diff
@@ -27,7 +27,7 @@
 #include <algorithm>
 #include "tier0/valve_minmax_on.h"
 
-#if defined(DOD_DLL) || defined(CSTRIKE_DLL)
+#if defined(DOD_DLL) || defined(CSTRIKE_DLL) || defined(TF_CLIENT_DLL)
```
SirYodaJedi: Why not just remove the `#ifdef`, and have it enabled for all games?  I don't see any reason to not have it in HL2DM (and I doubt it'll break HLDMS any further).

horiuchii: Updated the PR to do this instead

PR #1684, opened 2025-12-02
[TF2] Fix never falling back to secondary on spawn

nosoop:
# Description

A conditional in its current form attempts to switch to a secondary item only if it *can't* be selected.
This change fixes the typo, properly enabling the fallback to the secondary item if a primary is unavailable.

Note that taking this change in as-is results in Spy spawning with the Revolver out instead of the Knife.  Possible resolutions:

- Spy-specific code that prioritizes a knife change ahead of this, falling back to primary / secondary on failure
- Replace the entire default spawn selection code with `SwitchToNextBestWeapon( NULL );` and increase the weight of the knife in `scripts/tf_weapon_knife.ctx` (side effects of using this in `ManageRegularWeapons` are not known to me, and there may be other cases where the Revolver is preferable over the Knife)

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -4717,7 +4717,7 @@ void CTFPlayer::ManageRegularWeapons( TFPlayerClassData_t *pData )
 			if ( !pWeapon || !pWeapon->CanBeSelected() || !Weapon_Switch( pWeapon ) )
 			{
 				pWeapon = dynamic_cast<CTFWeaponBase*>( GetEntityForLoadoutSlot( LOADOUT_POSITION_SECONDARY ) );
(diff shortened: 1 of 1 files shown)


JoriKos:
I think Spy spawning with his Revolver is not that big of a deal, and probably better since most Spy players could use the hint.

nosoop:
> I think Spy spawning with his Revolver is not that big of a deal, and probably better since most Spy players could use the hint.

It's a notable change ("fix") in behavior, hence the callout.  Whether or not the current behavior in this case is desired, I'll leave up to the maintainer(s).  Open to amending the PR if they'd like.

nosoop:
On further investigation this bug also affects Demoman with a non-weapon primary slot item and weapon (non-shield) secondary.

Hex4dec:
Does this fix the Demoman A-pose bug? Because it would be a shame to see it go.

PR #1691, opened 2025-12-06
[TF2] Honourbound duels (Half-Zatoichi) is no longer affected by resistances

### LINKED ISSUE #6974: [TF2] It's possible to survive the Half-Zatoichi one-shot
Though this is mostly specific to Mannpower, it's possible to survive the supposed one-shot of the Half-Zatoichi if you're also wielding the Half-Zatoichi. A simple fix is in `tf_weapon_sword.cpp` at line 562 to increase `fDamage` from `pTFPlayerTarget->GetHealth() * 3` to a larger value.

JoriKos:
When using the Battalion's Backup and while using the Resistance powerup, it is possible to just barely survive the Half-Zatoichi instant kill. I've changed two resistance checks to also ignore honourbound duels by checking if both the attacker and victim have an honourbound weapon out.

Fixes https://github.com/ValveSoftware/Source-1-Games/issues/6974


allvei:
@JoriKos I suggest you use the same approach as the backstabs where we just ignore all resistances and continue dealing 3x damage regardless.

JoriKos:
> @JoriKos I suggest you use the same approach as the backstabs where we just ignore all resistances and continue dealing 3x damage regardless.

I did not get a notification for this, so I'm a bit late. I looked into it and this should be ignoring Mannpower powerups and the Batallion's Backup's resistance now. Not sure if there are any I missed. I also had to get an extra commit in because the original one didn't include it for some odd reason.

FlaminSarge:
Instead of checking the active weapon of the attacker, is it possible to check the weapon that dealt damage?
Forseeing a use case where custom weapons with projectiles can be set as Honorbound, in which case the active weapon may not be the one that's dealing the damage (there's argument that in that case if they've switched away it shouldn't get its instant kill effects, I guess, but just a thought).

JoriKos:
> Instead of checking the active weapon of the attacker, is it possible to check the weapon that dealt damage? Forseeing a use case where custom weapons with projectiles can be set as Honorbound, in which case the active weapon may not be the one that's dealing the damage (there's argument that in that case if they've switched away it shouldn't get its instant kill effects, I guess, but just a thought).

Checking it right now, it also ignores damage resistance when holding out the Half-Zatoichi when the damage isn't the Half-Zatoichi itself. While a niche situation, it is something I completely forgot about so I'll look into fixing it. I don't believe honourbound duels have a special damage type, it just multiplies the damage.

FlaminSarge:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1691#issuecomment-3908193593

My guess is that the weapon that dealt the damage should be the active weapon in order for Honorbound to proc, which avoids both the case I was worried about as well as the issue you saw.

The other option instead of this PR is to just increase the Honorbound multiplier from 3 to 5 or something so it kills through all resistance values available in stock gameplay, leaving any mods to adjust its setup as needed. This may need the Banner items' fill-up values tweaked if they aren't meant to instantly fill when getting an Honorbound kill (not sure if dealing more damage than the target's health lets it fill up instantly, should check that).

Third option: make the Honorbound attribute value scale the damage/be the actual multiplier. So Honorbound=5 for 5x damage when hitting other Honorbound targets. That'll require more of a rework + items_game change, but be way more flexible for modders in the future.

JoriKos:
My current plan was to simply create a custom damage type and base it on that. It creates a much simpler to read check that's similar to the backstab check + could be used for future resistances. Only issue is where to properly implement the duel check for weapons other than the Half-Zatoichi, as the only place the damage type is set is specifically in the sword file.

JoriKos:
Testing it, the duel now does the expected damage (tested it with Demo v Demo + Backup buff which did 525 damage, tested it with Soldier with Backup buff + Resist powerup which did the expected 858 damage). Not sure how to go about implementing the custom damage type into regular combat (as it currently only adds this custom damage type with the katana, but that's essentially also how it worked before it seems).

If there are any issues with it, please let me know!

src/game/shared/tf/tf_weapon_sword.cpp
```diff
@@ -563,6 +563,7 @@ float CTFKatana::GetMeleeDamage( CBaseEntity *pTarget, int* piDamageType, int* p
 			{
 				fDamage = MAX( fDamage, pTFPlayerTarget->GetHealth() * 3 );
 				*piDamageType |= DMG_DONT_COUNT_DAMAGE_TOWARDS_CRIT_RATE;
+				*piCustomDamage |= TF_DMG_CUSTOM_HONORBOUND_DUEL;
```
FlaminSarge: Custom damage is not a bitfield and can't be used like this with |=
You will have to either fully override DECAPITATE which may prevent the Zatoichi from decapitating, or figure out something else for this.

JoriKos: That makes sense, I just copied what was above it (I'm not too familiar with C++). I'll see if I can figure out something else.

Edit: I've got the decapitation to work (it wasn't working during a melee duel), the only issue is that the damage type is now explicitly a decapitation damage type. Not sure if it's necessary to change it (as the damage type only gets set in the sword file), or if any future uses of the honourbound mechanic would simply use a different, new damage type (maybe something like `TF_DMG_CUSTOM_HONOURBOUND_NODECAP`?).

JoriKos:
Before I re-open it (and possibly have to re-draft it), I hope this version is okay. I've implemented the custom damage type everywhere that `TF_DMG_CUSTOM_DECAPITATION` was also found, but I have been unable to test achievements and such with it (though I don't see a reason as to why they shouldn't work as it's just a simple extra check).

Please let me know if anything is wrong here, if I don't hear anything within a week or so I'll re-open it.

FlaminSarge:
> Before I re-open it (and possibly have to re-draft it), I hope this version is okay. I've implemented the custom damage type everywhere that `TF_DMG_CUSTOM_DECAPITATION` was also found, but I have been unable to test achievements and such with it (though I don't see a reason as to why they shouldn't work as it's just a simple extra check).
> 
> Please let me know if anything is wrong here, if I don't hear anything within a week or so I'll re-open it.

Will be able to take a look later this week, but something seems inherently incorrect about using a custom damage type/overriding the decapitate damage type for this case. I'm trying to see if there's a better way to handle this, but I could end up finding out that this is better than any other options. Sorry for the delay.

JoriKos:
Yeah the sword sets the damage type to decapitate somewhere, but it gets overriden with honourbound duels now so I've just added it to also be a decapitation everywhere else. Not sure if there is a better way to do it, but if there is please let me know.

FlaminSarge:
FYI I have a local commit for this that I feel is a lot cleaner. I'll open a PR against your fork and you can take a look and see how it goes.

JoriKos:
Yeah, sure.

FlaminSarge:
PR is up^ I'd highly recommend squashing your commits on this branch at some point, Valve tends to accept PRs without squashing and you've gone through several entirely different iterations of this.

(Out of curiosity, why was the initial 5x damage multiplier not a valid solution for this? From what I can tell, it correctly tells the game not to include the damage as a part of crit calculation, and I think the katana instantly filled buff banners anyways at 3x damage).

One thing I also noticed, there's a couple of places that `mod_pierce_resists_absorbs` takes effect that you don't have Katana handling for in your original implementation. Do we need to add it at those places? They wouldn't affect stock gameplay, only custom stuff, but they'd be a bit more involved as we'd want to not necessarily ignore the damage multipliers, only ignore them if they reduce damage. Probably best left as a separate PR.

JoriKos:
Closing this in favour of #1881.

FlaminSarge:
@JoriKos I'm still wondering why the x5 damage value idea was dropped, regarding stock gameplay specifically. I understand that the broader fix is useful for if mods change how this works.

JoriKos:
It's mostly future-proofing. Mannpower got some big balance changes a few years ago (2021?) when nothing else really got balance changes. Any changes to melee damage or resistances could possibly render the x5 damage obsolete and we'd be back with the same issue we do currently.

PR #1692, opened 2025-12-07
[TF2] Fix tf_allow_sliding_taunt

num-get:
# Description

tf_allow_sliding_taunt was introduced in the 2014 Love and War update in order to restore the old behavior where the player conserve their running momentum and slide a bit during the start of the taunt. But at no point did this console variable ever do anything meaningful at all since its introduction. So i fix that.

In addition to restoring taunt sliding, I also remove the taunt cooldown that was introduced in late 2013 to fix a taunt kill exploit if the tf_allow_sliding_taunt cvar is set to true. Back in 2013, if the player's taunt was interrupted by either sliding off an edge or an explosion, the taunt animation wouldn't get interrupted and will continue playing. On top of that, the player was also able to move freely. The player would take advantage of this exploit by taunting again in front of the enemy right when the animation trigger the taunt kill.
https://youtu.be/dULiiO2U1no

Since the Love and War update, the taunt animation will get properly removed if it is interrupted so removing the cooldown wouldn't introduce any bugs at all.

Before:

https://github.com/user-attachments/assets/fb54f59c-044b-4e59-bc97-c01f3000f20c

After:

https://github.com/user-attachments/assets/fa5e5f0b-5852-436c-8556-9e14984a3091

Edit:
The second commit removes the cooldown removal to prevent Shred Alert spam.

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -17742,9 +17742,7 @@ void CTFPlayer::OnTauntSucceeded( const char* pszSceneName, int iTauntIndex /*= 
 	m_Shared.m_iTauntIndex = iTauntIndex;
 	m_Shared.m_iTauntConcept.Set( iTauntConcept );
 	m_flTauntStartTime = gpGlobals->curtime;
-
-	if ( !tf_allow_sliding_taunt.GetBool() )
-		m_flTauntNextStartTime = m_flTauntStartTime + flDuration;
+	m_flTauntNextStartTime = m_flTauntStartTime + flDuration;
 
 	const itemid_t unTauntSourceItemID = m_TauntEconItemView.IsValid() ? m_TauntEconItemView.GetItemID() : INVALID_ITEM_ID;
 	m_Shared.m_unTauntSourceItemID_Low = unTauntSourceItemID & 0xffffffff;


FlaminSarge:
Heads up, I had been looking into this pretty extensively and there's more going on here; the reason sliding taunts worked in the first place before was due to latency: when the client taunts, the server still accepts any and all movement inputs from the client until the _client_ realizes it's taunting and stops sending them; if the client happened to get airborne due to those extra movement inputs the taunt would get interrupted and they would keep the animation and be able to do further inputs, but after their changes to add the cvar (but not make it work properly), the client would still send movement inputs but wouldn't be knocked out of the taunting state if put airborne.
With your change, the taunt gets interrupted when a client stops taunting, but in the code it looks like their original intent with the tf_allow_sliding_taunt cvar was to allow clients to keep their momentum after taunting but allow them to slide to a stop instead of stopping dead in their tracks.
I'd separate this into two cvars, one for sliding taunts and one for interrupting taunt when a client is pushed into the air (or maybe a value of 1 and 2 like with taunt-switch cvar), so that server operators can fine-tune the behavior, but just know that your current replacement of the cvar behavior doesn't fully cover the options that they seemed to have intended when they initially added the cvar.

I have a local commit that handles some of the latency jank around it which would prevent clients from setting fake lag cvars to widen the window in which they can perform this (fake lag would let someone taunt, then run all the way from one end of the roof to the other, before the game realizes they're taunting; we'd want to avoid that case while still letting players slide off roofs if they time it right). If you want, I can try to adapt this PR with my changes, let me know.

FlaminSarge:
> I also remove the taunt cooldown that was introduced in late 2013 to fix a taunt kill exploit

This taunt cooldown was added specifically for the Armageddon (Rainblower) and Shred Alert taunts as both of those spawn entities that don't get cleaned up too quickly, and exploiters could crash a server by overloading the entity list if they spammed the taunts fast enough. You may want to guard against that.

num-get:
> the client would still send movement inputs but wouldn't be knocked out of the taunting state if put airborne.

I think that has nothing to do with taunt sliding. Hence the removal.

> If you want, I can try to adapt this PR with my changes, let me know.

Feel free to make your PR. It's up to Valve to accept it.

> This taunt cooldown was added specifically for the Armageddon (Rainblower) and Shred Alert taunts as both of those spawn entities that don't get cleaned up too quickly, and exploiters could crash a server by overloading the entity list if they spammed the taunts fast enough. You may want to guard against that.

I'll figure out a way to prevent that. Spamming Shred Alert in spawn is annoying.

src/game/server/tf/tf_player.cpp
```diff
@@ -17740,7 +17742,9 @@ void CTFPlayer::OnTauntSucceeded( const char* pszSceneName, int iTauntIndex /*=
 	m_Shared.m_iTauntIndex = iTauntIndex;
 	m_Shared.m_iTauntConcept.Set( iTauntConcept );
 	m_flTauntStartTime = gpGlobals->curtime;
-	m_flTauntNextStartTime = m_flTauntStartTime + flDuration;
+
+	if ( !tf_allow_sliding_taunt.GetBool() )
+		m_flTauntNextStartTime = m_flTauntStartTime + flDuration;
```
ficool2: This specific change should be a separate PR

PR #1696, opened 2025-12-09
[HLDMS] HL1Port - fix (un)locked sounds for old doors

SirYodaJedi:
A few doors weren't changed to use an explicitly defined WAV or soundscript instead of a button index for the locked/unlocked sounds. The existing fallback behavior hardcoded such doors to use "buttons/button2.wav", instead of the proper button soundscript that matches the array that GoldSrc used; this edit uses the code from buttons.cpp to properly chose the correct soundscript to restore the "designer's original selection".

Rather than porting HLSDK code (which is technically against the license for both SDKs, for some reason), this PR duplicates a function from buttons.cpp. While I probably could've #included buttons.cpp, I don't trust that that wouldn't have broken things. HLSDK made the function accessible from anything that #includes cbase.h, which seems overkill.

This same fix can be applied to HLS, but that is a separate Git branch now, so it can't be applied in the same PR. In the unlikely event that this PR gets pulled upstream, make sure the cherry-pick it for the singleplayer branch as well.

### DIFF
--- a/src/game/server/doors.cpp
+++ b/src/game/server/doors.cpp
@@ -558,16 +558,16 @@ void CBaseDoor::Precache( void )
 		// Too short to be ANYTHING ".wav", so it must be an old button index.
 		// Call appropriate button soundscript to respect the designer's original 
 		// selection, so we don't get unresponsive doors.
-		m_ls.sLockedSound = MakeButtonSound( (int)m_ls.sLockedSound );
-		PrecacheScriptSound(m_ls.sLockedSound.ToCStr());
+		m_ls.sLockedSound = MakeButtonSound( atoi(STRING(m_ls.sLockedSound)) );
+		PrecacheScriptSound(STRING(m_ls.sLockedSound));
 	}
 	if( m_ls.sUnlockedSound != NULL_STRING && strlen((char*)STRING(m_ls.sUnlockedSound)) < 4 )
 	{
 		// Too short to be ANYTHING ".wav", so it must be an old button index.
 		// Call appropriate button soundscript to respect the designer's original 
 		// selection, so we don't get unresponsive doors.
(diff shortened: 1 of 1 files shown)


SirYodaJedi:
Note for SDK mod programmers: if you want to support the legacy GoldSrc-style locked/unlocked door sound behavior in your mod, just edit or remove the ifdefs in this PR.

Bitl:
Just read through both the Source SDK and HL1 SDK licences and nothing really comes out to me that says you can't port HL1 SDK code to the Source SDK and vice versa. The licenses only state that the code can be used to create mods for the respective engines (as long as the mods are free). There's no explicit guidelines against porting the code to different engines, and both licenses link to the Steam Subscriber Agreement which doesn't have any statements on the matter. Even then, Source mods have used HL1's SDK code for NPCs and such for years, so I don't find any issues with porting HL1 code.

SirYodaJedi:
> The licenses only state that the code can be used to create mods for the respective engines (as long as the mods are free).

It states that you can use it to create mods on the respective engine. It does not give any permissions regarding porting, which I'm sure a lawyer would say means you have none.

> Even then, Source mods have used HL1's SDK code for NPCs and such for years, so I don't find any issues with porting HL1 code.

I don't think Valve cares outside of cash cows as long as they stay within the Valve ecosystem (I'm sure the mod version of Black Mesa used HLSDK code), but some projects (ex: Mapbase) err on the safe side.

Regardless, the reason this particular PR doesn't use HLSDK code is unrelated to license issues, but rather is down to being a more "proper" (and certainly cleaner) solution.

Bitl:
> Regardless, the reason this particular PR doesn't use HLSDK code is unrelated to license issues, but rather is down to being a more "proper" (and certainly cleaner) solution.

Fair

PR #1697, opened 2025-12-10
[TF] Fix regressions caused by recent Caber explosion change

Bitl:
# Description

This PR fixes some minor regressions from the Caber explosion change introduced in #1456.

## The regressions introduced from the previous PR:

- Can now steal the PASS Time JACK.
- Can now destroy Sticky Bombs.
- Increased healing with Vampire powerup in Mannpower.
- Decreased damage with mini-crits in Mannpower.
- Slightly altered explosion damage and radius values due to attribute support.

## The changes introduced in this PR to fix the above regressions:

The Caber no longer uses DMG_MELEE explicitly. The changes are now checked amongst its existing TF_DMG_CUSTOM_STICKBOMB_EXPLOSION damage type, which brings the following changes:

- Can no longer steal the PASS Time JACK.
- The caber now properly pushes stickybombs with the force from the tf_grenade_forcefrom_blast value, but no longer breaks them as the Caber doesn't use DMG_IGNITE.
- Mannpower changes have been restored to previous settings.
- Removed attribute support and restored caber back to the original damage values.

### DIFF
--- a/src/game/shared/tf/tf_weapon_bottle.cpp
+++ b/src/game/shared/tf/tf_weapon_bottle.cpp
@@ -258,29 +258,9 @@ void CTFStickBomb::Smack( void )
 			if ( IsCurrentAttackACrit() )
 				dmgType |= DMG_CRITICAL;
 
-			float flDamage = 75.0f;
+			CTakeDamageInfo info( pTFPlayer, pTFPlayer, this, explosion, explosion, 75.0f, dmgType, TF_DMG_CUSTOM_STICKBOMB_EXPLOSION, &explosion );
 
-			float flAdjDamage = flDamage;
-			CALL_ATTRIB_HOOK_FLOAT( flAdjDamage, mult_dmg );
-
-			if ( flAdjDamage != flDamage )
-			{
-				flDamage = flAdjDamage;
-			}
-
-			CTakeDamageInfo info( pTFPlayer, pTFPlayer, this, explosion, explosion, flDamage, dmgType, TF_DMG_CUSTOM_STICKBOMB_EXPLOSION, &explosion );
-
-			float flRadius = 100.0f;
-
-			float flAdjRadius = flRadius;
-			CALL_ATTRIB_HOOK_FLOAT( flAdjRadius, mult_explosion_radius );
-
-			if ( flAdjRadius != flRadius )
-			{
-				flRadius = flAdjRadius;
-			}
-
-			CTFRadiusDamageInfo radiusinfo( &info, explosion, flRadius );
+			CTFRadiusDamageInfo radiusinfo( &info, explosion, 100.0f );
 			TFGameRules()->RadiusDamage( radiusinfo );
 		}
 #endif


mastercoms:
IMO, it would be way better to revert the DMG_MELEE dmg flag idea. It should never have been added. Instead, check against the weapon ID in the places where you _do_ want it to apply. This will guarantee the least amount of fallout from your changes, and won't require using up the DMG flag bits.

Bitl:
> IMO, it would be way better to revert the DMG_MELEE dmg flag idea. It should never have been added. Instead, check against the weapon ID in the places where you _do_ want it to apply. This will guarantee the least amount of fallout from your changes, and won't require using up the DMG flag bits.

It no longer uses any damage flags, and uses the pre-defined custom damage flag for the explosion (TF_DMG_CUSTOM_STICKBOMB_EXPLOSION)

Bitl:
Hi all, I've heard some reports that Caber damage is different before and after the update. While it isn't exactly proven by all users that it has indeed changed and may be based on placebo evidence, I have rolled out a change that attempts to fix this issue with a check. I will revert this if there's a more definite conclusion to the Caber damage issue that rules in favor of the damage not being adjusted. (which I'm pretty sure it hasn't).

Bitl:
Hi, I decided to revert the addition of the attributes entirely in case there was a difference with damage in gameplay. I still don't believe there was one, but i'll give @turtleology the benefit of the doubt here

Bitl:
Hello, I decided to revert the 2 damage commits as I decided to play in a casual server with the changes introduced in the update and have determined that **damage has not changed at all** in the new update. The caber feels the same. The caber is known for having inconsistent damage, even outside of this change, which may have resulted in the user finding these issues. The issue really is a problem with the caber having inconsistent explosion damage, and not that the addition of support for 2 attributes which have 0 to no change for the damage. The differences people have found seem to be placebo.

If you don't believe me, here's the demo file from that session.

[2025-12-10_20-35-07.zip](https://github.com/user-attachments/files/24093508/2025-12-10_20-35-07.zip)

turtleology:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1697#issuecomment-3639958182

not true. glad your 2 hours of gameplay trumps dedicated caber users.

Bitl:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1697#issuecomment-3640639027

I would appreciate any evidence or even like hard numbers for this claim, especially with the demo I provided, to see if I was wrong in this regard. That's the reason I provided it in the first place.

This is what confuses me: the caber base damage and base explosion radius were not altered in any way when I made the original change, and the only thing that alters it is if certain attributes are applied, which are not on the caber weapon itself outside of MvM and mods. Unless there's some odd game math regarding the explosion damage or radius, I don't think that any damage has been altered.

turtleology:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1697#issuecomment-3640722779

I believe that the game now is splitting the damage of melee and explosion differently. Your gameplay provides almost nothing in terms of multikill damage. 
fffffffffffffffffffffffffffffffffffff 

this image from your clip further separates the damage between melee and explosion. not to mention the 57 damage to the sentry as a secondary target which seems lower to me than normal.

Bitl:
The attribute support has been removed for the caber's explosion, and values SHOULD be at original levels.

**_this is the last alteration to this PR._**

ficool2:
Why remove the attribute support? The damage/radius attributes are not defined in items_game.txt so the values should remain the same. The attributes are useful for plugins/maps

Bitl:
This is out of date, please read https://github.com/ValveSoftware/source-sdk-2013/pull/1702

PR #1703, opened 2025-12-12
[TF2] Fix the 3d model in the hud using the incorrect animations while holding the passtime ball

gidi30:
# Description
Currently while holding the passtime ball the 3d model in the hud will use the allclass melee animations instead of the proper passtime ball animations.
This PR attempts to fix this by introducing a new weapon type for the passtime ball used for the animations.

**Implementing this will also require a change to the definition of TF_WEAPON_PASSTIME_GUN in items_game.txt**
```diff
  "item_name"	"#Weapon_Passtime_Gun"
  "item_slot"    "utility"
- "anim_slot"   "MELEE_ALLCLASS"
  "attach_to_hands""1"
  "hidden"	        "1"
```
to
```diff
  "item_name"	"#Weapon_Passtime_Gun"
  "item_slot"	"utility"
+ "anim_slot"	"PASSTIME_BALL"
  "attach_to_hands"	"1"
  "hidden"	"1"
```
# Before:

# After:


src/game/shared/tf/tf_item_constants.h
```diff
@@ -31,6 +31,7 @@ enum
 	TF_WPN_TYPE_PRIMARY2,
 	TF_WPN_TYPE_ITEM3,
 	TF_WPN_TYPE_ITEM4,
+	TF_WPM_TYPE_PASSTIME_BALL,
```
FlaminSarge, line 34: "WPM" seems like a typo

Tiagoquix: Indeed. @gidi30 @EricS-Valve

PR #1704, opened 2025-12-13
[TF2] Fix pyrovision effect for mods on TF2.

celisej567:
While pyrovision is on, overlays freaks out. This happens only in mods, and not in default tf. 

https://github.com/user-attachments/assets/608cb60d-c45c-4d5a-a9ce-fc4b54bcb0b7

After this commit.

https://github.com/user-attachments/assets/5c4cde5d-d7fd-494f-9c8c-f8b92c01a36a

Hardcoded check for gameinfo folder that prevented this was in materialsystem, but mods cant modify this, so check falsely sets mat_supportflashlight as true. Setting mat_supportflashlight in CHLClient::Init for TF fixes this issue for mods.

### DIFF
--- a/src/game/client/cdll_client_int.cpp
+++ b/src/game/client/cdll_client_int.cpp
@@ -1045,6 +1045,20 @@ int CHLClient::Init( CreateInterfaceFn appSystemFactory, CreateInterfaceFn physi
 	IGameSystem::Add( GetPredictionCopyTester() );
 #endif
 
+#ifndef DEDICATED
+
+	//usually thats in materialsystem but sdk dont have that
+	//overlays freaks out when flashlight support is on while having pyrovision
+	//this check is only for tf mods, because pyrovision is only tf thing
+#if defined( TF_CLIENT_DLL )
+	{
+		ConVarRef( "mat_supportflashlight" ).SetValue( false );
+	}
+#endif
+
+
+#endif
+
 	modemanager->Init( );
 
 	g_pClientMode->InitViewport();
--- a/game/mod_tf/cfg/mtp.cfg
+++ b/game/mod_tf/cfg/mtp.cfg
@@ -1,0 +1,14 @@
+"VisionFilterShadersMapWhitelist"
+{
+	"arena_badlands.bsp"	"1"
+	"ctf_2fort.bsp"		"1"
+	"cp_badlands.bsp"	"1"
+	"cp_dustbowl.bsp"	"1"
+	"cp_gravelpit.bsp"	"1"
+	"koth_badlands.bsp"	"1"
+	"koth_viaduct.bsp"	"1"
+	"plr_hightower.bsp"	"1"
+	"pl_badwater.bsp"	"1"
+	"pl_goldrush.bsp"	"1"
+	"sd_doomsday.bsp"	"1"
+}


ALIEN31ITA:
You have no idea how i was going crazy over this, apparently this bug DOESN'T HAPPEND to some users, my initial tought was like my Graphic card being faulty (despite not, and being fine around every other single game), and all of this because of flashlights?! 😭

celisej567:
In fact it DOES depend on GPU. Old gpus and intel. I think thats something regarding dx9 implementation on newer and some older gpus. Flashlight implementation is pretty messy.

ALIEN31ITA:
> In fact it DOES depend on GPU. Old gpus and intel. I think thats something regarding dx9 implementation on newer and some older gpus. Flashlight implementation is pretty messy.

it's a really crazy Discovery, for context, mine is a RTX 4060 TI, and it's fairly new, so stuff like this kinda makes you worried, but good finding.

src/game/client/cdll_client_int.cpp
```diff
@@ -1043,6 +1043,20 @@ int CHLClient::Init( CreateInterfaceFn appSystemFactory, CreateInterfaceFn physi
...
+#endif
+
+#ifndef DEDICATED
+
+	//usually thats in materialsystem but sdk dont have that
+	//overlays freaks out when flashlight support is on while having pyrovision
+	//this check is only for tf mods, because pyrovision is only tf thing
+#if defined( TF_CLIENT_DLL )
+	{
+		ConVarRef( "mat_supportflashlight" ).SetValue( false );
```
ficool2, line 1055: This is the wrong way to fix this, fix the issue in the shader rather than disabling flashlights entirely...

celisej567: I agree, but I'm very bad at shaders. And since I saw this way to fix this in the materialsystem, I decided that it would be faster for ordinary people to add these lines to fix it on the spot.

gaelcoral:
What do you mean by "This happens only in mods, and not in default tf"?
Because I also get these artifacts in TF2.
(btw RTX 4060 and Vulkan mode enabled)

https://github.com/user-attachments/assets/60b87f1c-fc80-4a93-9dd5-b1e2200f9787

JoriKos:
Pretty sure this is the true vision of what Pyro sees.

I had some similar issues outside of Pyrovision that kinda looked similar, shown [here](https://www.youtube.com/watch?v=0r0M-EaZH5Q). Not sure if it's related. I've also seen this same artifact happen in someone else's game who doesn't use Vulkan, so probably unrelated to that.

celisej567:
@JoriKos 
It is the same artifacts, but Im not sure if they are caused by the same reasons. Ive tested only dustbowl and 2fort, and both had this issues with pyrovision on. I'm not sure, but looks like in your vid there is no provision. Can you give me:

1. Entire soldier class loadout from this video.
2. Map name.
3. Launch options
4. Screenshot of your Graphics options (advanced as well).
5. List of mods. if you have them ofc.

JoriKos:
1. Community Wiki Cap, Self-Made Cap that Charity Donor medal, Strange Lord Cockswain's Novelty Mutton Chops and Pipe.
2. Rottenburg (`mvm_rottenburg`)
3. -novid -no_texture_stream -nosteamcontroller
4. I use mastercomfig which doesn't reflect accurately in these settings IIRC. It's mastercomfig flat mouse, high preset, null-cancelling movement, and no tutorial. I can provide screenshots if necessary, of course.
5. For mods it's the old Rocket Jumper/Sticky Jumper models, an older version of [NativeHUD](https://gamebanana.com/mods/329612), a 2013 backpack appearance restoration (which I can't find the link for anymore), a skybox texture edited to be entirely black (for HLMV purposes), and hit/killsound.

I've only had this bug in that specific spot, at specific view angles, on that map. I've never seen it happen anywhere else.

celisej567:
I wasnt able to reproduce the bug on that spot, but if(!!!) its the same issue, this fix should work.

JoriKos:
I should note I found some other spots with this issue, so it's not specific to that spot but just one I found it to happen quite consistently for me.

SansGits129:
Hi! Having this issue too but completely vanilla, I use an RTX 5060 Laptop

Jobima1st:
> I had some similar issues outside of Pyrovision that kinda looked similar

I remember see it in Portal 1 game too (Edit here it is https://youtu.be/fdy3WrRDeP0?t=2668)

sealhud:
Same problem on my 2 computers.
On my Acer Nitro laptop, with a GeForce 1050 GPU, and also on my main PC (just formated, no mods, all vanilla) with a RTX 5060 TI 16GB.

8l172:
Same issue. I haven't noticed it outside Pyroland/Pyrovision. AMD Ryzen-7 3700 8-Core CPU & RTX 3070 GPU on Windows 11.

I don't really understand the whole mat_supportflashlight thing that apparently fixes it? I'm not sure what I'm supposed to do with that exactly?

I noticed it on Pyro and Heavy. Verifying game file integrity does nothing to fix it & it persists even with mods completely removed from the game folder, but here they are still:
PeachHud
Heavy First-Person Animation Overhaul (Paysus)
Pyro First-Person Animation Overhaul (Paysus)
Femme Pyro Renovation REVAMP + Revertmod (Tman2)
Cueki's Casual Preloader

PR #1706, opened 2025-12-14
Add IsInDownloadables and AddToDownloadables to VScript

Hex4dec:
Closes https://github.com/ValveSoftware/Source-1-Games/issues/7744

### DIFF
--- a/src/game/server/vscript_server.cpp
+++ b/src/game/server/vscript_server.cpp
@@ -24,6 +24,7 @@
 #include "engine/IEngineSound.h"
 #include "vscript_utils.h"
 #include "netpropmanager.h"
+#include "networkstringtable_gamedll.h"
 #include "client.h"
 #include "tier0/vcrmode.h"
 #include "in_buttons.h"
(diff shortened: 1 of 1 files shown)


ficool2:
Maps should not have the ability to add downloadable content. 
In the vanilla engine you can use [resource lists ](https://developer.valvesoftware.com/wiki/Resource_list_(Source))to create download tables, or you can make a lightweight [VSP plugin](https://developer.valvesoftware.com/wiki/Server_plugins) in C++

Hex4dec:
What about scripts not specific to a map, like scripts that add custom weapons, or map-independent gamemodes?

Bitl:
this is exactly how a security vulnerability starts, i argue

Hex4dec:
Currently the AddToDownloadables function forbids absolute paths and paths that contain "..", which are the same checks as StringToFile. That means that the function can only download resources in a VPK (I think) and resources in the mod's or game's resource directory to the client. (and client-side vulnerabilities don't matter, since native plugins can exploit those anyway)

The IsInDownloadables just checks if the passed string exists in the table, I think this is perfectly safe though.

Hex4dec:
Though, maybe a compromise on security could be made in the same manner as changing ConVars from script with a whitelist system. Though I feel like that's unneeded since you're not gonna find anything confidential in resources anyway.

ficool2:
Regardless such a change is out of scope for VScript, which is intended first and foremost for maps, running it as "plugins" is not the intended usage. You can always extend the API yourself with C++

Hex4dec:
Huh, I wasn't aware that maps are the only purpose for VScript. I thought that VScript was simply a ["virtual machine for scripting that acts as an abstract binding layer between the Source and Source 2 engine and external scripts."](https://developer.valvesoftware.com/wiki/VScript), not explicitly a tool for map logic.

I'll keep this open for the maintainers to decide if this is in the scope of VScript or not, since that seems to be the main point of criticism here.

PR #1715, opened 2025-12-18
Address hitching with GenerateRandomItem

DR-Seal:
Within gamemodes such as Mann vs. Machine, on a listen server, notable hitching occurs when bots spawn in this is due to GenerateRandomItem which does a big lookup to find the item it wants to spawn. This pull request implements GenerateItem which follows the same rules but avoids the lookup by specifying the name directly which avoids that hitching.


HyperionCoding:
Related PR https://github.com/ValveSoftware/source-sdk-2013/pull/887

mastercoms:
Optimized it generically within the same caller:

https://github.com/mastercomfig/tc2/commit/a4a34bf657c223c9ceb42a03b28d8777475c2a0c

PR #1716, opened 2025-12-19
Check if player is alive before showing VM effects

RGBACatlord:
# Description
Fixes effects flooding the screen when spectating a dead teammate in first person with an unusual weapon


bobatealee:
Much better than doing a band-aid fix in particle systems. Thanks for this!

PR #1717, opened 2025-12-20
[TF2] fix caber explosion not being lag-compensated

wgetJane:
# Description

fixes ValveSoftware/Source-1-Games/issues/7761

unfixed, minimum interp time:

https://github.com/user-attachments/assets/2fb988b3-38b4-4055-9c6d-a9ee8426d1b9

unfixed, default interp time:

https://github.com/user-attachments/assets/b04544f6-37b6-4d73-943e-04de32398709

fixed, default interp time:

https://github.com/user-attachments/assets/05b56ed2-2cc3-4194-be3d-67baad76207e

### DIFF
--- a/src/game/shared/tf/tf_weapon_bottle.cpp
+++ b/src/game/shared/tf/tf_weapon_bottle.cpp
@@ -17,6 +17,7 @@
 #include "tf_player.h"
 #include "tf_fx.h"
 #include "tf_gamerules.h"
+#include "ilagcompensationmanager.h"
 #endif
 
 //=============================================================================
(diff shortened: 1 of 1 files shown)


wgetJane:
minor note: this could potentially not unlag all players that the explosion should reach because of the WantsLagCompensationOnEntity filter:

https://github.com/ValveSoftware/source-sdk-2013/blob/7191ecc418e28974de8be3a863eebb16b974a7ef/src/game/server/tf/tf_player.cpp#L20078

the filter includes players that are within a 45 degree cone from the attacker's crosshair, and players that are within a certain distance based on the attacker's max walk speed and the value of the sv_maxunlag cvar:

https://github.com/ValveSoftware/source-sdk-2013/blob/7191ecc418e28974de8be3a863eebb16b974a7ef/src/game/server/tf/tf_player.cpp#L20128-L20130

sv_maxunlag's default value is 1.0 and the caber user's walk speed is going to be 280 unit/s (or 308 with booties)

so under normal circumstances this distance is going to be 420 units (or 462 with booties), which is well above the caber explosion's radius of 100 units

if the caber user's max walk speed is somehow set really low (maybe through vscript idk) or sv_maxunlag is set to a very low value (i think older anticheat plugins do this), then the distance might not be enough to reach every player that should be lagcomped

though again, this is really minor, just something funny to note
(maybe weapons should have a WantsLagCompensationOnEntity method? probably out of scope for this pr though)

PR #1718, opened 2025-12-20
Reduce seam in water refraction

URAKOLOUY5:
This PR adjusts water clipping plane by 4 units up, enough to fix visible seam
**Note:** it leads to x2 performance usage as it renders view above but clips it early

Before:
image

After:
image

### DIFF
--- a/src/game/client/viewrender.cpp
+++ b/src/game/client/viewrender.cpp
@@ -6051,6 +6051,9 @@ void CAboveWaterView::CRefractionView::Draw()
 	g_pPortalRender->WaterRenderingHandler_PreRefraction();
 #endif
 
+	// Move camera 4u down to remove visible refraction gap
+	origin.z -= 4.f;
+
 	// Store off view origin and angles and set the new view
 	int nSaveViewID = CurrentViewID();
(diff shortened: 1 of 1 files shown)


src/game/client/viewrender.cpp
```diff
@@ -6038,7 +6038,7 @@ void CAboveWaterView::CRefractionView::Setup()
 
 	m_DrawFlags = DF_RENDER_REFRACTION | DF_CLIP_Z | 
 		DF_RENDER_UNDERWATER | DF_FUDGE_UP | 
-		DF_DRAW_ENTITITES ;
+		DF_DRAW_ENTITITES | DF_RENDER_ABOVEWATER;
```
ficool2: This leads to a significant performance penalty as geometry above the water will be rendered twice (`CAboveWaterView` setup also adds this flag)

Removing this flag shows that the offset height hack doesn't work in isolation. The flag by itself fixes the issue, but this trade off in performance is not worth it

image

ficool2:
The updated version of this PR leads to noticeable seams where the water intersects geometry

image

It also breaks the height depth blending effect, as noticeable here

Stock:
image

PR version:
image

PR #1719, opened 2025-12-21
[TF2] fix shortstop shove being prevented by reloading

wgetJane:
# Description

fixes ValveSoftware/Source-1-Games/issues/5326

i think that this can reasonably be considered a bug because the shortstop's alt-fire is the only action in the game that can't interrupt a reload, so i highly doubt that this was intentional at all especially since autoreload is enabled by default

this change allows the secondary fire to cancel a reload in the exact same way the primary fire cancels a reload

(i have also tested this change under high latency conditions)

### DIFF
--- a/src/game/shared/tf/tf_weapon_pistol.cpp
+++ b/src/game/shared/tf/tf_weapon_pistol.cpp
@@ -12,6 +12,7 @@
 #ifdef CLIENT_DLL
 #include "c_tf_player.h"
 #include "c_tf_gamestats.h"
+#include "prediction.h"
 // Server specific.
 #else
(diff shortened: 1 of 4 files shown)


wgetJane:
guess i'm blind, looks like there's already an existing pr that i missed: #1420

i'll check that other pr later since there's a bunch of issues i encountered while trying to make this, so that pr would be superior if those issues don't exist since its solution is a lot simpler

bigmazi:
> because the shortstop's alt-fire is the only action in the game that can't interrupt a reload

Together with altfire of Cleaner's Carbine.

PR #1727, opened 2025-12-29
[TF2] Fix oversight preventing see_enemy_health attribute from functioning properly on classes other than Medic and Heavy

AndrewBetson:
# Description
This PR fixes an oversight that prevents the `see_enemy_health`/`mod see enemy health` attribute from functioning when applied to any class other than Heavy or Medic.

Currently, the `CTargetID::UpdateID` function manually checks if the player is either a Heavy or Medic to determine whether to update target ID data or not.

This results in the enemy target IDs having no text when the attribute is applied to classes other than Heavy or Medic.

This PR fixes this issue by also checking `pLocalTFPlayer` for the `see_enemy_health` attribute.

The original manual checks for Heavy or Medic are left in place in case they are necessary for any obscure edge cases I'm not aware of.

Before             |  After
  |

### DIFF
--- a/src/game/client/tf/tf_hud_target_id.cpp
+++ b/src/game/client/tf/tf_hud_target_id.cpp
@@ -799,11 +799,14 @@ void CTargetID::UpdateID( void )
 			bool bMedic = pLocalTFPlayer->IsPlayerClass( TF_CLASS_MEDIC );
 			bool bHeavy = pLocalTFPlayer->IsPlayerClass( TF_CLASS_HEAVYWEAPONS );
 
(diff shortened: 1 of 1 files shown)


FlaminSarge:
This may have some overlap with #993.
In addition, it uses the 'same-team' target where the other PR uses the 'enemy' target (the other PR fixes up Spy behavior to use the 'enemy' target as well).

AndrewBetson:
Thanks for the heads up, it completely slipped my mind to check for any similar pre-existing PRs.

Closing.

PR #1729, opened 2025-12-29
Fix speed boost (Concheror, DA) prediction error when changing weapons

Ashe-tf:
# Description
Fixes a prediction error with the speed boost effect. Tested on a listen server, fakelag 100.

Not sure why this was enforced server-side only.

https://github.com/user-attachments/assets/33fc6543-14c7-4cb9-ad7e-e91969489402

https://github.com/user-attachments/assets/c51bd980-558b-42d9-9e70-93cd92d7c538


Ashe-tf:
Videos without the fix are shown [here](https://github.com/ValveSoftware/Source-1-Games/issues/6080#issuecomment-2282256298)

PR #1730, opened 2025-12-30
all games: Use cl_interp_npcs for non-player NextBot interpolation

copperpixel:
# Description
This PR makes NPC NextBots such as tf_zombie and MvM/Halloween bosses respect `cl_interp_npcs` value for interpolation rather than using base `cl_interp`. Previously this ConVar would've only affected HL1, HL2 and ASW NPCs.

The upside to this change is that it lets you keep lower interp for players while being able to give NPCs a higher value so their animations won't be laggy. Old behavior can be restored by setting `cl_interp_npcs` lower than `cl_interp` which will make it use the base value again.

`cl_interp_npcs` ConVar definition has been also updated to be more inline with `cl_interp`.

Both of the comparison videos below were recorded with `cl_interp_ratio 2;cl_interp 0.03;cl_interp_npcs 0.1`.

Before:

[cl_interp_nb_before.webm](https://github.com/user-attachments/assets/599bd867-6cb6-4d55-9515-083f8c444e45)

After:

[cl_interp_nb_after.webm](https://github.com/user-attachments/assets/a921c039-a9be-46a9-8d51-f454f429df6b)

### DIFF
--- a/src/game/client/c_baseentity.cpp
+++ b/src/game/client/c_baseentity.cpp
@@ -78,7 +78,7 @@ void cc_cl_interp_all_changed( IConVar *pConVar, const char *pOldString, float f
 
 
 static ConVar  cl_extrapolate( "cl_extrapolate", "1", FCVAR_CHEAT, "Enable/disable extrapolation if interpolation history runs out." );
-static ConVar  cl_interp_npcs( "cl_interp_npcs", "0.0", FCVAR_USERINFO, "Interpolate NPC positions starting this many seconds in past (or cl_interp, if greater)" );  
+static ConVar  cl_interp_npcs( "cl_interp_npcs", "0.1", FCVAR_USERINFO | FCVAR_NOT_CONNECTED | FCVAR_ARCHIVE, "Interpolate NPC positions starting this many seconds in past (or cl_interp, if greater)", true, 0.0f, true, 0.5f );
 static ConVar  cl_interp_all( "cl_interp_all", "0", 0, "Disable interpolation list optimizations.", 0, 0, 0, 0, cc_cl_interp_all_changed );
 ConVar  r_drawmodeldecals( "r_drawmodeldecals", "1", FCVAR_ALLOWED_IN_COMPETITIVE );
 extern ConVar	cl_showerror;
@@ -5901,7 +5901,7 @@ static float AdjustInterpolationAmount( C_BaseEntity *pEntity, float baseInterpo
 		{
 			while ( pEntity )
 			{
-				if ( pEntity->IsNPC() )
+				if ( pEntity->IsNPC() || ( !pEntity->IsPlayer() && pEntity->IsNextBot() ) )
 					return minNPCInterpolation;
 
 				pEntity = pEntity->GetMoveParent();


wgetJane:
can you check if it gets lagcomped correctly? set it to 0.5 secs of interp and shoot at a moving target with sv_showlagcompensation set to 1

ficool2:
> can you check if it gets lagcomped correctly? set it to 0.5 secs of interp and shoot at a moving target with sv_showlagcompensation set to it 1

Only players are lag compensated, adding lag comp for them should be another PR

copperpixel:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1730#issuecomment-3701850079

this^. I'm gonna address that later by pulling lag compensation system changes from Alien Swarm.

wgetJane:
> Only players are lag compensated, adding lag comp for them should be another PR

i wasn't aware that skeletons weren't lagcomped (i dont really play halloween stuff much)

if they were made to be lagcomped, afaik the game currently wont take cl_interp_npcs into account

this is from different source engine game (gmod) but i think the same behaviour will apply:

cl_interp 0.5
cl_interp_npcs 0

https://user-images.githubusercontent.com/104243160/168848922-738973e9-409e-4dba-95e8-ee0802ae4346.mp4

cl_interp 0
cl_interp_npcs 0.5

https://user-images.githubusercontent.com/104243160/168850691-8f6ea04c-20f5-411e-8d9b-31265417e400.mp4

copperpixel:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1730#issuecomment-3702358483

Fixing this just requires some refactoring in CLagCompensationManager::StartLagCompensation, replace the compute target time logic with something like this:
```cpp
	// Get true latency
	float flLatency = 0.f;
	INetChannelInfo *nci = engine->GetPlayerNetInfo( player->entindex() );

	if ( nci )
	{
		// add network latency
		flLatency = nci->GetLatency( FLOW_OUTGOING );
	}

	auto lambdaCalcTargetTick = [ & ]( int nLerpTicks )
	{
		// correct is the amout of time we have to correct game time
		float correct = flLatency;

		// add view interpolation latency see C_BaseEntity::GetInterpolationAmount()
		correct += TICKS_TO_TIME( nLerpTicks );

		// check bouns [0,sv_maxunlag]
		correct = clamp( correct, 0.0f, sv_maxunlag.GetFloat() );

		// correct tick send by player 
		int targettick = cmd->tick_count - nLerpTicks;

		// calc difference between tick send by player and our latency based tick
		float deltaTime = correct - TICKS_TO_TIME( gpGlobals->tickcount - targettick );

		if ( fabs( deltaTime ) > 0.2f )
		{
			// difference between cmd time and latency is too big > 200ms, use time correction based on latency
			// DevMsg("StartLagCompensation: delta too big (%.3f)\n", deltaTime );
			targettick = gpGlobals->tickcount - TIME_TO_TICKS( correct );
		}

		return targettick;
	};

	float flBaseTargetTime = TICKS_TO_TIME( lambdaCalcTargetTick( TIME_TO_TICKS( player->m_fLerpTime ) ) );
	float flNpcTargetTime  = TICKS_TO_TIME( lambdaCalcTargetTick( TIME_TO_TICKS( player->m_fNpcLerpTime ) ) );
```
And make it use `flNpcTargetTime` when the compensated entity is an NPC.

CBasePlayer needs to be also extended with `m_fNpcLerpTime` member which should contain the clamped value of the client's `cl_interp_npcs`.

PR #1731, opened 2025-12-31
Fix Revolver reload time exploit

Ashe-tf:
# Description
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/4324

Before:

https://github.com/user-attachments/assets/76ec4953-7bd5-4bfe-8800-0efaeb8bb585

After:

https://github.com/user-attachments/assets/5db6fc2e-eb32-4113-a857-2a826e800499


wgetJane:
i will note that this is most likely going to be controversial if merged, if my assessment of The Spy Main Community opinion on this is correct

PR #1732, opened 2025-12-31
Fix Wrap Assassin's bauble bouncing off from brush entities

Ashe-tf:
# Description
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/4826. Instead of just breaking the bauble when hitting the world, it now breaks if it hits anything that isn't a player.

Tested with doors, players, payload, everything seems to work fine.

https://github.com/user-attachments/assets/f62c81f7-43f6-41d8-a3e8-ccc1b409723a


wgetJane:
have you tried to see if it works fine with the weird props at gullywash 2nd

Ashe-tf:
Yes

https://github.com/user-attachments/assets/5c0ea85b-c6b6-40e3-8b65-748828a74138

PR #1735, opened 2026-01-02
[TF2] Check for MVM instead of Upgrades in mvm logic

MasterOfRespawn:
# Description
This changes two checks for enabled upgrades to checks for MVM.
This removes blue player sentry knockback immunity on maps with upgrades like freaky fair and reduces the amount of time tf_bots are stunned when hit by a sandman ball outside of MVM.


TheShermanTanker:
No wonder sentries felt like they didn't do much against robots in MVM, hadn't known this was a feature all along. You really do learn something new every day.

PR #1736, opened 2026-01-02
[TF2] Update tf_bots to support newer flag types

MasterOfRespawn:
# Description
This adds all currently implemented flag types to `CTFBot::GetFlagToFetch` so tf_bots will try to capture these flag types.
This also updates `CTFBot::GetFlagCaptureZone` to ignore disabled capture zones and return  neutral capture zones.
This makes the tf_bots functional on sd_doomsday and should allow them to play all player destruction maps with a capture zone inside the normal area.


TheBetaM:
The check for disabled capture zones is not sufficient for all maps (like pd_snowville_event). 
Some maps way of disabling the triggers changes `CBaseTrigger::m_bDisabled` but not `CCaptureZone::m_bDisabled`, causing the check to pass for those disabled zones.

PR #1737, opened 2026-01-02
[TF2] check for collision before breaking penetrating huntsman arrows

MasterOfRespawn:
# Description
This fixes https://github.com/ValveSoftware/Source-1-Games/issues/7745.
Adds some checks from other trigger based projectiles to `CTFProjectile_Arrow::ArrowTouch` to prevent processing of bounding box collisions if the arrow did not actually touch the model.

### DIFF
--- a/src/game/server/tf/tf_projectile_arrow.cpp
+++ b/src/game/server/tf/tf_projectile_arrow.cpp
@@ -744,9 +744,6 @@ void CTFProjectile_Arrow::ArrowTouch( CBaseEntity *pOther )
 	bool bShield = pOther->IsCombatItem() && !InSameTeam( pOther );
 	CTFPumpkinBomb *pPumpkinBomb = dynamic_cast< CTFPumpkinBomb * >( pOther );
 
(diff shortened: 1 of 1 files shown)


src/game/server/tf/tf_projectile_arrow.cpp
```diff
@@ -760,6 +757,19 @@ void CTFProjectile_Arrow::ArrowTouch( CBaseEntity *pOther )
...
+		   pOther->IsFuncLOD() ||
+		   pOther->IsBaseProjectile() ||
+		   !ShouldTouchNonWorldSolid( pOther, pTrace ) ||
+		   pOther->IsSolidFlagSet( FSOLID_TRIGGER | FSOLID_VOLUME_CONTENTS ) )
+		 && !pPumpkinBomb && !bShield )
+	{
+		return;
+	}
+
 	CTFMerasmusTrickOrTreatProp *pMerasmusProp = dynamic_cast< CTFMerasmusTrickOrTreatProp* >( pOther );
```
ficool2, line 773: This is unreadable and I think its the wrong approach to solving the issue, the arrow should be do proper collision tests against the entity rather than ignoring it

PR #1733, opened 2026-01-01
[TF2] Fix deflected grenades dealing 85-170 self damage to Pyro

num-get:
# Description
Fix the deflected grenades for using the incorrect self damage value of 85-170 (flamethrower damage value) instead of 38-75 (same as demoman's grenade jump).

Before:

https://github.com/user-attachments/assets/4d200d45-7b3c-4ed6-89b4-817c061d2863

After:

https://github.com/user-attachments/assets/97d599fd-c802-4ac1-abc2-b116e57b0f54

PR #1741, opened 2026-01-04
Reflected Rescue Ranger bolts heal buildings.

muster128:
Reflected Rescue Ranger bolts now heal buildings that are on the same team as the reflector.

# Description

### DIFF
--- a/src/game/server/tf/tf_projectile_arrow.cpp
+++ b/src/game/server/tf/tf_projectile_arrow.cpp
@@ -182,8 +182,6 @@ void CTFProjectile_Arrow::InitArrow( const QAngle &vecAngles, const float fSpeed
 
 	// Save the scoring player.
 	SetScorer( pScorer );
-	//Save the original scoring player
-	SetOriginalScorer(pScorer);
 
 	// Create a trail.
 	CreateTrail();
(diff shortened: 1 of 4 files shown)


wgetJane:
looks good (other than not fitting valve's whitespace style) but i personally think similar reflect bugs should be factored in

see: https://github.com/ValveSoftware/source-sdk-2013/pull/1733#issuecomment-3703419792

im not really sure how to go about this other than a big pr that consolidates related fixes, idk if that's the right approach or not

muster128:
I'm not totally sure either. I was told that PRs should focus on one single topic, and this and the grenade bug might be different enough to qualify as separate issues. Maybe there should be a GetOriginalProjectileCreator() function to CBaseProjectile, which could help with both issues?
Is there a guide on code styling requirements I can look up? This is my first time trying to make TF2 changes like this, so I'm a little out of my depth.

wgetJane:
imo maybe move the definition to the base tf2 projectile class instead of just the arrow projectile

> Is there a guide on code styling requirements I can look up? This is my first time trying to make TF2 changes like this, so I'm a little out of my depth.

there isn't really a style guide, but the rest of code base has a space after an opening parenthesis and a space before a closing parenthesis

so this:
```cpp
if (!pOriginalAttacker)
```

should be like this:
```cpp
if ( !pOriginalAttacker )
```

muster128:
Made a new commit where I expose GetAttackerEntity() as a public member function. Does this work better?

PR #1743, opened 2026-01-04
[All] Fix func_wall `Use` input

SirYodaJedi:
This PR fixes the `Use` input and function for func_wall to toggle between frames 0 and 1, as it did in GoldSrc.

When porting this entity, `pev->frame` was changed to a `m_nState`. Unlike GoldSrc's pev->frame, m_nState doesn't do anything on its own. This PR uses it as it seems to have been intended: to store the on/off state of the entity so that `SetTextureFrameIndex()` can be called accordingly. The Use input has also been updated to properly handle USE_TYPE.

Yes, I am aware that this can be achieved with env_texturetoggle, and that func_wall is deprecated by func_brush. This was an exercise and demonstration in how the underutilized [SetTextureFrameIndex()](https://developer.valvesoftware.com/wiki/SetTextureFrameIndex()) can be used to toggle textures via code.

While this replicates GoldSrc behavior, no HLSDK code was used (as it would not have worked).

### DIFF
--- a/src/game/server/bmodels.cpp
+++ b/src/game/server/bmodels.cpp
@@ -36,6 +36,7 @@ public:
 	bool	CreateVPhysics( void );
 	void	Use( CBaseEntity *pActivator, CBaseEntity *pCaller, USE_TYPE useType, float value );
 
+	// formerly pev->frame
 	int		m_nState;
 };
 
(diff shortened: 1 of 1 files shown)


SirYodaJedi:
converted to draft because the USE_TYPE handling needs to be fixed

SirYodaJedi:
Potential fix for USE_TYPE handling; not sure if I'm doing it correctly.
https://github.com/Yuis-Archives/source-sdk-2025/compare/func_wall_use_fix...Yuis-Archives:source-sdk-2025:func_wall_use_fix_fix

PR #1744, opened 2026-01-05
[TF2] fix miniguns having inconsistent spin-down time based on which mouse button is released

wgetJane:
# Description

when you unrev the minigun by releasing your left mouse button, it takes ~1.15 secs to fully spin down

when you unrev the minigun by releasing your right mouse button, it takes ~1.44 secs to fully spin down

this is because when `SendWeaponAnim( ACT_VM_SECONDARYATTACK )` is used to set the weapon's idle rev animation, it also inadvertently changes the weapon's "idle time" as a side effect which adds a delay:

https://github.com/ValveSoftware/source-sdk-2013/blob/7191ecc418e28974de8be3a863eebb16b974a7ef/src/game/shared/basecombatweapon_shared.cpp#L2425

unfixed:

https://github.com/user-attachments/assets/d57b4547-8c95-421f-9f47-645748293b95

fixed:

https://github.com/user-attachments/assets/8e001858-b9aa-451f-b917-4df95b7c5938

note: green crosshair = no mouse buttons pressed

### DIFF
--- a/src/game/shared/tf/tf_weapon_minigun.cpp
+++ b/src/game/shared/tf/tf_weapon_minigun.cpp
@@ -435,6 +435,12 @@ void CTFMinigun::SharedAttack()
 		}
 	}
 
+	if ( m_iWeaponState == AC_STATE_FIRING || m_iWeaponState == AC_STATE_SPINNING )
+	{
+		// make sure both primary and secondary fire will spin down at identical times
+		SetWeaponIdleTime( gpGlobals->curtime + 0.01 );
+	}
+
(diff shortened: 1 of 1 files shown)


mastercoms:
This breaks Heavy bots (causes them to spin up and spin down constantly). Need to have `me->PressAltFireButton( 0.3f )` in `tf_bot_behavior.cpp`, in the block that says `// if we're a heavy and just saw a bad guy, keep the barrel spinning (unless we're in a hurry)`

PR #1747, opened 2026-01-07
[TF2] fix fire rate being slower than intended for some weapons

wgetJane:
# Description

explained here: https://www.youtube.com/watch?v=7puuYqq_rgw

fixes ValveSoftware/Source-1-Games/issues/5982

this is most likely gonna be contentious since it buffs some weapons as a result, but i wanted to see what ppl think about this anyway

for example:

- buffs MvM firing speed upgrades if they're misaligned with the server's tick interval (default: 0.015 seconds)

- minigun, syringe gun, and smg average attack interval will be 0.1 seconds instead of 0.105 seconds, which slightly buffs their fire rate from 571 RPM to 600 PM

- currently, the pretty boy's pocket pistol has a "+15% faster firing speed" attribute which actually is functionally only +10% rather than +15%, since it fires every 0.135 seconds (444 RPM) instead of the intended 0.1275 seconds (470 RPM) while the stock pistol's attack interval is 0.15 seconds (400 RPM), this pr will fix this

- currently, the cleaner's carbine has a "25% slower firing speed" attribute which is functionally 35%, since it fires every 0.135 seconds (444 RPM) instead of the intended 0.125 seconds (480 RPM), this pr will fix this

- this does not affect flamethrowers and sentryguns

### DIFF
--- a/src/game/shared/tf/tf_weaponbase_gun.h
+++ b/src/game/shared/tf/tf_weaponbase_gun.h
@@ -54,8 +54,7 @@ public:
 	virtual void SecondaryAttack( void );
 	virtual bool Holster( CBaseCombatWeapon *pSwitchingTo );
 
-	virtual float GetCorrectedNextAttackTime( float flFireDelay, float flAttackTime ) const;
-	virtual float GetCorrectedNextAttackTime( float flFireDelay ) const { return GetCorrectedNextAttackTime( flFireDelay, m_flNextPrimaryAttack ); }
+	float GetCorrectedNextAttackTime( float flAttackTime, float flFireDelay ) const;
 
 	// Derived classes call this to fire a bullet.
 	//bool TFBaseGunFire( void );
--- a/src/game/shared/tf/tf_weaponbase_gun.cpp
+++ b/src/game/shared/tf/tf_weaponbase_gun.cpp
@@ -125,7 +125,7 @@ void CTFWeaponBaseGun::PrimaryAttack( void )
 		if ( GetOwner() && GetAmmoPerShot() > GetOwner()->GetAmmoCount( m_iPrimaryAmmoType ) )
 		{
 			WeaponSound( EMPTY );
-			m_flNextPrimaryAttack = GetCorrectedNextAttackTime( flFireDelay );
+			m_flNextPrimaryAttack = GetCorrectedNextAttackTime( m_flNextPrimaryAttack, flFireDelay );
 			return;
 		}
 	}
@@ -181,7 +181,7 @@ void CTFWeaponBaseGun::PrimaryAttack( void )
 	}
 
 	// Set next attack times.
-	m_flNextPrimaryAttack = GetCorrectedNextAttackTime( flFireDelay );
+	m_flNextPrimaryAttack = GetCorrectedNextAttackTime( m_flNextPrimaryAttack, flFireDelay );
 
 	// Don't push out secondary attack, because our secondary fire
 	// systems are all separate from primary fire (sniper zooming, demoman pipebomb detonating, etc)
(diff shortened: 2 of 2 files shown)


Churkinator:
Merging this PR to the game will bring more good than bad in my opinion. I personally don't mind gameplay being slightly changed if the game in the end will be more polished and stable, especially since we probably will never ever see the balance updates anymore. If Valve or other people don't want to see such balance changes, then, in this case, weapons stats can be changed so the weapons can work similarly, if not the same as they were before; I don't think tweaking some numbers here and there is that much of a hard work.

Also, maybe you can look into this too, if you're interested of course. As far as I understand, it touches similar problem with fire rate, but for the Sentry Gun: [#6384](https://github.com/ValveSoftware/Source-1-Games/issues/6384)

wgetJane:
note: this doesn't affect the flamethrower, which has an attack interval of 0.02 seconds that gets rounded up to 0.03 seconds

https://github.com/ValveSoftware/source-sdk-2013/blob/7191ecc418e28974de8be3a863eebb16b974a7ef/src/game/shared/tf/tf_weapon_flamethrower.cpp#L68

i didnt bother fixing this because it would be a massive +50% increase in fire rate instead of the tiny single-digit percent increases for the other guns

src/game/shared/tf/tf_weaponbase_gun.h
```diff
@@ -54,6 +54,9 @@ class CTFWeaponBaseGun : public CTFWeaponBase
 	virtual void SecondaryAttack( void );
 	virtual bool Holster( CBaseCombatWeapon *pSwitchingTo );
 
+	virtual float GetCorrectedNextAttackTime( float flFireDelay, float flAttackTime ) const;
+	virtual float GetCorrectedNextAttackTime( float flFireDelay ) const { return GetCorrectedNextAttackTime( flFireDelay, m_flNextPrimaryAttack ); }
```
ficool2: These functions shouldn't be virtual unless it will actually be overriden by something.

I also don't see the point in the overload existing, being explicit about what attack timer is being corrected is more clear

In addition it makes more sense to move this into basecombatweapon as it affects all games

wgetJane: i keep it to tf2 since i remember that's also cs:s and csgo's approach, iirc they have a similar function that's only in the base cs gun class

wgetJane: how about moving this helper function to basecombatweapon and then just using it within tf2?

Brain-dawg:
~~Would cause a noticeable buff to sentry firing speed upgrades in MvM, the 3rd tick in particular is notoriously completely non-functional.  Valve dropped the vaccinator uber duration bugfix in a random update though, arguably a far more significant bug/balance change than this would be.~~

I cannot read and this comment is irrelevant.

wgetJane:
> Would cause a noticeable buff to sentry firing speed upgrades in MvM, the 3rd tick in particular is notoriously completely non-functional. Valve dropped the vaccinator uber duration bugfix in a random update though, arguably a far more significant bug/balance change than this would be.

this pr doesn't affect sentry guns (though i forgot to make this clear)

BreavyTF2:
Doesn't predict correctly causing lower firerate at higher ping, tested with net_fakelag 100.

wgetJane:
> Doesn't predict correctly causing lower firerate at higher ping, tested with net_fakelag 100.

thanks, i'll test again later

wgetJane:
@BreavyTF2 after some testing, i've concluded that that issue happens in live tf2 as well, and only in your own listen server

this pr does not cause that issue and it's extremely minor anyway since the only way to get high ping to your own listen server is with the net_fakelag command

Primary-Ad2848:
Wishing this will get merged, it will be very good for game.

Ddavs03:
I really liked testing the Cleaner’s Carbine with its correct firing interval, it’s the weapon that benefits the most from this change, even if it still isn’t particularly strong.

FlaminSarge:
Am I correct in saying that this effectively sets the next attack time one tick earlier if it calculates that it's lagging behind?

wgetJane:
> Am I correct in saying that this effectively sets the next attack time one tick earlier if it calculates that it's lagging behind?

yes

PR #1748, opened 2026-01-07
[TF2] fix being unable to detonate stickybombs while switching weapons or attacking with melee

wgetJane:
# Description

fixes ValveSoftware/Source-1-Games/issues/2185

this fixes two sticky det delay bugs:

1\) not being able to detonate stickies while swinging your melee

2\) if you switch weapons (to either primary or melee) while holding down right-click, your stickies will not detonate until your primary/melee is fully deployed

the 2nd bug was supposed to have been patched by the [December 13, 2017 Patch](https://wiki.teamfortress.com/wiki/December_13,_2017_Patch), however the patch doesn't actually work as you can see in the video below
> Fixed not being able to detonate stickybombs when switching to melee with secondary attack held down

unfixed:

https://github.com/user-attachments/assets/f6a5881f-3dea-4e26-8a91-d02b69dd4f54

fixed:

https://github.com/user-attachments/assets/3f3b160d-611b-45b1-8da4-4fb9e2bd98d9

crosshair colour note for these videos:
green = no buttons held
red = only left-click held
cyan = only right-click held
magenta = both buttons held

### DIFF
--- a/src/game/shared/tf/tf_weaponbase.cpp
+++ b/src/game/shared/tf/tf_weaponbase.cpp
@@ -2366,14 +2366,27 @@ void CTFWeaponBase::ItemBusyFrame( void )
 		return;
 	}
 
+	bool bDoSpecialSkill = false;
+
 	if ( ( pOwner->m_nButtons & IN_ATTACK2 ) && /*m_bInReload == false &&*/ m_bInAttack2 == false )
 	{
-		pOwner->DoClassSpecialSkill();
+		bDoSpecialSkill = true;
 		m_bInAttack2 = true;
 	}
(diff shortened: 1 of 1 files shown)


wgetJane:
fixes #ValveSoftware/Source-1-Games/issues/2185

old702:
is there a similar bug with detonator flare?

wgetJane:
> is there a similar bug with detonator flare?

the detonator flares cannot be detonated at all after you switch weapons, which i think is probably intentional because pyro uses alt-fire for airblast and the detonator flares last for a really short time compared to stickybombs

it is also handled differently so it's out of scope for this pr

PR #1749, opened 2026-01-08
[TF2] fix Fists of Steel overheal penalty getting bypassed by switching weapons

wgetJane:
# Description

bug: the `mult_patient_overheal_penalty_active` attribute on the Fists of Steel is taken into account only when the medigun starts healing the patient, rather than while the weapon is active

this bug allows players to easily bypass an intended downside of the weapon, which makes this fix technically a balance change, but i believe that the majority of players, even heavy mains, will agree that this fix is warranted

fixes ValveSoftware/Source-1-Games/issues/7356

unfixed:

https://github.com/user-attachments/assets/9dc8318a-f6f3-4455-b05b-70b3d939a829

fixed:

https://github.com/user-attachments/assets/720a0131-9b63-457a-99a5-e2b111f71e46


JoriKos:
For what it's worth, I don't believe it's a balance change if it fixes the weapon's functionality, but rather just making the weapon work as intended.

PR #1750, opened 2026-01-08
[TF2] fix resupply lockers not resetting critheal timer when used

wgetJane:
# Description

bug/oversight: resupply lockers don't reset your critheal timer for some reason even though it resets every other ailment on the player, which gives players with a resupply bind (aka the "b4nny bind") even more of an advantage


palmtopangie:
Valve merge this PR and my life is yours

PR #1751, opened 2026-01-08
[TF2] disable "spawn wallhacks" for disguised enemy spies and visible teammates

wgetJane:
# Description

the `tf_enable_glows_after_respawn` feature is intended to guide players towards their teammates, which helps keep the game more focused since it subtly encourages everyone to play closer together and prevents people from getting lost in an unfamiliar map

however, the way this interacts with spy is extremely controversial, which is why the community has dubbed it as "spawn wallhacks"

when you respawn you get 10 seconds where you're just able to see any disguised spies through walls at any distance, and they instantly give themselves away by cloaking, decloaking, undisguising, changing disguise, or even just by looking suspicious or out of position which is really unfair especially when they can't see you through walls like you can

a common suggestion is to just disable the outlines for enemy spies, however this is usually countered with the fact that it would make disguised enemy spies immediately obvious because you'll see their lack of outline

my solution to that is to simply not draw outlines for all teammates that are already visible, which also has the added benefit that it saves performance and limits visual clutter since it's kinda pointless to highlight unoccluded teammates anyway

this won't completely eliminate the ability to catch spies using the respawn glow effect, but i think this is the best compromise available

demonstration:

https://github.com/user-attachments/assets/4c73acbc-9253-4cee-b7ef-caa676f8fad6

closes ValveSoftware/Source-1-Games/issues/4361

### DIFF
--- a/src/game/client/tf/tf_hud_spectator_extras.cpp
+++ b/src/game/client/tf/tf_hud_spectator_extras.cpp
@@ -119,9 +119,12 @@ void CTFHudSpectatorExtras::OnTick()
 		return;
 	}
 
+	bool bLocalPlayerIsAlive = pLocalPlayer->IsAlive();
+	bool bLocalPlayerTeamGlows = bLocalPlayerIsAlive && pLocalPlayer->m_Shared.InCond( TF_COND_TEAM_GLOWS );
+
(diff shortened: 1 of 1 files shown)


seanmcgeehan:
What a wonderful change

PR #1753, opened 2026-01-09
[TF2] fix explosions being blocked by small protrusions from the ground's geometry (splash bug/stair bug)

wgetJane:
# Description

fixes ValveSoftware/Source-1-Games/issues/2486
fixes ValveSoftware/Source-1-Games/issues/4556

what this pr does is it performs a second line-of-sight check from a little bit above where the explosion is
(the offset used is the player stair step height that maps are built for)

image

this only works if the explosion is close to the ground specifically so that this fix won't affect the game's balance,
because otherwise soldiers would be able to deal splash damage against targets on higher ground by simply shooting at the edge of the ledge from below, which would let soldiers be able to deny high ground extremely easily

image

note:
another pr exists with an alternative solution: #1273

however, i think its solution (and other similar solutions) would be a pretty significant buff for the soldier and demo's ability to defend chokes such as doorways, which they're already extremely good at

image

it's really only splashbugs on the ground that players actually complain about, so i don't think it's worth changing splash behaviour around corners horizontally since it would just open a big can of worms

-----

unfixed:

https://github.com/user-attachments/assets/7b208f97-aff3-40f2-b296-f0ca6acb794d

fixed:

https://github.com/user-attachments/assets/1a63cebf-9fdb-4f3d-8681-8d044098e45d

### DIFF
--- a/src/game/shared/tf/tf_gamerules.cpp
+++ b/src/game/shared/tf/tf_gamerules.cpp
@@ -5753,6 +5753,7 @@ void CTFRadiusDamageInfo::CalculateFalloff( void )
 // Purpose: Attempt to apply the radius damage to the specified entity
 //-----------------------------------------------------------------------------
 ConVar tf_radiusdamage_los_simple( "tf_radiusdamage_los_simple", "0", FCVAR_NONE, "Use simpler line-of-sight checks for radius damage" );
+ConVar tf_radiusdamage_los_debug( "tf_radiusdamage_los_debug", "0", FCVAR_CHEAT );
 int CTFRadiusDamageInfo::ApplyToEntity( CBaseEntity *pEntity )
 {
 	if ( pEntity == pEntityIgnore || pEntity->m_takedamage == DAMAGE_NO )
@@ -5781,6 +5782,14 @@ int CTFRadiusDamageInfo::ApplyToEntity( CBaseEntity *pEntity )
 		UTIL_TraceLine( vecSrc, vecSpot, MASK_RADIUS_DAMAGE, &filterSelf, &tr );
 	}
 
+	bool bDebugLoS = tf_radiusdamage_los_debug.GetBool() && ( pEntity->IsPlayer() || tf_radiusdamage_los_debug.GetInt() == 2 );
+
+	if ( bDebugLoS )
+	{
+		NDebugOverlay::Line( tr.startpos, tr.endpos, 255, 0, 0, true, 3 );
+		NDebugOverlay::Line( tr.endpos, vecSpot, 96, 0, 0, true, 3 );
+	}
+
 	// If we don't trace the whole way to the target, and we didn't hit the target entity, we're blocked
 	if ( tr.fraction != 1.f && tr.m_pEnt != pEntity )
 	{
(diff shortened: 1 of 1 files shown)


src/game/shared/tf/tf_gamerules.cpp
```diff
@@ -5781,9 +5781,6 @@ int CTFRadiusDamageInfo::ApplyToEntity( CBaseEntity *pEntity )
 		UTIL_TraceLine( vecSrc, vecSpot, MASK_RADIUS_DAMAGE, &filterSelf, &tr );
 	}
 
-	NDebugOverlay::Line( tr.startpos, tr.endpos, 255, 0, 0, true, 3 );
```
ficool2: Keep these as a convar

wgetJane: alright, would be useful i guess

Ashe-tf:
Could you test with more complex surfaces, like round pillars?

wgetJane:
> Could you test with more complex surfaces, like round pillars?

i've tested with various geometry across multiple maps in typical splashbug spots

round pillars wouldn't do anything special, again this just performs a 2nd LoS check from a raised position from the ground

the fix from this pr does NOT let explosions go around corners, and this is a deliberate choice because that would otherwise buff soldier and demoman

the entire point of this fix is solely just to stop small steps on the ground from blocking explosion damage, nothing else

JoriKos:
Would this affect Sentry Busters and ducking behind certain geometry? Most common example is the crate on Mannhattan next to the big tank, with the Sentry Buster being in front of the crate and the Engi behind the crate.

wgetJane:
> Would this affect Sentry Busters and ducking behind certain geometry? Most common example is the crate on Mannhattan next to the big tank, with the Sentry Buster being in front of the crate and the Engi behind the crate.


no, sentry busters don't use this kind of radius damage:

https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/game/server/tf/bot/behavior/missions/tf_bot_mission_suicide_bomber.cpp#L341


also again, this pr only affects explosions that happen less than 18 units from the ground and the new LoS check will not go over obstacles that are more than 18 units tall

you can tell if something is less than 18 units tall in tf2 because you will automatically step onto them without having to jump

if you have to jump over something, then it is not affected by this pr

mastercoms:
I still do feel like wood beams on walls and other errant geometry does block explosions at lot of times, and those are still bad cases.

wgetJane:
> I still do feel like wood beams on walls and other errant geometry does block explosions at lot of times, and those are still bad cases.

i think splashbugs on walls is a lot less common vs splashbugs on the ground, and im not really sure of a clean way to fix them without significantly affecting balance

the same trick of checking how far from the surface the explosion is could be used but it's going to be more complicated, you'd need to check more directions than just downwards

stuff sticking out of walls like wood beams don't seem to have a standard max length unlike stair steps, and mapper intention kind of feels a lot less clear about what shouldn't be causing splashbugs

here's a bunch i could find in cp_process:










i have a feeling that different people would disagree on which of these should count or not, im not really sure about the general community opinion about non-floor splashbugs

PR #1761, opened 2026-01-12
Fix annoying Eureka Effect + 2-way teleporter interaction

kvog-git:
Hi. This PR adds a short cooldown that blocks the player from taking their own teleporter after warping to an exit. Previously, teleporting to an exit with the "2-Way Teleporters" upgrade would immediately teleport you back to the entrance.

The cooldown is 3 seconds, matching the recharge time of a level 3 teleporter.

I tested to ensure this does not break demos recorded on the master branch.

Fixes: https://github.com/ValveSoftware/Source-1-Games/issues/7217

https://github.com/user-attachments/assets/157db9e0-1a4b-4593-85ac-1bc03f13376d

### DIFF
--- a/src/game/server/tf/tf_obj_teleporter.cpp
+++ b/src/game/server/tf/tf_obj_teleporter.cpp
@@ -226,6 +226,8 @@ CObjectTeleporter::CObjectTeleporter()
 	m_flCurrentRechargeDuration = 0.0f;
 	m_flRechargeTime = 0.0f;
 
+	m_flEurekaEffectCooldown = 0.0f;
+
 	ListenForGameEvent( "player_spawn" );
 	ListenForGameEvent( "player_team" );
 }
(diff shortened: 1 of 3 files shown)


wgetJane:
why not put the condition on the teleporter instead of the player?

kvog-git:
> why not put the condition on the teleporter instead of the player?

That's just what came to mind first, but it would be cleaner if I moved it to the teleporter. I'll force-push later.

src/game/server/tf/tf_obj_teleporter.cpp
```diff
@@ -1537,4 +1543,9 @@ void CObjectTeleporter::FireGameEvent( IGameEvent *event )
 			SetTeleportingPlayer( NULL );
 		}
 	}
+}
+
+void CObjectTeleporter::AddEurekaEffectCooldown()
+{
+	m_flEurekaEffectCooldown = gpGlobals->curtime + g_iTeleporterRechargeTimes[3];
```
HyperionCoding, line 1550: Use GetUpgradeLevel()

kvog-git: The cooldown should be constant and not related to teleporter level.

PR #1768, opened 2026-01-15
[TF2] Fixed teleporter recharge bar not working when teleporting from exit

rabscootle:
This PR fixes the teleporter recharge bar not properly displaying incorrect progress when teleporting from exit (2 way teleporter)

https://github.com/user-attachments/assets/dffe7674-dad7-4fe3-b235-2f9323408d3f


rabscootle:
Issue is already fixed in live tf2 (oops)

PR #1769, opened 2026-01-16
[TF2] Set the launcher of Sentry Rockets to the sentrygun

ChaseTownsend:
Sentry rockets never get this set, so it's almost impossible to trace a rocket back to the owner after it's deflected.

Note that this still does not fix sentry rockets not sending a `object_deflected` game event


GrubPL:
mind if I ask, why exactly was it closed?

ChaseTownsend:
I realized that I did not have to pass in another parameter, since it's already passed in the sentrygun

PR #1773, opened 2026-01-18
[TF2] improve medigun heal target selection to prioritize teammate directly under crosshair when teammates are close together

wgetJane:
# Description

this pr solves a minor annoyance with medic where teammates would block you from being able to heal someone that's close to them

this issue is caused by how the medigun traces for player collision hulls for leniency instead of the more accurate player hitboxes

the solution in this pr is to simply trace a second time for player hitboxes and prioritise the result when applicable

before:

https://github.com/user-attachments/assets/962db943-84dd-45de-b17a-7369553e927a

after:

https://github.com/user-attachments/assets/57c5fa97-05f2-4ca3-a971-a67b62e578e9

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -7002,6 +7002,15 @@ void C_TFPlayer::UpdateIDTarget()
 
 	if ( tr.m_pEnt && tr.m_pEnt->IsPlayer() )
 	{
+		trace_t trShot;
+		// use the shot mask to replicate the medigun's trace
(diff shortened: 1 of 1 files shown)


FlaminSarge:
I don't know how viable this would be, but if currently healing a target, quickly retargeting (either letting go of and then quickly repressing +attack, or pressing +attack again if the 'medigun continues healing' setting is set) should potentially switch to the next valid target, away from the current target.

That's probably a separate feature PR, though.

Primary-Ad2848:
as a medic Main, It  would remove lots of infuriating moment from the game for me and would make it a overall better experience. Hoping they will merge this.

Primary-Ad2848:
Congratulations! Thank for your contribution for TF2! and thanks for Eric Smith for caring about this game!

PR #1778, opened 2026-01-24
[TF2] Remove the hard-coded F4 ready up bind for mp_tournament matches

### LINKED ISSUE #7742: [TF2] Replace hardcoded function keys on with console commands
Currently, to ready-up in competitive and MvM, the game is hardcoded to use KEY_F4.
This raises the following issues:
- Modern small keyboard layouts without the function key row will need a workaround.
- Lack of user customisation or conflicting keybinds when F4 is bound by the user.

This also applies generally with the voting system where KEY_F1-KEY_F5 are hardcoded.

Instead, TF2 should have console command for these actions, bounded to the same function keys by default, therefore allowing more customisation.

dilbertron2:
## **The Issue**
An issue with readying up in matches that utilise `mp_tournament 1` (including Mann VS Machine) is that F4 is permanently bound to the `player_ready_toggle` command even if you bind another key to that action or unbind F4 via the console.

Looking at the code shows that `player_ready_toggle` simply simulates an F4 key press and feeds it into the `CHudTournamentSetup::ToggleState` function.

## **The Solution**
Take the code from the `CHudTournamentSetup::ToggleState` function that handles readying up and put it into a new function called `CHudTournamentSetup::TogglePlayerReadiness`. Now update the `player_ready_toggle` command to call `TogglePlayerReadiness`.

This is the first part of the fix done but now we need to handle the text (Which is currently hard-coded to display F4 as the specified key.)

Update the `#Tournament_Instructions` localisation string from `F4 = change team name/status` to `%s1 = change team name/status`, and change `#Tournament_Instructions_Ready` from `F4 = Toggle Ready` to `%s1 = Toggle Ready`.

Now in all applicable places in `tf_hud_tournament.cpp` we look for the key that is bound to `player_ready_toggle` and insert it into the localisation string! This initially caused truncation in the ready up message for Mann VS Machine so I also increased the width of `TournamentInstructionsLabel` and `TournamentInstructionsLabelShadow` from 190 to 300 to ensure enough room, under the `if_mvm` section in `Resource/UI/HudTournament.res`

This also requires the `xpos` of both UI elements to be edited aswell to line up with the new width. `TournamentInstructionsLabel` gets its `xpos` set to c-150 and `TournamentInstructionsLabelShadow` gets its `xpos` set to c-149

# **Video**
https://youtu.be/kE7cYbmGdeM <-- PR in action

## **Possible Issue**
I am unaware if there is a way to dynamically center text based on length, if there is then I either missed it or cant understand it. Because of this the MvM ready up text wont always be centered(?)

## Extra Steps for Valve
All other localisation files will need to be adjusted. `Resource/UI/HudTournament.res` will need to be modified. `config_default.cfg` will need to have `bind "F4" "player_ready_toggle"` added.

## Files
Here are the modified .res and tf_english localisation file :)
[Dynamic Ready Up Extra Files.zip](https://github.com/user-attachments/files/24837584/Dynamic.Ready.Up.Extra.Files.zip) (OLD VERSION)
[Dynamic Ready Up Extra Files 2.zip](https://github.com/user-attachments/files/24846621/Dynamic.Ready.Up.Extra.Files.2.zip) (CURRENT VERSION)

Partially fixes [ValveSoftware/Source-1-Games #7742](https://github.com/ValveSoftware/Source-1-Games/issues/7742)

### DIFF
--- a/src/game/client/tf/tf_hud_tournament.cpp
+++ b/src/game/client/tf/tf_hud_tournament.cpp
@@ -311,7 +311,7 @@ void CHudTournament::PreparePanel( void )
 					}
 
 					g_pVGuiLocalize->ConvertANSIToUnicode(key, wszKey, sizeof(wszKey));
-					g_pVGuiLocalize->ConstructString_safe(wszLocalized, g_pVGuiLocalize->Find("Tournament_Instructions_Ready"), 1, wszKey);
+					g_pVGuiLocalize->ConstructString_safe(wszLocalized, g_pVGuiLocalize->Find( "Tournament_Instructions_Ready" ), 1, wszKey);
 					SetDialogVariable("readylabel", wszLocalized);
 				}
 				else
@@ -319,7 +319,6 @@ void CHudTournament::PreparePanel( void )
 					SetDialogVariable("readylabel", g_pVGuiLocalize->Find(pszLabelText));
 				}
 
-				// SetDialogVariable( "readylabel", g_pVGuiLocalize->Find( pszLabelText ) );
 				SetDialogVariable( "tournamentstatelabel", g_pVGuiLocalize->Find( "Tournament_WaitingForTeam" ) );
 				SetPlayerPanelsVisible( true );
 				m_pModeImage->SetVisible( m_bCompetitiveMode );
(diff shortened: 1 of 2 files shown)


Squid-Eevee:
you could center the label by setting it's xpos to `cs-0.5` or `c-<width/2>`
you can then offset the drop shadow by using `pin_to_sibling`

FlaminSarge:
I'm not sure if string comparison with == works the way you intend here, but it'd probably also be better to either use a separate boolean for that case, or to just do that case's logic inside the block where the 'ready' text is set.

PR #1779, opened 2026-01-27
[TF2] fix Jarate and Mad Milk splash AoE using a cube rather than a sphere

### LINKED ISSUE #5905: [TF2] spheres are not spheres
image

the source engine's `UTIL_EntitiesInSphere` function that is often used by tf2 actually uses a cube rather than a sphere

examples of mechanics that are affected by this:

- jarate and mad milk: https://youtube.com/watch?v=1hYZbyUK170

- airblast: https://github.com/ValveSoftware/Source-1-Games/issues/5522#issuecomment-2074381306

- short circuit

- dispenser

- explosive headshot

- deflector minigun

- halloween spells

- etc

it is unknown if this is intentional or not for any of these mechanics

wgetJane:
# Description

partially fixes ValveSoftware/Source-1-Games/issues/5905

[https://youtube.com/watch?v=1hYZbyUK170](https://youtube.com/watch?v=1hYZbyUK170)

i will note that the consequence of this fix is definitely a noticeable balance change, however the affected weapons generally seem to be deemed to be overpowered by the community anyway

maybe this fix might need to be disabled specifically for mvm? idk what the opinion is for this

video:

https://github.com/user-attachments/assets/fe44f5ba-b427-4158-98b8-0d8f9698292e

### DIFF
--- a/src/game/shared/tf/tf_weapon_jar.cpp
+++ b/src/game/shared/tf/tf_weapon_jar.cpp
@@ -340,6 +340,16 @@ void JarExplode( int iEntIndex, CTFPlayer *pAttacker, CBaseEntity *pOriginalWeap
 			if ( !pPlayer->IsAlive() )
 				continue;
 
+			// UTIL_EntitiesInSphere actually uses a cube rather than a sphere
+			CCollisionProperty *pCollisionProp = pPlayer->CollisionProp();
+			if ( pCollisionProp )
+			{
+				Vector vecNearest;
+				pCollisionProp->CalcNearestPoint( vContactPoint, &vecNearest );
+				if ( vContactPoint.DistToSqr( vecNearest ) > flRadius * flRadius )
+					continue;
+			}
+
 			// Do a quick trace to see if there's any geometry in the way.
 			// Pee isn't stopped by other entities. Splishy splashy.
 			trace_t trace;


src/game/shared/tf/tf_weapon_jar.cpp
```diff
@@ -340,6 +340,16 @@ void JarExplode( int iEntIndex, CTFPlayer *pAttacker, CBaseEntity *pOriginalWeap
 			if ( !pPlayer->IsAlive() )
 				continue;
 
+			// UTIL_EntitiesInSphere actually uses a cube rather than a sphere
```
ficool2, line 343: Rather than having to hack this in at every place where UTIL_EntitiesInSphere is called, this should just be done in `UTIL_EntitiesInSphere` itself

wgetJane: i agree but this is supposed to be a bit more of a conservative approach for now since this would affect quite a lot of other things (airblast, dispensers, etc)

old702:
Maybe, to preserve the interaction volume so it doesn’t feel like a balance change, a coefficient could be applied that keeps the sphere’s volume equivalent to the box’s volume.
Then this could be applied to airblast, dispensers, etc.

Something like:
r(new) = r × 1.24 (so the volume roughly matches the old AABB).

I think that because of the quirks of not rotating boxes, this could also be applied to airblast, which would reduce the blind spots in the cone check logic that occur due to alignment with world axes.

src/game/shared/tf/tf_weapon_jar.cpp
```diff
@@ -340,6 +340,16 @@ void JarExplode( int iEntIndex, CTFPlayer *pAttacker, CBaseEntity *pOriginalWeap
...
+			CCollisionProperty *pCollisionProp = pPlayer->CollisionProp();
+			if ( pCollisionProp )
```
mastercoms, line 345: When will collision prop be null? Do we want to just skip the entity in that case?

PR #1781, opened 2026-01-28
[ALL] Fix color picker frame not regenerating the preview texture after being closed

SanyaSho:
# Description
When CProceduralTexturePanel calls it's CleanUp function it [decrements](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/vgui2/matsys_controls/proceduraltexturepanel.cpp#L98) the current texture handle ref index value and since the texture name never[[1]](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/vgui2/matsys_controls/colorpickerpanel.cpp#L718-L719)[[2]](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/vgui2/matsys_controls/proceduraltexturepanel.cpp#L54) gets changed you're left with a invalid texture until you restart the game

This fix will make each texture created by CProceduralTexturePanel have it's own unique ID, it will also make sure that allocated TextureID is cleared and simplify the texture creation code.

Video demonstration of a bug:

https://github.com/user-attachments/assets/81eb548b-ac54-49be-82a2-86c5bd87154c

Fixed version:

https://github.com/user-attachments/assets/0d75c893-26e9-43e5-8dee-02215afe25c6

### DIFF
--- a/src/vgui2/matsys_controls/proceduraltexturepanel.cpp
+++ b/src/vgui2/matsys_controls/proceduraltexturepanel.cpp
@@ -66,7 +66,7 @@ bool CProceduralTexturePanel::Init( int nWidth, int nHeight, bool bAllocateImage
 	pVMTKeyValues->SetString( "$basetexture", pTemp );
 	pVMTKeyValues->SetInt( "$nocull", 1 );
 	pVMTKeyValues->SetInt( "$nodebug", 1 );
-	m_ProceduralMaterial.Init( MaterialSystem()->CreateMaterial( pTemp, pVMTKeyValues ));
+	m_ProceduralMaterial.Init( pTemp, pVMTKeyValues );
 
 	m_nTextureID = MatSystemSurface()->CreateNewTextureID( false );
 	MatSystemSurface()->DrawSetTextureMaterial( m_nTextureID, m_ProceduralMaterial );


robotboy655:
So, what I am seeing here are material and texture leaks. The panel creates a new texture and material with the exact same name, even if one already exists and never destroys them. (At least in my tests with GMod)
The PR would (kind of) fix the texture leak, but not the material leak.

Since they are procedural textures and materials, they will always have a refcount of at least 1 (`ITextureInternal::CreateProceduralTexture` and `CMaterialSystem::CreateMaterial` add one) and therefore never be freed.

This panel also creates a new TextureID every time the panel is created, but never destroys it. (`CreateNewTextureID`)

SanyaSho:
> So, what I am seeing here are material and texture leaks. The panel creates a new texture and material with the exact same name, even if one already exists and never destroys them. (At least in my tests with GMod) The PR would (kind of) fix the texture leak, but not the material leak.
> 
> Since they are procedural textures and materials, they will always have a refcount of at least 1 (`ITextureInternal::CreateProceduralTexture` and `CMaterialSystem::CreateMaterial` add one) and therefore never be freed.
> 
> This panel also creates a new TextureID every time the panel is created, but never destroys it. (`CreateNewTextureID`)

Thank you for the review! I'll fix the TextureID leak soon.

I don't actually see how I can fix the material leak without replacing the CMaterialReference with a raw pointer or adding a new function that will not increment the refcount like CTextureReference::InitProceduralTexture.

src/vgui2/matsys_controls/proceduraltexturepanel.cpp
```diff
@@ -53,10 +53,22 @@ bool CProceduralTexturePanel::Init( int nWidth, int nHeight, bool bAllocateImage
...
-			m_nWidth, m_nHeight, IMAGE_FORMAT_BGRX8888, 
-			TEXTUREFLAGS_CLAMPS | TEXTUREFLAGS_CLAMPT | TEXTUREFLAGS_NOMIP | 
-			TEXTUREFLAGS_NOLOD | TEXTUREFLAGS_PROCEDURAL | TEXTUREFLAGS_SINGLECOPY );
+	ITexture *pTex = NULL;
+	if ( MaterialSystem()->IsTextureLoaded( pTemp ) )
+	{
+		pTex = MaterialSystem()->FindTexture( pTemp, TEXTURE_GROUP_VGUI );
+		Assert( pTex );
+
+		pTex->AddRef();
```
robotboy655: Just to let you know, this `AddRef` call is unnecessary because `CTextureReference` (`m_ProceduralTexture`) handles its own refcounting, so what this call does is just constantly increase the refcount of the texture and never decrease it.

The texture will never reach refcount of 0 either way because nothing removes the one refcount added by `CreateProceduralTexture`. I think that's fine since its only used in `-tools` anyway as far I know.

(The example you copied this from does not use `CTextureReference`)

SanyaSho: I could technically replace this mess with a static serial for each texture/material (as I wrote in PR description) and call CTextureReference::InitProceduralTexture instead

~~But the material (CMaterialReference) will still have the refcount incremented, even if texture does not, but i guess it's fine since there's alot of materials that aren't getting decremented correctly even inside of the engine~~
No more CMaterial leaks happening for me after the last commit

> (The example you copied this from does not use CTextureReference)

Yep, you're right. I rechecked the code when I saw your review

PR #1784, opened 2026-01-31
[TF2] Add queue button to contracts

Pandaptable:
This PR makes the flow of activating contracts and queuing for a match a little nicer. Players no longer have to activate a contract, deselect all maps, search for that one specific map they need, and then finally queue. Instead there is a "Start Search" button in place of the "Activate" button.

image

The button only shows if the contract is not fully finished and if the contract is active. If required it will automatically select the required maps/modes and then start the casual queue, otherwise it will simply start the queue with the current settings without modifying them. If no maps are selected at all a fallback will load the players saved maps which internally falls back to the entire core category if there are no saved maps.

I also added handling for the usual queue restrictions such as bans, not being party leader, etc. However this includes an ugly hack because `CTFPartyClient::BCanQueueForMatch` also returns an error if there are no maps selected, which for us isn't an issue since we will automatically select at least one map anyways. The only way to check for this case is to manually compare the translation strings, which works fine but isn't nice.

The following button would have to be added into `resource/ui/econ/questviewsubpanel.res`

```
"JoinQueueButton"
{
	"ControlName"	"CExButton"
	"fieldName"		"JoinQueueButton"
	"xpos"			"3"
	"ypos"			"26"
	"zpos"			"10"
	"wide"			"80"
	"tall"			"16"
	"autoResize"	"0"
	"pinCorner"		"0"
	"visible"		"1"
	"enabled"		"1"
	"tabPosition"	"0"
	"labeltext"		"#TF_Matchmaking_StartSearch"
	"textAlignment"	"center"
	"dulltext"		"0"
	"brighttext"	"0"
	"default"		"0"
	"sound_depressed"	"UI/buttonclick.wav"
	"sound_released"	"UI/buttonclickrelease.wav"
	"Command"		"join_queue"
	"proportionaltoparent" "1"
	"actionsignallevel"	"2"
	"textinsety"	"0"
	"eatmouseinput"	"0"	
	"font"			"QuestMap_Large"
	"roundedcorners"	"0"

	"paintbackground"	"1"

	"defaultBgColor_override"	"StoreGreen"
	"armedBgColor_override"		"CreditsGreen"
	"depressedBgColor_override" "CreditsGreen"

	"defaultFgColor_override"	"TanLight"
	"armedFgColor_override"		"TanLight"
	"depressedFgColor_override" "TanLight"
} // JoinQueueButton
```

The entire modified file is also attached to this PR:
[questviewsubpanel.res.txt](https://github.com/user-attachments/files/24983773/questviewsubpanel.res.txt) (had to upload as .txt because github sucks)

Also fixes https://github.com/ValveSoftware/Source-1-Games/issues/4406

### DIFF
--- a/src/game/client/tf/vgui/tf_quest_map_node_view_panel.cpp
+++ b/src/game/client/tf/vgui/tf_quest_map_node_view_panel.cpp
@@ -176,6 +176,8 @@ CQuestViewSubPanel::CQuestViewSubPanel( Panel* pParent,
 	m_pObjectivePanel = new CQuestObjectivePanel( this, "objectives" );
 	m_pActivateButton = new CExButton( this, "ActivateButton", (const char*)NULL, this, "activate_node" );
 	m_pActivateButton->PassMouseTicksTo( this, true );
+	m_pJoinQueueButton = new CExButton( this, "JoinQueueButton", (const char*)NULL, this, "join_queue" );
+	m_pJoinQueueButton->PassMouseTicksTo( this, true );
 	m_pBGImage = new ImagePanel( this, "BGImage" );
 	m_pAcceptTooltipHack = new EditablePanel( this, "EditableTooltip" );
 	m_pInfo = new ImagePanel( this, "ObjectivesInfoImage" );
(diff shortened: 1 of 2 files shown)


rabscootle:
If you would like to add it to the post, this would resolve https://github.com/ValveSoftware/Source-1-Games/issues/4406

PR #1787, opened 2026-02-02
[TF2] Added option to display owned weapons' name/strange count in Player Status HUD

rabscootle:
Resolves https://github.com/ValveSoftware/Source-1-Games/issues/4608

This PR aims to add additional functionality to the Player Status HUD via a new convar (cl_hud_playerclass_display_weapon_info, default = 0).

When enabled (value = 1), the Player Status HUD will display the item name underneath the player portrait (similar to when picking up a dropped weapon). This function differs from the one used for dropped weapons and supports more detailed names:
Photoshop_WmR6mI0fyH

'Dropped by' portion of UI has been retooled for player's own weapons to display primary Strange count (when one is available). Code dynamically pulls localized string token to support other languages:
Discord_6voY9TZzEI

PR has been tested in community modes with custom weapons (Zombie Infection, VSH):
Photoshop_o3UKMDybHY

Feature has been disabled in Mann vs Machine due to conflicts with existing UI (image below is image of weapon info UI if kept visible in MvM)
Discord_1Yj9t4aJ5e

Known issue is that weapon info panel is known to conflict with the Killstreak meter if an item has a long enough name:
Discord_QzAe5xORQh

A similar issue can currently be experienced when inspecting dropped weapons.

If constructed item name is too long for your liking, replace _pItem->GetItemName()_ with _CEconItemLocalizedFullNameGenerator(GLocalizationProvider(), pItem->GetItemDefinition(), pItem->GetItemQuality()).GetFullName()_

image

### DIFF
--- a/src/game/client/tf/tf_hud_playerstatus.cpp
+++ b/src/game/client/tf/tf_hud_playerstatus.cpp
@@ -41,6 +41,8 @@ ConVar cl_hud_playerclass_use_playermodel( "cl_hud_playerclass_use_playermodel",
 
 ConVar cl_hud_playerclass_playermodel_showed_confirm_dialog( "cl_hud_playerclass_playermodel_showed_confirm_dialog", "0", FCVAR_ARCHIVE | FCVAR_HIDDEN );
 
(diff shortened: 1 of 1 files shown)


GrubPL:
kinda sad that valve only seems to merge bugfix PR's (and even with that they only merge a few PR's)
it would be quite cool to see it in the actual game

PR #1788, opened 2026-02-02
[TF2] Fix taunt cam zoom in effect not working on finished taunts

rawra89:
This PR fixes the taunt cam zoom in effect being broken when players finish taunting. This used to be a feature added around 2009-early 2010s, but broke sometime in 2015.

Before:

https://github.com/user-attachments/assets/2b3c0edf-b521-43a0-aaae-f26f4a28e2b4

After:

https://github.com/user-attachments/assets/81655a8b-f44c-4e8f-95f1-96d19800256e


rawra89:
~~There's currently a bug where if you changelevel during the zoom in effect, the player will start in thirdperson. Will fix later today.~~ Fixed.

rawra89:
~~Found another bug where if you switch classes while thirdperson is on, it will force you back to first person. Going to convert this into a draft for now until it's fixed.~~ This appears to be an issue only when ``sv_cheats`` is on.

src/game/client/tf/c_tf_player.cpp
inco-cc:
```suggestion
		m_flTauntCamTargetDist = 0.f;
		m_flTauntCamTargetDistUp = 0.f;
```

This will reset the camera height offset target in addition to the camera distance.

inco-cc:
```suggestion
```
This is causing the third person camera to ignore `m_flTauntCamCurrentDistUp`, making the camera jump back to zero height offset immediately after the taunt ends and starts interpolating back to first person view.

inco-cc:
```suggestion
	g_ThirdPersonManager.SetOverridingThirdPerson( false );

```
This seems like an appropriate place to call `g_ThirdPersonManager.SetOverridingThirdPerson`. This way the camera override will stop when the taunt camera has fully stopped and we're back in first person view, allowing `m_flTauntCamTargetDistUp` to affect the camera height during `C_TFPlayer::TauntCamInterpolation`.

src/game/client/tf/c_tf_player.cpp
```diff
@@ -5706,14 +5706,12 @@ void C_TFPlayer::TauntCamInterpolation()
 	{
 		if ( m_flTauntCamCurrentDist != m_flTauntCamTargetDist )
 		{
-			m_flTauntCamCurrentDist += Sign( m_flTauntCamTargetDist - m_flTauntCamCurrentDist ) * gpGlobals->frametime * tf_tauntcam_speed.GetFloat();
-			m_flTauntCamCurrentDist = clamp( m_flTauntCamCurrentDist, m_flTauntCamCurrentDist, m_flTauntCamTargetDist );
+			m_flTauntCamCurrentDist = Approach( m_flTauntCamTargetDist, m_flTauntCamCurrentDist, gpGlobals->frametime * tf_tauntcam_speed.GetFloat() );
 		}
```
inco-cc:
```suggestion
			m_flTauntCamCurrentDistUp = ( m_flTauntCamCurrentDist / tf_tauntcam_dist.GetFloat() ) * m_flTauntCamTargetDistUp;
		}
```
`m_flTauntCamCurrentDistUp` can instead be calculated based on `m_flTauntCamCurrentDist`, making interpolation much smoother for taunts with a camera height offset.

src/game/client/tf/c_tf_player.cpp
inco-cc, line 5565:
```suggestion
```
This is causing the third person camera to ignore `m_flTauntCamCurrentDistUp`, making the camera jump back to zero height offset immediately after the taunt ends and starts interpolating back to first person view.

inco-cc, line 5603:
```suggestion
	g_ThirdPersonManager.SetOverridingThirdPerson( false );
```
This seems like an appropriate place to call `g_ThirdPersonManager.SetOverridingThirdPerson`. This way the camera override will stop when the taunt camera has fully stopped and we're back in first person view, allowing `m_flTauntCamTargetDistUp` to affect the camera height during `C_TFPlayer::TauntCamInterpolation`.

src/game/client/tf/c_tf_player.cpp
inco-cc, line 5710:
```suggestion
			m_flTauntCamCurrentDistUp = ( m_flTauntCamCurrentDist / tf_tauntcam_dist.GetFloat() ) * m_flTauntCamTargetDistUp;
		}
```
`m_flTauntCamCurrentDistUp` can instead be calculated based on `m_flTauntCamCurrentDist`, making interpolation much smoother for taunts with a camera height offset.

Note that this uses `tf_tauntcam_dist` instead of `m_flTauntCamTargetDist` since the latter changes back to zero when zooming back in, which of course would cause the game to crash in this calculation.

src/game/client/tf/c_tf_player.cpp
```diff
@@ -5706,14 +5706,12 @@ void C_TFPlayer::TauntCamInterpolation()
...
 		
 		if ( m_flTauntCamCurrentDistUp != m_flTauntCamTargetDistUp )
 		{
-			m_flTauntCamCurrentDistUp += Sign( m_flTauntCamTargetDistUp - m_flTauntCamCurrentDistUp ) * gpGlobals->frametime * tf_tauntcam_speed.GetFloat();
-			m_flTauntCamCurrentDistUp = clamp( m_flTauntCamCurrentDistUp, m_flTauntCamCurrentDistUp, m_flTauntCamTargetDistUp );
+			m_flTauntCamCurrentDistUp = Approach(m_flTauntCamTargetDistUp, m_flTauntCamCurrentDistUp, gpGlobals->frametime * tf_tauntcam_speed.GetFloat() );
 		}
```
inco-cc, line 5715: This if statement is no longer needed when `m_flTauntCamCurrentDistUp` is calculated at the same time as `m_flTauntCamCurrentDist`.

src/game/client/tf/c_tf_player.cpp
inco-cc, line 5737:
```suggestion
		if ( m_flTauntCamCurrentDist == m_flTauntCamTargetDist )
```
After moving the new `m_flTauntCamCurrentDistUp` calculation to the same block as `m_flTauntCamCurrentDist` at the start of this method, this statement should be changed to only check `m_flTauntCamCurrentDist`, otherwise the taunt camera will never finish since `m_flTauntCamTargetDistUp` doesn't reset back to zero.

PR #1756, opened 2026-01-10
[TF2] Fixed war paint items not displaying rarity color in various UI

rabscootle:
Partial fix for https://github.com/ValveSoftware/Source-1-Games/issues/3520

This PR fixes rarities not applying properly in various UI when handling war-painted items.

When dealing with dropped weapons, the live TF client currently typically points to static data in the item schema for weapon details, including item rarities. However, War Paint items bypass requiring their own respective schema entries by having attributes applied to individual instances of base paintkit items. This change meant that while the same attributes are present on war paint items, certain game code was not updated to support this:
image

Game code has been updated to extract details from dropped weapons rather than static schema entry to display rarity for War Paints in the following UI:
**Weapon Pickup (Target HUD)**
image
**Loadout - Item Selection Screen**
image
**Player Status HUD (when using dropped weapon)**
image

PR #1792, opened 2026-02-06
Warpaint id fix

rabscootle:
Partial fix for https://github.com/ValveSoftware/Source-1-Games/issues/3520

This PR fixes item names not constructing properly when handling war-painted items in certain UI (Player Status HUD, Target ID):
image

When dealing with dropped weapons, the live TF client currently typically points to static data in the item schema for weapon details, including paint kit index IDs. However, War Paint weapons bypass requiring their own respective schema entries by having attributes applied to individual instances of generic paintkit weapon entries. This approach breaks some UI code where it is expecting the attributes to be on the item schema definition:
image

Game code has been updated to store a paintkit index within CGameItemDefinition_EconItemInterfaceWrapper, which can be referenced in situations where we cannot point to an actual item instance.

When paired with #1756, War Paint items should have full feature parity with existing Decorated Weapons and will display War Paint name + Rarity Color in affected UI:
image

### DIFF
--- a/src/game/shared/econ/econ_item_description.h
+++ b/src/game/shared/econ/econ_item_description.h
@@ -351,7 +351,7 @@ enum EGenerateLocalizedFullItemNameFlag_t
 class CEconItemLocalizedFullNameGenerator
 {
 public:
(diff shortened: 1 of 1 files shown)


rabscootle:
Resubmitting this because I messed up the PR name (edit perms seem to have been locked down recently)

PR #856, opened 2025-02-26
[TF2] Fixes for War Paint support + additional item details in several UI

rabscootle:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/3520

When dealing with dropped weapons, the live TF client currently typically points to static data in the item schema for weapon details, particularly paintkit IDs and item rarities. However, War Paint items bypass requiring their own respective schema entries and instead have attributes applied to individual instances of base paintkit items. This change meant that while the same attributes were present on war paint items, they were not being picked up by the game for specific UI.
image (2)

Game code has been updated to extract details from currently held weapon rather than static schema entry for compatibility with War Paint weapons + additional item details.

Fixed rarity and weapon skin name not displaying on dropped weapons:
image
Additional check for Self-Made quality has been added to match other economy UI behavior.

Dropped weapon UI will also now display additional item details (ie. Australium, Strange Rank). It will also display custom names to their owners:
image

Fixed rarity and weapon skin name not displaying in playerstatus UI for dropped war paint items:
image

Fixed color border not updating to match rarity of war paint items in Loadout Selection UI:
image

PR #1793, opened 2026-02-06
[TF2] Fixed several UI being unable to pull paintkit index/pattern name for War Painted weapons

rabscootle:
Partial fix for https://github.com/ValveSoftware/Source-1-Games/issues/3520

**This PR is a newer, less lazy approach to what was previously done in the now-outdated #856.**

This PR fixes item names not constructing properly when handling war-painted items in certain UI (Player Status HUD, Target ID):
image
image

When paired with #1756, War Paint items should have full feature parity with existing Decorated Weapons and will display War Paint name + Rarity Color in affected UI:
image

When dealing with dropped weapons, the live TF client currently typically points to static data in the item schema for weapon details, including paint kit index IDs. However, War Paint weapons bypass requiring their own respective schema entries by having attributes applied to individual instances of generic paintkit weapon entries. This approach breaks some UI code where it is expecting the attributes to be on the item schema definition:
image

Game code has been updated to store a paintkit index within CGameItemDefinition_EconItemInterfaceWrapper, which can be referenced in situations where we cannot point to an actual item instance.

A side effect of this fix is that item qualities will now properly display on constructed item names in the affected UI, which was caused by a missing argument.
image

### DIFF
--- a/src/game/shared/econ/econ_item_description.h
+++ b/src/game/shared/econ/econ_item_description.h
@@ -351,7 +351,7 @@ enum EGenerateLocalizedFullItemNameFlag_t
 class CEconItemLocalizedFullNameGenerator
 {
 public:
-	CEconItemLocalizedFullNameGenerator( const CLocalizationProvider* pLocalizationProvider, const CEconItemDefinition* pItemDef, bool bUseingHashContext = false, entityquality_t eQuality = AE_UNIQUE, uint32 nPaintkitDefIndex = 0 );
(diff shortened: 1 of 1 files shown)


kisak-valve:
Minor side note that you can give me a ping and request adjustments if GitHub isn't being reasonable. This and #1792 could have been a git force push to the git branch you were using when you opened #856 and adjusting the description. Having written that, since we've already gotten to this point, there's no point in more housekeeping.

rabscootle:
Sorry! This is where I am still learning Git, will account for this next time.

PR #1794, opened 2026-02-07
[TF2] Allow TF_COND_INVULNERABLE_HIDE_UNLESS_DAMAGED to function outside MvM

FlaminSarge:
# Description
This allows `TF_COND_INVULNERABLE_HIDE_UNLESS_DAMAGED` to display its effects properly outside MvM (shows Ubercharge skin when damaged). In order to prevent networking data when not necessary, it only sets `m_flMvMLastDamageTime` for players that have the condition (this netprop is not used for any other feature in the game).
In addition, it also adds the relevant logic to `C_TFPlayer::IsOverridingViewmodel` and `C_TFPlayer::DrawOverriddenViewmodel` so that a player's viewmodels gain the same effect (only showing the Uber skin when taking damage, instead of live behavior of showing it at all times).

Demonstration: https://youtu.be/ospEI-UsMTY

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -8334,7 +8334,9 @@ bool C_TFPlayer::IsOverridingViewmodel( void )
 		pPlayer = assert_cast<C_TFPlayer*>(pLocalPlayer->GetObserverTarget());
 	}
 
-	if ( pPlayer->m_Shared.IsInvulnerable() )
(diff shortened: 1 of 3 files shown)


FlaminSarge:
Note that the netprop and related methods haven't been renamed, despite no longer being MvM-specific, to maintain compatibility.

GrubPL:
Maybe this will be merged as part of the mvm update?
I guess we'll just have to wait and see

Issue #612, opened 2025-02-18
Data race in CUtlSymbolTableMT lookup.


RoyaleNoir:
[`CUtlSymbolTableMT::Find` uses a read lock](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/public/tier1/utlsymbol.h#L197), where it should be using a write lock, as the [function it is wrapping writes to a member variable of `CUtlSymbolTable`](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/tier1/utlsymbol.cpp#L184).

dimhotepus:
THIS IS COMMON BUG IN AT LEAST HL2 / HL2 Episodes / Portal / TF2 GAMES.

For example it causes HL2 coast or TF2 maps with water begin to blink due to `Water` shader can't find fog variables in
`CUtlSymbolTable`.

The problem is `CUtlSymbolTableMT::Find` may be called by multiple threads at once.  It creates
read lock and calls `CUtlSymbolTable::Find`.  `CUtlSymbolTable::Find` in turn stores
searched string inside class to access it in the tree lookup.  It means when `CUtlSymbolTableMT::Find`
called by multiple threads, its search string may be changed in the middle time during lookup as
read lock doesn't protect against writes.

At least one known bug: https://github.com/ValveSoftware/Source-1-Games/issues/4937

There are different possible ways to fix the one:
1. Mark user search string as `CThreadLocal`: https://github.com/Source-Authors/Obsoletium/commit/663a12a69ec750bc829219f5ab9de7b757316a8f. Downside is it adds TLS access overhead in regular `CUtlSymbolTable`. 
2. Extend `Find` context for `CUtlRBTree` aka add context to comparator: https://github.com/Source-Authors/Obsoletium/commit/b995016bd7d298ff0abe5d2598741e716522eb6d. It is a bit ugly (guess can be done better), but no performance overhead for TLS access as we just use search string from stack-local context (no thread safety issues).

ficool2:
I fixed this months ago in my mod but I've opened a PR for it now https://github.com/ValveSoftware/source-sdk-2013/pull/1807

PR #1807, opened 2026-02-15
Fix race condition in CUtlSymbolTableMT lookup

ficool2:
See https://github.com/ValveSoftware/source-sdk-2013/issues/612 for an explanation of how the race condition arises

This PR puts the search string on the stack, which prevents threads from reading a stale pointer that got switched out between execution

This mainly fixes flickering water materials. Materialsystem uses a CUtlSymbolTableMT lookup for material vars internally, and therefore it will occasionally read a stale state for fog parameters. It also fixes various other random bugs such as eyes becoming missing textures

### DIFF
--- a/src/public/tier1/utlsymbol.h
+++ b/src/public/tier1/utlsymbol.h
@@ -139,6 +139,17 @@ protected:
 		unsigned short m_iOffset;	// Index into the string pool.
 	};
 
+	// Helper to store search string on the stack for safe thread access
+	class CStringPoolIndexSearch : public CStringPoolIndex
+	{
+	public:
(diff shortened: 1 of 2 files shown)


src/public/tier1/utlsymbol.h
```diff
@@ -139,6 +139,17 @@ class CUtlSymbolTable
 		unsigned short m_iOffset;	// Index into the string pool.
 	};
 
+	// Helper to store search string on the stack for safe thread access
+	class CStringPoolIndexSearch : public CStringPoolIndex
+	{
+	public:
+		CStringPoolIndexSearch( const char* pString ) : CStringPoolIndex( 0xFFFF, 0xFFFF )
```
dimhotepus, line 146: Nice trick, thanks!

dimhotepus (approved):

PR #1815, opened 2026-02-23
[DODS] Increase MAX_PLAYERS to 101

kuninyao:
# Description

As fans of Day Of Defeat: Source, we want support for 101 players. We tested it on our compiled build and 101 players on the new engine is playable (tested on 101 bots and 36 real players). You can see this in the screenshot: we had 18 players on one team and 18 on the other.

photo_2026-02-23_18-38-44

Unfortunately, we were only able to gather 36 real players, but that was because players had difficulty downloading custom DLLs. It would be a lot of fun to have 101 players in a game like this.

This game is approaching its 20th anniversary (released September 26, 2005). We hope that at least for the anniversary (by Valve Time™), it will receive support for 101 players.

PR #1816, opened 2026-02-26
[MP] Globalize -unrestricted_maxplayers behavior across all mods

pivotman319-owo:
# Description
This PR eliminates the janky per-mod maxplayers `#ifdef`s in `shareddefs.h` by adding VPC conditional var `UNRESTRICTED_MAXPLAYERS_ENABLE` for all mp mod projects, which enables C++ preprocessor block `UNRESTRICTED_MAXPLAYERS`. Doing this required a slight refactor of the gameinterface object files to account for the more flexible maxplayers restrictions on amd64 and x86/i686, including the `-unrestricted_maxplayers` launch arg that is only exclusive to 32-bit builds. Mods can choose to opt-in or out of this by (not) including this conditional var. It is enabled by default in TF and HL2MP mods.

Note that this PR requires applying the same gameinterface code rewrites for other mods that do not have public source code, particularly `cstrike` and `dod` (the latter of which is exactly why this PR was made in the first place - this is essentially a cleaner way of implementing #1815, but the changes here apply to ALL games).

`UNRESTRICTED_MAXPLAYERS_ENABLE == 0` on amd64:

VPC var UNRESTRICTED_MAXPLAYERS_ENABLE disabled on hl2mp mod

`UNRESTRICTED_MAXPLAYERS_ENABLE == 1` on amd64:

VPC var UNRESTRICTED_MAXPLAYERS_ENABLE enabled on hl2mp mod

### DIFF
--- a/src/game/server/base_gameinterface.cpp
+++ b/src/game/server/base_gameinterface.cpp
@@ -11,29 +11,31 @@
 // memdbgon must be the last include file in a .cpp file!!!
 #include "tier0/memdbgon.h"
 
-void CServerGameClients::GetPlayerLimits( int& minplayers, int& maxplayers, int &defaultMaxPlayers ) const
+void CServerGameClients::GetPlayerLimits(int& minplayers, int& maxplayers, int& defaultMaxPlayers) const
 {
 	minplayers = defaultMaxPlayers = 1;
 
 	// Pivot (25/02/2026): If targeting amd64 platforms AND conditional UNRESTRICTED_MAXPLAYERS_ENABLE
 	// set to 1 in game project vpc, raise to MAX_PLAYERS limit specified in game/shared/shareddefs.h.
-	
+
 	// Code adapted from game/server/tf/tf_gameinterface.cpp
-#if defined( PLATFORM_64BITS ) && defined( UNRESTRICTED_MAXPLAYERS )	// Assuming amd64
+#ifdef UNRESTRICTED_MAXPLAYERS
+#ifdef PLATFORM_64BITS	// Assuming amd64
 	maxplayers = MAX_PLAYERS;
-#elif !defined ( PLATFORM_64BITS ) && defined( UNRESTRICTED_MAXPLAYERS )	// Assuming x86/i686
-	if ( CommandLine()->HasParm("-unrestricted_maxplayers") )
+#else	// Assuming x86/i686
+	if (CommandLine()->HasParm("-unrestricted_maxplayers"))
 	{
 		static bool s_bWarned = false;
-		if ( !s_bWarned )
+		if (!s_bWarned)
 		{
-			Warning( "The use of -unrestricted_maxplayers is NOT supported and definitely NOT recommended and may be unstable.\n" );
+			Warning("The use of -unrestricted_maxplayers is NOT supported and definitely NOT recommended and may be unstable.\n");
 			s_bWarned = true;
 		}
 		maxplayers = MAX_PLAYERS;
 	}
 	else
 		maxplayers = 33;	// Accounting for one extra SourceTV client.
+#endif
 #else 
 	maxplayers = MAX_PLAYERS;	// MAX_PLAYERS = 33
 #endif
@@ -44,6 +46,6 @@ void CServerGameClients::GetPlayerLimits( int& minplayers, int& maxplayers, int 
 // Mod-specific CServerGameDLL implementation.
 // -------------------------------------------------------------------------------------------- //
 
-void CServerGameDLL::LevelInit_ParseAllEntities( const char *pMapEntities )
+void CServerGameDLL::LevelInit_ParseAllEntities(const char* pMapEntities)
 {
 }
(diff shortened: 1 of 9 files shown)


vrad-exe:
Not sure why you've added explicit checks for the singleplayer games, considering those are on a separate branch and also you said you wanted to remove "janky per-mod maxplayers #ifdefs" which are exactly what those are. Multiplayer in Portal 1 is partially functional and there are some people who make use of it, so it would be appreciated if you didn't arbitrarily lock that out (in the event that the SP/MP branches ever get merged).

pivotman319-owo:
> Not sure why you've added explicit checks for the singleplayer games, considering those are on a separate branch and also you said you wanted to remove "janky per-mod maxplayers #ifdefs" which are exactly what those are. Multiplayer in Portal 1 is partially functional and there are some people who make use of it, so it would be appreciated if you didn't arbitrarily lock that out (in the event that the SP/MP branches ever get merged).

Addressed in d0f8020 - the original justification for that was with regards to crashes in the AI code always assuming one player in the server, and never multiple. Apologies for not knowing earlier.

JoriKos:
Adding your name to nearly every comment added is also unnecessary, unless you specifically need to be addressed if this ever gets changed.

kuninyao:
Yes, that's certainly better, since it applies to a wider range of games than Day of Defeat: Source.

anzz1:
Isn't 100 arbitrary here too? Could it be 255 (byte max)?

JoriKos:
I think 100 is fine, it was probably done to match what TF2 has and seeing as that's already unsupported, not sure that a higher amount is necessary.

etkaar:
Since we are speaking of lifting a restriction rather then raising a maximum, `255` which is the maximum of an `uint8` (1 byte = 8 bits = 0–255 if unsigned) makes more sense to me.

pivotman319-owo:
> Since we are speaking of lifting a restriction rather then raising a maximum, `255` which is the maximum of an `uint8` (1 byte = 8 bits = 0–255 if unsigned) makes more sense to me.

That would cause stability to go down the drain given current engine limitations, you'd be exhausting the networked edict limit immediately at this point (even 100 players on TF2 is too much if you account for a `CTF_Player` ent, a player viewmodel, 3 weapons and 3 cosmetics + 1 misc - that would already be ~900 edicts occupied without any combat going on on a 100 player server!!!).

If we had 255 players at once, the game would just crash straight away on an official map because we'd be occupying ~2,295 out of 2048 networked edicts for just the players alone, which is *unplayable* for any reasonable sort of gameplay use case. If every one of those players were a Demo on an empty map wearing a fully-kitted loadout with cosmetics and a grenade/sticky combo, and ~~at least one of them drained their entire sticky launcher reserve, the server would no longer have any space to occupy any additional edicts.~~ all of them spawned into the server at once, the engine would die immediately.

etkaar:
That might be true, but then

- should other limits be increased as well if `-unrestricted_maxplayers` is enabled, such as the edicts limit—I believe in Garry's Mod it is already 8192—, or
- it shouldn't be named `-unrestricted_maxplayers` but `-increased_maxplayers`.

vrad-exe:
Also who cares if it's hard to run without hitting the edict limit, servers can just use a lower limit if 255 doesn't work for them, run cosmetic disabling plugins (which many 100 player servers have to do anyway) and/or simplified maps, etc. I personally am in favor of making unrestricted maxplayers actually allow up to the engine player limit of 255.

anzz1:
> If every one of those players were a Demo on an empty map wearing a fully-kitted loadout with cosmetics and a grenade/sticky combo

Why are you talking about TF2 specifically when the PR is for the source SDK that is not just TF2, which is the entire point here. I'm getting the feeling that the OP isn't really listening to the good arguments here that `-unrestricted_maxplayers` should indeed unrestrict the player count as the name implies to the engine limit of 255 instead of some arbitrary limit of 100 and let the server owners worry about testing what works, so instead of trying to convince OP it might be worthwhile to just create a new PR.

Increasing the entity limit from current 6144/2048 (non-networked/networked), for example to Garry's mod 8192/8192, or perhaps the 16384/16384 the Source2 engine uses, would be another nice change as indeed the edict limit is often the problem for highly customized servers and more complex maps.

eepycats:
you would get a lag spike for like 2 seconds each time you enter a pvs with 255 players

src/game/server/base_gameinterface.cpp
```diff
@@ -11,17 +11,41 @@
 // memdbgon must be the last include file in a .cpp file!!!
 #include "tier0/memdbgon.h"
 
-void CServerGameClients::GetPlayerLimits( int& minplayers, int& maxplayers, int &defaultMaxPlayers ) const
+void CServerGameClients::GetPlayerLimits(int& minplayers, int& maxplayers, int& defaultMaxPlayers) const
 {
-	minplayers = defaultMaxPlayers = 1; 
+	minplayers = defaultMaxPlayers = 1;
+
+	// Pivot (25/02/2026): If targeting amd64 platforms AND conditional UNRESTRICTED_MAXPLAYERS_ENABLE
```
ficool2, line 18: The convention is to not put your username and dates in PRs

PR #1819, opened 2026-03-02
[ALL] Restore ent_fire functionality while still blocking harmful commands

ghost:
Currently, the **ent_fire** console command is only allowed to be used listen server hosts.

There is a large comment in the SDK explaining that this was done because otherwise it could be used by anyone to run any command on a `sv_cheats 1` server using the `command` input on a **point_servercommand** entity. Players could use this to change the rcon password and grief servers. Supposedly, there would be no way to avoid this misuse as the point_servercommand entity can be **renamed** and the input can be sent with a **delay**.

The person who made that change seems to have not realized that **you can simply check if the input of an ent_fire is** `command` **or not**, and avoid the event from being created if it is. Doing this would be far less restrictive and still allow players to play around with `sv_cheats 1` without being able to change the rcon password.

It would be a godsend for scripting / sv_cheats 1 servers (which I still regularly encounter in HL2DM and HLDMS) and restore much of the fun that was lost due to this change

### DIFF
--- a/src/game/server/baseentity.cpp
+++ b/src/game/server/baseentity.cpp
@@ -6010,33 +6010,15 @@ public:
 
 			target = STRING( AllocPooledString(command.Arg( 1 ) ) );
 
-			// Don't allow them to run anything on a point_servercommand unless they're the host player. Otherwise they can ent_fire
-			// and run any command on the server. Admittedly, they can only do the ent_fire if sv_cheats is on, but 
-			// people complained about users resetting the rcon password if the server briefly turned on cheats like this:
-			//    give point_servercommand
-			//    ent_fire point_servercommand command "rcon_password mynewpassword"
-			//
-			// Robin: Unfortunately, they get around point_servercommand checks with this:
-			//	  ent_create point_servercommand; ent_setname mine; ent_fire mine command "rcon_password mynewpassword"
-			// So, I'm removing the ability for anyone to execute ent_fires on dedicated servers (we can't check to see if
-			// this command is going to connect with a point_servercommand entity here, because they could delay the event and create it later).
-			if ( engine->IsDedicatedServer() )
-			{
-				// We allow people with disabled autokick to do it, because they already have rcon.
-				if ( pPlayer->IsAutoKickDisabled() == false )
-					return;
-			}
-			else if ( gpGlobals->maxClients > 1 )
-			{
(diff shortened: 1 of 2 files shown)


src/game/server/baseentity.cpp
```diff
@@ -6010,33 +6010,15 @@ class CEntFireAutoCompletionFunctor : public ICommandCallback, public ICommandCo
...
-				CBasePlayer *pHostPlayer = UTIL_GetListenServerHost();
-				if ( pPlayer != pHostPlayer )
-					return;
-			}
+			
 
 			if ( command.ArgC() >= 3 )
 			{
 				action = STRING( AllocPooledString(command.Arg( 2 )) );
+				if (StringHasPrefix(action, "command")) //Block players misusing point_servercommand
```
ficool2, line 6018: Any measures like this are useless because they can be bypassed by other commands like ent_create

ghost: Elaborate

kuninyao:
You can already use ent_fire command with full functionality, can't you? Run the **status** command in the console to find your client ID, for example:

```
2 "⁧⁧Kuninyao" [U:1:666]    38:42       35    0 active loopback
^ your client id
```

then run 
```
mp_disable_autokick 2 
                    ^ client id from status command
```
and the command’s functionality will be restored.
`ent_fire !picker kill`, etc, will work.

ghost:
@kuninyao I guess that would work for listen servers, didn't think about that. Would still be nice if Volvo could fix it so that it works by default for both listen and dedicated servers while still blocking misuse

src/game/server/baseentity.cpp
bakugo: If a server has `sv_cheats 1` and either has `sv_allow_point_servercommand always` or is running a map made by Valve, it's currently possible to execute arbitrary server commands regardless of any of these protections. I won't share the method, but it's not difficult.

If you absolutely *need* to enable `sv_cheats` on a public server, you should use a server plugin that fully blocks any dangerous cheat commands such as `ent_create` and `ent_fire` (example [here](https://github.com/ficool2/TF2Ware_Ultimate/blob/main/addons/sourcemod/scripting/tf2ware_ultimate.sp) in ficool2's tf2ware mod).

PR #1821, opened 2026-03-02
[ALL] Allow godmode despite deathmatch being enabled

ghost:
For some reason the `god` command is always blocked when `gpGlobals->deathmatch` is enabled, even when `sv_cheats 1` is on

What's the point of this? It doesn't seem to break anything or cause any glitches. If cheats are on then you should be able to enter godmode IMO

### DIFF
--- a/src/game/server/client.cpp
+++ b/src/game/server/client.cpp
@@ -1231,16 +1231,6 @@ void CC_God_f (void)
 	if ( !pPlayer )
 		return;
 
-#ifdef TF_DLL
-   if ( TFGameRules() && ( TFGameRules()->IsPVEModeActive() == false ) )
-   {
-	   if ( gpGlobals->deathmatch )
-		   return;
-   }
-#else
-	if ( gpGlobals->deathmatch )
-		return;
-#endif
 
 	pPlayer->ToggleFlag( FL_GODMODE );
 	if (!(pPlayer->GetFlags() & FL_GODMODE ) )


inco-cc:
You should provide a brief summary of the changes in your commit messages. All of your recent PRs have commit messages like `Update client.cpp` or `Fix`, which doesn't make it obvious what changes were made. Using the titles you used for your PRs would be perfectly adequate.

JoriKos:
Wouldn't it be better to add a check for `sv_cheats` being set to `1` rather than deleting it outright, especially if the purpose is to allow it only when cheats are enabled? There may be niche cases that break if removed entirely.

callumok2004:
> Wouldn't it be better to add a check for `sv_cheats` being set to `1` rather than deleting it outright, especially if the purpose is to allow it only when cheats are enabled? There may be niche cases that break if removed entirely.

this check already exists above

PiMoNFeeD:
Note that `FL_GODMODE` will not actually work in most games, because it's only checked for in `CBasePlayer` for some reason, while most games (CSS, DODS, even HL1DM and TF2) skip over `CBasePlayer::OnTakeDamage` and call straight into `CBaseCombatCharacter::OnTakeDamage`, which means you have to move the check from player code to BCC code

ghost:
> Note that `FL_GODMODE` will not actually work in most games, because it's only checked for in `CBasePlayer` for some reason, while most games (CSS, DODS, even HL1DM and TF2) skip over `CBasePlayer::OnTakeDamage` and call straight into `CBaseCombatCharacter::OnTakeDamage`, which means you have to move the check from player code to BCC code

I've been occasionally hosting a HLDMS server for like a year now where this exact change is patched in and the godmode works fine, even without making any edits to Cbaseplayer or Cbasecombatcharacter

PiMoNFeeD:
> > Note that `FL_GODMODE` will not actually work in most games, because it's only checked for in `CBasePlayer` for some reason, while most games (CSS, DODS, even HL1DM and TF2) skip over `CBasePlayer::OnTakeDamage` and call straight into `CBaseCombatCharacter::OnTakeDamage`, which means you have to move the check from player code to BCC code
> 
> I've been occasionally hosting a HLDMS server for like a year now where this exact change is patched in and the godmode works fine, even without making any edits to Cbaseplayer or Cbasecombatcharacter

You made me double-check and I was wrong about TF2, because it implements the check in it's own `OnTakeDamage` function, and I actually missed what HL1 does: instead of just denying damage like every other game, it restores old health in `OnTakeDamage_Alive` which is a massive hack and should really be removed...
<details>
<summary>Decompiled code</summary>

(I meant to say `CNetworkVarBase` instead of `CNetworkHandleBase` in the comment)
image
</details>

Regardless, since not every game implements it and not every game calls `CBasePlayer`'s damage function, its best to just move the check out of player class

PiMoNFeeD:
> Wtf why are you decompiling the game when the source code is public? HLDMS files are in the 2020 TF2 leak along with every other 2013 multiplayer game. It's 100x easier to just check those

Do you even understand where you're commenting this..? No wonder Valve hate Source licensing nowadays


> I agree that the godmode could be better though (and thank you for letting me know about it). But shouldn't that be a different issue / PR then? Since this is technically only about the check for gpGlobals->deathmatch

Probably, but I'd repurpose this into a "make godmode available in all games" commit tbh, all games on this branch are gonna have deathmatch 1 regardless (since they're multiplayer), but like I said before, not all games actually implement `FL_GODMODE`

PR #1822, opened 2026-03-02
Autocomplete for chained console commands

callumok2004:
# Description

Adds autocomplete for chained console commands (seperated by `;`)

mod_tf_win64_icry09RI3Q


GrubPL:
This is peak

ficool2:
This has a bug when switching back to previous delimited commands with the up/down arrow key

To replicate:

`echo hello;echo world`

Then press up arrow and down key and notice how the command is appending itself constantly

callumok2004:
> This has a bug when switching back to previous delimited commands with the up/down arrow key
> 
> To replicate:
> 
> `echo hello;echo world`
> 
> Then press up arrow and down key and notice how the command is appending itself constantly

Good catch, not sure how I missed that. Should be working as expected now 

https://github.com/user-attachments/assets/c5c58ba3-04bf-40ca-9f96-ca7892e0f82d

Generalisk:
Oh god, thank you! I have been wanting this for ages!

PR #1823, opened 2026-03-03
[ALL] Remove redundant preprocessor directives from movevars_shared.cpp

ghost:
**ISSUE:** Movevars_shared.cpp contains multiple `#if ... #else` statements which have the exact same code in the `#if` blocks as the `#else` blocks. There is literally no point in using an if statement here

I propose that they are simplified into single statements which do not make use of preprocessor directives.


mastercoms:
They're structured like that for TF2, to separate changes for the vars. Historically, it was to add cheat flags. If TF diverges again, then the conditionals need to be reintroduced. I say it's better to keep it since it's not really hurting anything.

ghost:
Alright well fml then

PR #1599, opened 2025-10-30
Updated Weapon Extra Wearables and Body Groups  to be compatible with disguises

seanmcgeehan:
# Description

This commit solves one of the bugs related to missing wearables and body groups for spy disguises: https://github.com/ValveSoftware/Source-1-Games/issues/275 (at least a partial fix for this issue)

Particularly this solves wearables for weapons that have "extraWearables". Upon pressing last disguise on the relevant item slot, The spy will receive the wearable until he switches his class disguise even if he switches disguise weapons. This version of the fix increases the skill ceiling of spy while reducing code blast radius.

This fix also solves bad body group settings for weapons like the quickfix and vaccinator in its second revision. In the medigun case this results in correct backpacks being placed for all mediguns for disguised spies. 

Tested and confirmed:
* The QuickFix
* Battalions Back up
* Buff Banner
* Thermal thruster (Only inactive state, but this allows for additional valid disguises so is an improvement)
* Councheror
* Vaccinator
* Base Jumper

Before fix:
Screenshot (2)

After fix:
Screenshot (10)
Screenshot (12)
Screenshot (11)

PR #1824, opened 2026-03-05
Fix all known broken spy disguises

### LINKED ISSUE #275: TF2 - Spy Disguise missing wearable equipped items
When a Spy disguises as someone equipped with Ali Baba's Wee Booties or the Bootlegger, these items do not appear on the Spy.

This means you see a DemoMan with no legs running around and it's extremely obvious that player is a Spy.

### LINKED ISSUE #5883: [TF2] Pyro Mask/Head Dissapearing when used as disguise by enemy Spy.
When an enemy Spy disguises as the pyro in my team, the mask/head in the Spy's disguise is completely gone. I dont know if the cosmetics used by the Pyro affected it but I will attach pictures of what he was wearing and how the Spy looked with the disguise, it only affected the Spy since the Pyro itself did not present this bug never in the match, and it seems it occurs randomly.
**This is how the mask is entirely gone when the Spy is disguised as the Pyro in my team.**
enemy spy
cosmetics2
cosmetics3

### LINKED ISSUE #7498: [TF2] The Quick-Fix's backpack breaks when Spy disguises
According to the TF Wiki, if a Spy disguises themselves as a Medic that happens to have the Quick-Fix equipped, the backpack will default to the stock Medigun's rather than the Quick-Fix's.

seanmcgeehan:
# Description

This commit solves almost all bugs related to spy disguises. Particularly this solves:

* Wearables for weapons that have "extraWearables". Upon pressing last disguise on the relevant item slot, The spy will receive the wearable until he switches his class disguise even if he switches disguise weapons. This version of the fix increases the skill ceiling of spy while reducing code blast radius. This was leaked checked and extra wearables are cleaned up correctly. 

* Bad body group settings for weapons like the quickfix and vaccinator. In the medigun case this results in correct backpacks being placed for all mediguns for disguised spies. 

* Weapon attachment models. The Kritzkrieg now renders correctly. Festivizers and botkillers should also now work (not tested due to item test limitations)

* Bodygroups are now cleaned up on disguise change and removal. This removes several bugs resulting in headless pyro, footless scout, footless demo. This was tested by spawning a scout with the bat hat which removes the head body group from pyro + the feet bodygroup from demo. I believe body groups from the target player were being transferred over  even if the player wasnt the right class. I also believe disguise body groups weren't cleaned up on change. 

Tested and confirmed:
* The QuickFix
* Battalions Back up
* Buff Banner
* Thermal thruster (Only inactive state, but this allows for additional valid disguises so is an improvement)
* Councheror
* Vaccinator
* Base Jumper
* Kritzkrieg
* Huntsman (only if the target or the spy has a wearable)
*  sydney sleeper (only if the target or the spy has a wearable)
*  Gunslinger (only if the target or the spy has a wearable)
* short circuit  (only if the target or the spy has a wearable) + this should be your last weapon disguise as engi to avoid body group issues
*  GRU + other heavy melees (only if the target or the spy has a wearable) + this should be your last weapon disguise as Heavy to avoid body group issues
* Fixed body groups for pyro head, demo feet, scout feet

None of these conditions are regression and should cover 99% of disguises in game

**Before fix:**
Screenshot (2)
Screenshot 2026-03-04 195150

**After fix:**
Screenshot (10)
Screenshot (12)
Screenshot (11)
Screenshot 2026-03-04 205357
Screenshot 2026-03-04 194425
Screenshot 2026-03-04 194212
Screenshot 2026-03-04 194115
Screenshot 2026-03-04 174719

Testing:

Testing was done in item test for both teams. Swaps were done to other disguise to test against regressions. Leaks were tested for with report_entities. I did not find a regression and this results in many more valid disguises. 

 I got a fix for every visual bug listed on the wiki https://wiki.teamfortress.com/wiki/Disguise_Kit.

closes https://github.com/ValveSoftware/Source-1-Games/issues/275
closes https://github.com/ValveSoftware/Source-1-Games/issues/5883
closes https://github.com/ValveSoftware/Source-1-Games/issues/7498

  Short circuit is a bit of a hack :) 

 These commits are merged to prevent regressions involving bodygroups, since bodygroups are actually processed. Code Hygiene was enforced with the help of Claude. These bug fixes were done to the best of my ability.  A prior version of this PR was here and will be closed. https://github.com/ValveSoftware/source-sdk-2013/pull/1599

### DIFF
--- a/src/game/shared/tf/tf_item_wearable.cpp
+++ b/src/game/shared/tf/tf_item_wearable.cpp
@@ -10,6 +10,7 @@
 #include "tf_gamerules.h"
 #include "animation.h"
 #include "basecombatweapon_shared.h"
+#include "tf_weapon_mechanical_arm.h"
 #ifdef CLIENT_DLL
 #include "c_tf_player.h"
 #include "model_types.h"
@@ -578,6 +579,19 @@ bool CTFWearable::UpdateBodygroups( CBaseCombatCharacter* pOwner, int iState )
 		{
 			// We must use team 0 for disguise weapons.
 			UpdateDisguiseBodygroups( pTFOwner, pDisguiseTarget, pItem, 0, iState );
+
+			CTFMechanicalArm* pMechArm = dynamic_cast<CTFMechanicalArm*>(pDisguiseWeapon);
+			if (pMechArm) {
+				// Hack, Short circuit is special case and should be off if it is the disguise weapon.
+				// Directly set the bodygroup since UpdateDisguiseBodygroups doesn't work for this weapon
+				int iDisguiseBody = pTFOwner->m_Shared.GetDisguiseBody();
+				int iBodyGroup = pDisguiseTarget->FindBodygroupByName("rightarm");
+				if (iBodyGroup != -1)
+				{
+					::SetBodygroup(pDisguiseTarget->GetModelPtr(), iDisguiseBody, iBodyGroup, 2);
+					pTFOwner->m_Shared.SetDisguiseBody(iDisguiseBody);
+				}
+			}
 		}
 	}
 
--- a/src/game/shared/tf/tf_item_wearable.cpp
+++ b/src/game/shared/tf/tf_item_wearable.cpp
@@ -517,6 +517,32 @@ void CTFWearable::ValidateModelIndex( void )
 #endif
 
 //-----------------------------------------------------------------------------
+// Purpose: Helper to apply bodygroups from an item to a disguise target.
+//-----------------------------------------------------------------------------
+void CTFWearable::UpdateDisguiseBodygroups( CTFPlayer *pTFOwner, CTFPlayer *pDisguiseTarget, CEconItemView *pItem, int iTeam, int iState )
+{
+	if ( !pTFOwner || !pDisguiseTarget || !pItem )
+		return;
+
+	int iDisguiseBody = pTFOwner->m_Shared.GetDisguiseBody();
+	int iNumBodyGroups = pItem->GetStaticData()->GetNumModifiedBodyGroups( iTeam );
+
+	for ( int i = 0; i < iNumBodyGroups; ++i )
+	{
+		int iBody = 0;
+		const char *pszBodyGroup = pItem->GetStaticData()->GetModifiedBodyGroup( iTeam, i, iBody );
+		int iBodyGroup = pDisguiseTarget->FindBodygroupByName( pszBodyGroup );
+
+		if ( iBodyGroup == -1 )
+			continue;
+
+		::SetBodygroup( pDisguiseTarget->GetModelPtr(), iDisguiseBody, iBodyGroup, iState );
+	}
+
+	pTFOwner->m_Shared.SetDisguiseBody( iDisguiseBody );
+}
+
+//-----------------------------------------------------------------------------
 // Purpose: Hides or shows masked bodygroups associated with this item.
 //-----------------------------------------------------------------------------
 bool CTFWearable::UpdateBodygroups( CBaseCombatCharacter* pOwner, int iState )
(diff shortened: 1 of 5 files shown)


seanmcgeehan:
This PR Fixes ~almost~ every bug in this video with 300,000 views. The actual bugs were mostly misdiagnosed. They were lack of instrumenting along side not setting items as disguise items in the helper functions

https://www.youtube.com/watch?v=3jUPs_hIhSE

seanmcgeehan:
Revision 2: We have a hack for the short circuit. The huntsman + gunslinger were fixed in the first commit. Short circuit is on its own commit incase you don't want to take the hack. These three weapons require a wearable to be equipped since the fix is instrumented in the wearables bodygroup override. IMO the fix needs to go there since the weapon/econitem bodygroup override is super broken and not an easy fix. If that's the reason for the delayed merge, please communicate with me. I'm willing to spend time fixing any issues with this if it will get it merged. 



Screenshot 2026-03-12 192600
Screenshot 2026-03-12 192612
Screenshot 2026-03-12 095547
Screenshot 2026-03-12 191309

JoriKos:
Does this also fix impossible combinations, i.e., an SMG-wielding Darwin's Danger Shield Sniper?

AgentArtemis:
On behalf of all Spymains, thank you so much.

seanmcgeehan:
@JoriKos No thats separate and would be quite difficult to clean up. I think that can be easily avoided though unlike the ones cleaned up in the patch

JoriKos:
> @JoriKos No thats separate and would be quite difficult to clean up. I think that can be easily avoided though unlike the ones cleaned up in the patch

Yeah, mostly asking for Wiki purposes to know if all the bugs have been removed or if there are still some listed there.

tabbytobias2:
Thanks!

JoriKos:
Relevant issue: https://github.com/ValveSoftware/Source-1-Games/issues/8021 seems like you can cause a server crash by spawning infinite Botkiller attachments when switching disguise weapons.

Koi-TF2:
> Relevant issue: [ValveSoftware/Source-1-Games#8021](https://github.com/ValveSoftware/Source-1-Games/issues/8021) seems like you can cause a server crash by spawning infinite Botkiller attachments when switching disguise weapons.

Reposting here: 
Valve please see @seanmcgeehan's fix: https://github.com/ValveSoftware/source-sdk-2013/pull/1929

SammyROCK:
It seems there is a side effect to this:
https://www.youtube.com/watch?v=whwSbRgAcfQ

bakugo:
@seanmcgeehan was trying to fix the remaining issues in this PR and noticed something I'd like to ask about.

What was the logic behind adding this:
```
#ifdef CLIENT_DLL
	pDisguiseTarget->SetBodygroupsDirty();
#endif
```
In `CTFPlayerShared::DetermineDisguiseWearables()`?

It's dead code, because that function is server-only. And either way, I can't figure out why you would call this on the disguise target instead of the disguising spy.

seanmcgeehan:
@bakugo This was added during the investigation into bodygroups incorrectly persisting between disguises (headless pyro, footless scout). At the time it appeared that this had multiple causes: carrying over between disguises, body groups being fetched  from target on class miss match. I could conclusively prove that the carry over was the primary cause, but i was not convinced that class miss match was not happening. On the deployed commits i aggressively tested to try to replicate bad body groups but could not. Given all of that I did not think it was wise to modify this portion of the fix further as the risk of regression felt high and given the inconsistent and quite nasty nature of the bug the trade of did not feel worth it. in testing I had determined that body groups were constantly recalculated in practice so the overhead was minimal. 

Additionally Bad bodygroup settings blocks everything else in this PR: Extra wearables needs correct body group settings due to conflicts, and correct body group settings can not be transferred between disguises. 

I added you on steam so you can pick my head.

ZMBanshee:
@seanmcgeehan Hey, did you test this with Your Eternal Reward disguises? (disguises from backstabs)

I ask because I've seen a couple of broken bodygroups still after this fix was merged, here's an example today of a Spy that stabbed me using Your Eternal Reward, the bullet belt bodygroup seems to be erroneously disabled on his Heavy disguise (My cosmetics are only Hound Dog and Earbuds, nothing that should disable this bodygroup).

seanmcgeehan:
@ZMBanshee YER was the main case causing bad body groups in general, and although I hadn't tested YER at the time I have exclusively run YER in my games and am certain this fix applies to YER. 

There is a mechanism related to body groups being tied to disguise target that can still happen and can persist through disguise changes, but in game it should be very rare in practice: https://github.com/ValveSoftware/source-sdk-2013/pull/1284 

EDIT: 1284 is likely incompatible with the weapon body group fix as is. This line likely cleans up bodygroups between weapon swaps and for most of them (and the important ones) we want them to stay https://github.com/ValveSoftware/source-sdk-2013/pull/1284/changes#diff-9f282c24791ef0b164ca81859b430e0050e00ef33f637aacdd2cbe5103075e56R13590

Edit2: Just spun up item test and confirmed the fix as written applies to YER. What ZmBanshee iss seeing is likely the bug fixed in 1284.  I was aware of this and the exact fix. I believe Bakugo was going to add this to a future PR so i hadn't bothered cutting it.

PR #1827, opened 2026-03-09
[TF2] Bots will now follow players during mimic taunts

FlaminSarge:
# Description
This implements a 'partner' param for CTFBotTaunt and causes bots to attempt to follow their partner during a taunt.
* Uses AimHeadTowards behavior to stay aiming their body within 10 degrees on either side of the target, and presses Forward button to move within 50 HU of their target's collision boxes. In practice, this ends up being a reasonable follow distance behind the target, though the bots do end up a bit slower than normal players when pressing Forward during taunt for some reason.
* Every 5 seconds (outside of setup time), bots will roll to stop taunting based on `tf_bot_taunt_stop_chance` cvar. (After about 30 seconds at the default of 30%, the chance they've stopped taunting is about 88%, as an example. There are no escalating odds.)
* Bots will also stop taunting if their partner is missing, no longer taunting, or no longer in LOS or 600 range.
* This PR also adds bounds to `tf_bot_taunt_victim_chance` and corrects the check for that cvar, which could previously proc at value 0.
* TauntMove() now looks at Bots' buttons for what magnitude to set flMoveDir, since TauntMove assumes players are using cl_forwardspeed/cl_backspeed which isn't true for bots (they set m_flForwardMove/etc to GetRunSpeed() instead).

Note that the 'target yaw' is set based on a stored var rather than GetCurrentYaw, since AimHeadTowards (and therefore current yaw) may lag behind the target yaw, and so without letting target yaw accumulate, bot turning gets slowed to a crawl. Also note that we have to initialize this after the bot is confirmed as taunting, otherwise the bot tends to snap to some arbitrary angle before properly trying to aim at the player, due to that not being initialized properly.

The taunt turn speed logic is pulled from C_TFPlayer::CreateMove()'s taunt turn speed handling.

Demonstration: soon™

### DIFF
--- a/src/game/server/tf/bot/behavior/tf_bot_taunt.cpp
+++ b/src/game/server/tf/bot/behavior/tf_bot_taunt.cpp
@@ -52,8 +52,7 @@ ActionResult< CTFBot > CTFBotTaunt::Update( CTFBot *me, float interval )
 				return Done( "Taunt finished" );
 			}
 
-			CTFPlayer *partner = m_partner.Get();
-			if ( partner && partner->IsTaunting() && me->CanMoveDuringTaunt() )
+			if ( m_partner != NULL && m_partner->IsTaunting() && me->CanMoveDuringTaunt() )
 			{
 				if ( m_tauntStopCheckTimer.HasStarted() && m_tauntStopCheckTimer.IsElapsed() )
 				{
@@ -80,12 +79,12 @@ ActionResult< CTFBot > CTFBotTaunt::Update( CTFBot *me, float interval )
 				const float maxFacingAngle = 10.f;
 
 				// if outside range of the target, move towards them and turn to face them
-				if ( me->IsRangeGreaterThan( partner, nearbyRange ) )
+				if ( me->IsRangeGreaterThan( m_partner, nearbyRange ) )
 				{
 					// Force move forwards
 					me->PressForwardButton( interval );
 
-					Vector toPartner = partner->GetAbsOrigin() - me->GetAbsOrigin();
+					Vector toPartner = m_partner->GetAbsOrigin() - me->GetAbsOrigin();
 					toPartner.z = 0.f;
 					// diff > maxFacingAngle or < -maxFacingAngle requires turning
 					float diff = UTIL_AngleDiff( UTIL_VecToYaw( toPartner ), m_targetTauntYaw );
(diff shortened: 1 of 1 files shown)


GrubPL:
Wouldn't this possibly let 1 player to basically make a bunch of bots (if not all of them) basically useless until the player stops taunting/the bots manage to die

FlaminSarge:
> Wouldn't this possibly let 1 player to basically make a bunch of bots (if not all of them) basically useless until the player stops taunting/the bots manage to die

Probably. If you have any suggestions on how to curtail that, let me know.
One thing I can think of is a cvar that causes the end timer to start even with the long taunt conditions.

Oddly enough, bots are _not_ set up to override CTFBotTaunt with other actions like 'respond to threats or attacks', so they don't interrupt the taunt behavior for those. Other behaviors _are_ interruptible by higher-priority behaviors. I could probably set the partner taunts to be a lower priority than other behaviors like that.

One other situation I could probably handle is if the bot hasn't moved from its current location in a while (e.g. a player has herded it into a corner) it could stop taunting. I could also just have a hard limit of like 30 seconds on the taunt in general, too.

Someone else suggested LOS or FOV checks.

FlaminSarge:
I've got a pretty good idea: if during setup, let them taunt forever; if not during setup, roll for them to stop the taunt with a cvar-configurable percentage every 5-15 seconds, maybe even with escalating odds.

GrubPL:
> I've got a pretty good idea: if during setup, let them taunt forever; if not during setup, roll for them to stop the taunt with a cvar-configurable percentage every 5-15 seconds, maybe even with escalating odds.

That sounds good to me

FlaminSarge:
Alright this should work out nicely now, added various stop behaviors.

src/game/shared/tf/tf_playeranimstate.cpp
```diff
@@ -380,9 +380,6 @@ void CTFPlayerAnimState::Update( float eyeYaw, float eyePitch )
 	// Compute the player sequences.
 	ComputeSequences( pStudioHdr );
 
-	CTFPlayer *pTauntPartner = pTFPlayer->GetTauntPartner();
-
-	Vector vPositionToFace = ( pTauntPartner ? pTauntPartner->GetAbsOrigin() : vec3_origin );
```
FlaminSarge, line 385: This was entirely unused, and hence cleaned up.

src/game/server/tf/bot/behavior/tf_bot_behavior.cpp
```diff
@@ -571,18 +572,18 @@ EventDesiredResult< CTFBot > CTFBotMainAction::OnOtherKilled( CTFBot *me, CBaseC
 
 		if ( !ToTFPlayer( victim )->IsBot() && me->IsEnemy( victim ) && me->IsSelf( info.GetAttacker() ) )
 		{
-			bool isTaunting = !me->HasTheFlag() && RandomFloat( 0.0f, 100.0f ) <= tf_bot_taunt_victim_chance.GetFloat();
+			bool shouldTaunt = !me->HasTheFlag() && RandomFloat( 0.0f, 100.0f ) < tf_bot_taunt_victim_chance.GetFloat();
```
FlaminSarge, line 575: the `<=` was causing a bug where even at cvar 0 they could taunt

FlaminSarge:
This is mostly ready to go, but ~~one problem I'm running into that I can't quite track down is that the bots are very slightly slower than humans when they're holding Forward on a taunt that they can control movement for. I just can't figure out why that's happening; setting them to hold Forward for longer than 'interval' didn't help, so it's not that they're not holding it for long enough each tick. Setting maxFacingAngle to 0 so they always aim directly at the player also didn't help.~~
EDIT: it looks like this is because humans set their forwardMove to cl_forwardspeed, but bots set it to their run speed which is almost always lower than cl_forwardspeed; TauntMove then does forwardMove/cl_forwardspeed to find what speed to move the player...

PR #1828, opened 2026-03-10
[TF2] Allow bots to realize decloaking spies

inco-cc:
This allows bots to realize a spy that is decloaking. Currently, bots will only realize a spy that is cloaking, due to `TF_COND_STEALTHED` being removed immediately when decloaking.

This replaces the condition check with the new proposed `CTFPlayerShared::IsFullyVisible`, which will check if the spy is either partially or fully cloaked.

Please note that the comparison videos shown below require audio to be unmuted to hear the bot call the player a spy. The bot does not attack the player due to the round being in the setup phase.

Before:

[before.webm](https://github.com/user-attachments/assets/3da3ece1-a1ee-4ac1-bc11-c578688a7bcc)

After:

[after.webm](https://github.com/user-attachments/assets/54b5eb14-e72a-4fb1-b03e-ff159f3f3da7)

### DIFF
--- a/src/game/server/tf/bot/tf_bot_vision.cpp
+++ b/src/game/server/tf/bot/tf_bot_vision.cpp
@@ -277,7 +277,7 @@ bool CTFBotVision::IsIgnored( CBaseEntity *subject ) const
 			return true;
 		}
 
-		if ( enemy->m_Shared.IsStealthed() )
+		if ( !enemy->m_Shared.IsFullyVisible() )
 		{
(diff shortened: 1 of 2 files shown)


a584-droid:
An after and before are just the same recordings. No difference.

Professor-Heavy:
> An after and before are just the same recordings. No difference.

There's sound to it, showing that the bot is aware of the Spy.

inco-cc:
Sorry, I forgot to mention this in the original comment. GitHub mutes videos by default, so you need to unmute them manually. I've edited the comment to explain this.

PR #1829, opened 2026-03-10
[TF2] Fix next-map vote UI becoming unresponsive to mouse input

Mewiof:
# Description

Fixes the bug in Casual mode where the next-map voting panel becomes unresponsive to mouse input (no hover highlights, no clicks) for all players on the server. Players are unable to interact with the vote user interface, but it is still visible, and the timer and percentages are still updating. Using `next_map_vote <0/1/2>` in the console is a known workaround.

https://github.com/user-attachments/assets/ba26226e-2d2b-4052-b83b-e78836955cd7

# Fix
Tested on live Casual servers using memory injection. In a match where the UI became unresponsive for the entire server (confirmed via a party member playing on a stock game client), the patched client worked as expected.

### DIFF
--- a/src/game/client/tf/vgui/tf_matchmaking_dashboard_next_map_voting.cpp
+++ b/src/game/client/tf/vgui/tf_matchmaking_dashboard_next_map_voting.cpp
@@ -87,7 +87,7 @@ public:
 		}
 	}
 
-	virtual bool ShouldBeActve() const OVERRIDE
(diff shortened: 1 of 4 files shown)


Gavitro:
I was worried this wouldn't be able to get fixed through a Github pull, given that it's a problem with Casual Mode. Please, this should be the highest-priority commit on the pull list right now.

pupnoodle:
@Gavitro matchmaking ui code is in the sdk.

PR #704, opened 2025-02-21
Remove knockback-attribute-specific animation overrides for CTFScattergun

FlaminSarge:
Closes https://github.com/ValveSoftware/Source-1-Games/issues/3959 and allows for usage of the attribute in custom weapons without animation breakage.

#### prefabs.weapon_force_a_nature.visuals.animation_replacement
```
diff --git a/tf/scripts/items/items_game.txt b/tf/scripts/items/items_game.txt
index eae0400..cc9a2bd 100644
--- a/tf/scripts/items/items_game.txt
+++ b/tf/scripts/items/items_game.txt
@@ -17977,6 +17977,17 @@

 				"animation_replacement"
 				{
+					"ACT_VM_IDLE"					"ACT_ITEM2_VM_IDLE"
+					"ACT_VM_DRAW"					"ACT_ITEM2_VM_DRAW"
+					"ACT_VM_HOLSTER"				"ACT_ITEM2_VM_HOLSTER"
+					"ACT_VM_PULLBACK"				"ACT_ITEM2_VM_PULLBACK"
+					"ACT_VM_PRIMARYATTACK"			"ACT_ITEM2_VM_PRIMARYATTACK"
+					"ACT_VM_SECONDARYATTACK"		"ACT_ITEM2_VM_SECONDARYATTACK"
+					"ACT_VM_RELOAD"					"ACT_ITEM2_VM_RELOAD"
+					"ACT_VM_DRYFIRE"				"ACT_ITEM2_VM_DRYFIRE"
+					"ACT_VM_IDLE_TO_LOWERED"		"ACT_ITEM2_VM_IDLE_TO_LOWERED"
+					"ACT_VM_IDLE_LOWERED"			"ACT_ITEM2_VM_IDLE_LOWERED"
+					"ACT_VM_LOWERED_TO_IDLE"		"ACT_ITEM2_VM_LOWERED_TO_IDLE"
 					"ACT_PRIMARY_VM_INSPECT_START"		"ACT_ITEM2_VM_INSPECT_START"
 					"ACT_PRIMARY_VM_INSPECT_IDLE"		"ACT_ITEM2_VM_INSPECT_IDLE"
 					"ACT_PRIMARY_VM_INSPECT_END"		"ACT_ITEM2_VM_INSPECT_END"
```

PR #1835, opened 2026-03-13
Fix: scattergun that has knockback cannot be fired on preround

M60TM:
# Description
- During the pre-round (before setup time finishes), the `CTFScattergun` with the `set_scattergun_has_knockback` attribute set to 1 cannot be fired because the code bypasses `BaseClass::FireBullet`. This PR fixes that behavior.

### DIFF
--- a/src/game/shared/tf/tf_weapon_shotgun.cpp
+++ b/src/game/shared/tf/tf_weapon_shotgun.cpp
@@ -318,14 +318,6 @@ void CTFScatterGun::FireBullet( CTFPlayer *pPlayer )
 {
 	if ( HasKnockback() )
 	{
-		// Perform some knock back.
-		CTFPlayer *pOwner = ToTFPlayer( GetPlayerOwner() );
-		if ( !pOwner )
-		{
-			BaseClass::FireBullet( pPlayer );
-			return;
-		}
-
 		// No knockback during pre-round freeze.
 		if ( TFGameRules() && (TFGameRules()->State_Get() == GR_STATE_PREROUND) )
 		{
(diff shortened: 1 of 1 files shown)


JoriKos:
I could be 100% wrong but isn't `if ( !pOwner )` a check that's seeing if pOwner successfully contains a player, and if it doesn't then it shouldn't proceed with anything below it because the game crashes?

bakugo:
That check makes no sense regardless because the `pPlayer` passed into the function is already the owner of the weapon. So `pOwner` can be removed entirely and replaced with `pPlayer`.

JoriKos:
Wouldn't it still, in its current state, cause a crash if it doesn't contain a player?

bakugo:
It will always contain a player, as this was already checked higher up in the call stack:

https://github.com/ValveSoftware/source-sdk-2013/blob/c623a7c30d5cb7275cc64ed0b866f61f4a64c6eb/src/game/shared/tf/tf_weaponbase_gun.cpp#L94-L97

M60TM:
Would it be better to replace `pOwner` with `pPlayer`? `pOwner` and `pPlayer` are used interchangeably in the code.

mastercoms:
Maybe best to just change HasKnockback. Or else some other effects will possibly run in ApplyPostHitEffects and other places.

M60TM:
The codes in FireBullet is intended to prevent movement via knockback before the round starts. Modifying the HasKnockback function could potentially introduce side effects in other parts of the code.

bakugo:
Yes, modifying `HasKnockback()` directly would affect the hardcoded animation overrides in `CTFScatterGun::SendWeaponAnim()`.

Also, it's probably important to note that these changes do not affect the entirety of setup time, only the short freeze time at the start where players cannot move. This check was likely added purely to prevent scouts from boosting themselves forward during that time, and disabling on-hit effects isn't necessary.

FlaminSarge:
> Yes, modifying `HasKnockback()` directly would affect the hardcoded animation overrides in `CTFScatterGun::SendWeaponAnim()`.

The animation overrides were removed in #704, not to say there won't be any other weirdness from having setup time disable HasKnockback directly.

PR #1836, opened 2026-03-13
[TF2] Modify check in `CTFGameRules::GetAssister`

berzerkerbek:
This modifies the existing medic check in `CTFGameRules::GetAssister` to make sure the medic is not the victim which without the check can result in medics getting an assist when they get killed by a disguised spy they were healing

https://github.com/user-attachments/assets/5afb01e1-e6e4-4438-b269-a5b4f579607d


tysonwite:
bad (joke i know him)

seanmcgeehan:
I think the existing behavior is intended?

Professor-Heavy:
Personally, I think that if there are any unintended side effects (such as leaderboard points **which I have not confirmed is the case, this is a hypothetical**) then they could be patched out individually. I personally think the killfeed aspect of this is best left alone if this were to be changed.

It'd even be better if this behavior was preserved, but if the killfeed could reflect teamkill assists correctly by changing the colours. Call it out more explicitly that your own teammate indirectly helped a Spy kill you.

Professor-Heavy:
> It'd even be better if this behavior was preserved, but if the killfeed could reflect teamkill assists correctly by changing the colours. Call it out more explicitly that your own teammate indirectly helped a Spy kill you.

Actually, this has already been marked as a pull request in #1506.

PR #1838, opened 2026-03-14
[HLDMS] Fix incorrectly configured default value for mp_teamlist

ghost:
**ISSUE:** By default, `mp_teamlist` (which sets the amount of teams and their names in Half-Life Deathmatch Source) has the value `hgrunt;scientist`. This will however lead to only one team being created when `mp_teamplay 1` is turned on, as the seperator between team names is supposed to be `,`. See hl1mp_gamerules.cpp

**FIX:** Change the default value `hgrunt;scientist` into `hgrunt,scientist`

### DIFF
--- a/src/game/server/game.cpp
+++ b/src/game/server/game.cpp
@@ -42,7 +42,7 @@ ConVar	flashlight( "mp_flashlight","0", FCVAR_NOTIFY );
 #endif
 ConVar	aimcrosshair( "mp_autocrosshair","1", FCVAR_NOTIFY );
 ConVar	decalfrequency( "decalfrequency","10", FCVAR_NOTIFY );
-ConVar	teamlist( "mp_teamlist","hgrunt;scientist", FCVAR_NOTIFY );
+ConVar	teamlist( "mp_teamlist","hgrunt,scientist", FCVAR_NOTIFY );
 ConVar	teamoverride( "mp_teamoverride","1" );
 ConVar	defaultteam( "mp_defaultteam","0" );
 ConVar	allowNPCs( "mp_allowNPCs","1", FCVAR_NOTIFY );


PiMoNFeeD:
This is actually wrong, because teamplay gamerules still use `;` as separator (although not set from the ConVar itself for some reason), the HL1MP gamerules should ideally just be fixed to match base (and HL1) code behaviour instead by using `;` instead of `,`, so server admins don't have to change the same commands between two games

ghost:
> This is actually wrong, because teamplay gamerules still use `;` as separator (although not set from the ConVar itself for some reason), the HL1MP gamerules should ideally just be fixed to match base (and HL1) code behaviour instead by using `;` instead of `,`, so server admins don't have to change the same commands between two games

Not sure what you mean with this exactly? I wish I could actually make a pull request for the HLDMS gamerules like you said though, but sadly the code is not in here

PiMoNFeeD:
> > This is actually wrong, because teamplay gamerules still use `;` as separator (although not set from the ConVar itself for some reason), the HL1MP gamerules should ideally just be fixed to match base (and HL1) code behaviour instead by using `;` instead of `,`, so server admins don't have to change the same commands between two games
> 
> Not sure what you mean with this exactly? I wish I could actually make a pull request for the HLDMS gamerules like you said though, but sadly the code is not in here

What I mean is, it's better to fix the syntax to match base gamerules and HL1, so you have the same command working the same way in both games. What I was referring to about teamplay gamerules, is it has its own team creation logic which does in fact use `;` instead of `,`, so there's no point in dividing the behaviour, which is unfortunately out of the scope of the SDK like you said, but if this ever somehow gets pulled into HL1DM, there's no reason to duct-tape it like this and instead it should just be fixed properly by changing the separator from `,` to `;`

PR #1839, opened 2026-03-14
[HLDMS] Fix all map entities being spawned twice

ghost:
**ISSUE:** For years, Half Life Deathmatch Source has had this bug where all entities in a .bsp file are spawned twice. E.g. 1 prop_physics entity in hammer will end up becoming 2 prop_physics in the exact same spot in game.

This is LITERALLY game breaking. Any map that is more than a static battlefield is broken. Even stock maps have dysfunctional doors, elevators, etc. because having 2 spawn in the same place breaks them.

**FIX:** The issue is that `CServerGameDLL::LevelInit_ParseAllEntities` in hl1mp_gameinterface.cpp calls `MapEntity_ParseAllEntities()` when it already gets called in `CServerGameDLL::LevelInit` (non mod-specific part of the SDK). It can be fixed by placing the call for `LevelInit_ParseAllEntities()` behind an `#ifndef HL1MP_DLL` statement.

### DIFF
--- a/src/game/server/gameinterface.cpp
+++ b/src/game/server/gameinterface.cpp
@@ -1051,7 +1051,9 @@ bool CServerGameDLL::LevelInit( const char *pMapName, char const *pMapEntities, 
 		g_pServerBenchmark->StartBenchmark();
 
 		// Now call the mod specific parse
-		LevelInit_ParseAllEntities( pMapEntities );
+		#ifndef HL1MP_DLL
+			LevelInit_ParseAllEntities( pMapEntities );
+		#endif
 	}
 
 	// Check low violence settings for this map


1157:
I hope it gets pushed after 13+ years.

YourSourceBoiii:
@misyltoad Please merge this

You are the most important person on the PLANET right now for the HLDMS community. All you have to do to save our souls is clicking the accept button
something

Artem13327:
> **FIX:** The issue is that `CServerGameDLL::LevelInit_ParseAllEntities` in hl1mp_gameinterface.cpp calls `MapEntity_ParseAllEntities()` when it already gets called in `CServerGameDLL::LevelInit` (non mod-specific part of the SDK). It can be fixed by placing the call for `LevelInit_ParseAllEntities()` behind an `#ifndef HL1MP_DLL` statement

This is a problem that needs to be fixed not in the SDK, but in the game itself.

ghost:
> > **FIX:** The issue is that `CServerGameDLL::LevelInit_ParseAllEntities` in hl1mp_gameinterface.cpp calls `MapEntity_ParseAllEntities()` when it already gets called in `CServerGameDLL::LevelInit` (non mod-specific part of the SDK). It can be fixed by placing the call for `LevelInit_ParseAllEntities()` behind an `#ifndef HL1MP_DLL` statement
> 
> This is a problem that needs to be fixed not in the SDK, but in the game itself.

That would be cleaner yes but sadly I can't make a pull request for the HLDMS game files here. This achieves the exact same thing though and I'd rather have it be fixed like this than not having it fixed at all

PR #1841, opened 2026-03-14
[TF2] Put other games' impulse commands behind preprocessor directive

ghost:
**ISSUE:** Honestly, what's the point of having all these game specific `impulse` console commands in the engine-wide code? `impulse 82` spawns the HL2 jeep for example, but HL2 is the only game that has those.

They should be put behind corresponding `#ifdef` statements if you ask me

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -6300,6 +6300,7 @@ void CBasePlayer::CheatImpulseCommands( int iImpulse )
 
 	switch ( iImpulse )
 	{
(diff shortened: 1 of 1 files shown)


JoriKos:
Would this prevent spawning an airboat in Portal? It's a critical part of that game :^)

ghost:
> Would this prevent spawning an airboat in Portal? It's a critical part of that game :^)

It wouldn't because Portal players use the `ch_createairboat` command, which is unaffected here

PR #1845, opened 2026-03-19
[TF2] Fix attachments on dropped weapons having broken textures for certain war paints / skins

Piogre:
Dropped weapons that have attachments on them can have broken textures on the attachments if the weapons are Decorated/War Painted with skins that use "material_override" -- including War Paints using the Macaw Masked materials (colloquially referred to as "albedo tint" skins).  The override gets applied to the attachments when the weapon is dropped on death, causing the textures to break.

This fixes that issue for dropped weapons.

This issue appears most prominently for festivizers on weapons with these skins, but also breaks the pilot lights on flamethrowers, and the screens on rescue rangers.

ft
rr

Tangentially, this fix removes the only obstacle to adding festivizer support to the Golden Wrench (which also uses material_override to apply its unique skin) -- this bug is the only reason the Golden Wrench wasn't in the "easy adds" pile for festivizers so with this fix it can be implemented in its items_game definition ("169") with the standard "can_be_festivized" tag and "attached_models_festive" block in the visuals section (models/weapons/c_models/c_wrench/c_wrench_festivizer.mdl can be reused for this with no changes)

gw

### DIFF
--- a/src/game/shared/tf/halloween/tf_weapon_spellbook.cpp
+++ b/src/game/shared/tf/halloween/tf_weapon_spellbook.cpp
@@ -3006,7 +3006,9 @@ public:
 				pEntity->SetGroundChangeTime( gpGlobals->curtime + 0.5f );
 			}
 
-			pEntity->Teleport( NULL, NULL, &vecVelocity );
+			pEntity->SetAbsVelocity( vecVelocity );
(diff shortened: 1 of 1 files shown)


Piogre:
wrong source branch, will re-request

PR #1846, opened 2026-03-19
[TF2] Fix Lightning Ball Spell constantly phasing players in and out of Purgatory

### LINKED ISSUE #4492: [TF2] Hitting a player with the Ball o' Lightning spell in an Underworld spams "has escaped the underworld" message
https://youtu.be/-9-JROY3sJM as shown in this clip. 

This works on multiple maps, additionally it also applies the Übercharge and healing effect that leaving the underworld normally applies.

Piogre:
Currently, a player hit with the Lightning Ball spell while in Purgatory will repeatedly gain and lose the TF_COND_PURGATORY condition for as long as they're in the spell's effective radius, which will cause them to stay ubered and kritzed for the duration, while the whole server is constantly spammed with the "escaped the underworld" chat alert and bell sound.

This is because the lightning ball spell calls the "teleport" function on players inside its radius, passing only a velocity vector, to continually re-apply an absolute velocity and draw those players toward its center.  Mechanically, it's only setting velocity on the player, but because it's technically teleporting them it has other side effects, including apparently re-triggering trigger_add_tf_player_condition entities they're inside of and thus removing/re-applying the condition.  This is noticeable for the TF_COND_PURGATORY condition, which applies the uber, kritz, server-wide sound alert and chat alert upon its removal.

Having the Lightning Ball spell set the velocity directly instead of unnecessarily teleporting the player (calling the same functions teleport would call for velocity) fixes this issue.

See below demonstration on pd_pit_of_death_event (sound on).

https://github.com/user-attachments/assets/5f64a350-32a2-4f34-980f-16713f375a1e

edit: fixes ValveSoftware/Source-1-Games#4492

### DIFF
--- a/src/game/shared/tf/halloween/tf_weapon_spellbook.cpp
+++ b/src/game/shared/tf/halloween/tf_weapon_spellbook.cpp
@@ -3006,7 +3006,9 @@ public:
 				pEntity->SetGroundChangeTime( gpGlobals->curtime + 0.5f );
 			}
 
-			pEntity->Teleport( NULL, NULL, &vecVelocity );
+			pEntity->SetAbsVelocity( vecVelocity );
+			pEntity->SetBaseVelocity( vec3_origin );
+
 		}
 
 		SetContextThink( &CTFProjectile_SpellLightningOrb::VortexThink, gpGlobals->curtime + 0.2f, "VortexThink" );


FlaminSarge:
It's worth investigating why Teleport causes the trigger to fire to add or remove the condition repeatedly (it's probably baseentity.cpp TeleportEntity setting solid flags to not-solid...)

bakugo:
@FlaminSarge That is indeed the cause, adding `FSOLID_NOT_SOLID` to the player during entity simulation makes all their touches end until the next frame.

I've hit this bug before in sourcemod plugins and I'd like to see it properly fixed, though I'm not sure what the proper fix actually is. I assume the solid change is there for a reason, but maybe not (it's been there since the HL2 beta).

<details>
<summary>Detailed explanation of what I believe is happening</summary>

- Game frame starts
- `Physics_RunThinkFunctions()` runs, simulates all entities and runs their thinks in order
- Players come first, so the player's touch links and their `touchStamp`s are all updated and valid for this frame
- Then the spell projectile (`CTFProjectile_SpellLightningOrb`) thinks, which calls `TeleportEntity()` on the player
- `TeleportEntity()` calls `AddSolidFlags( FSOLID_NOT_SOLID )`, which then calls `CCollisionProperty::SetSolidFlags()`
- `SetSolidFlags()` sees that the player's `FSOLID_NOT_SOLID` changed, calls `CCollisionProperty::CheckForUntouch()`
- `CheckForUntouch()` sees that the player is not solid, calls `CBaseEntity::SetCheckUntouch()`
- `SetCheckUntouch()` increments player's `touchStamp`, invalidating all touch links, and schedules an untouch check
- `Physics_RunThinkFunctions()` ends
- `FrameUpdatePostEntityThinkAllSystems()` runs, calls scheduled `PhysicsCheckForEntityUntouch()` on player
- `PhysicsCheckForEntityUntouch()` sees that the player's touch links are invalid, ends the touches
- At the start of the next frame, when the player is simulated again, the touches will start again

If `TeleportEntity()` is called outside of this (I first tried testing by creating a command to teleport), the bug won't happen.
</details>

Piogre:
There's a solid case for reworking the teleport function to not do this, but since the lightning orb doesn't actually need to teleport to achieve its effect in the first place, I didn't want to go diving into a fairly core function that's used elsewhere, potentially causing changes to behavior elsewhere in the game, just to fix a quirk of one holiday-specific feature.

FlaminSarge:
> I'm not sure what the proper fix actually is

A bit hacky but maybe it'd be sufficient to have Teleport not do the unsolid thing if the location vector is null? I'm assuming it does all of this so that the entity being teleported makes all triggers forcefully recalculate touching (which is causing the issue with the underworld trigger since it doesn't expect untouch+quick retouch but the 'recalculate' case expects to do exactly that). I assume addcond/etc triggers also weren't built with that recalculate behavior in mind.

Piogre:
If one were to rework the teleport function to avoid retriggers we'd probably want something a bit more robust than that -- there exist short-range teleports that WOULD set a location vector but potentially stay within the same addcond trigger volume  and we'd have the same problem.

Again for the purpose of this fix I didn't want to go making changes to core engine code to fix a random feature that doesn't need to touch that code at all (since the lightning ball spell's function is not intended to teleport) and risk changing behavior elsewhere -- A "real" solution would likely involve retaining a list of prior trigger volumes for the recalculation to compare against -- something I'm not going to risk breaking for a halloween spell that doesn't need to teleport at all.

PR #1852, opened 2026-03-24
[TF2] Fix Half-Zatoichi honor kill being survivable with damage reduction runes

### LINKED ISSUE #6974: [TF2] It's possible to survive the Half-Zatoichi one-shot

PhiTiet:
Fixes ValveSoftware/Source-1-Games#6974

The honor-bound one-shot uses `GetHealth() * 3`. The Resist rune's 0.5x multiplier applied downstream can reduce this below lethal, especially when stacked with other reductions. Increased to `* 10`.


JoriKos:
It's what I initially did on #1691 except with `* 5`. Currently working on a solution to properly add a new damage type for it and check for that instead, as simply bypassing the resistance is much cleaner than just upping the damage by a ton.

PhiTiet:
@JoriKos Your exact suggestion is implemented correctly now: the Half-Zatoichi honor-bound duel kill uses a dedicated custom damage type and cleanly bypasses resist-rune reduction instead of inflating the damage value.

JoriKos:
Now it's pretty much the same as my PR, but it doesn't bypass regular damage resistance. Issue is that basically copying #1691 isn't very productive unless it superceeds it.

FlaminSarge:
This is not the right approach, and this PR is effectively a duplicate of #1691.
I've opened https://github.com/JoriKos/source-sdk-2013/pull/1 to make improvements to the implementation of #1691.

Also please disclose if this PR was AI-generated.

FlaminSarge:
This may be superseded by #1881

PR #1857, opened 2026-03-26
[TF2] Attempt to expose functionality for turning a team into robots outside of Man vs Machine

TheShermanTanker:
# Description
Much of the code that turns every player (Fake or not) on a given team, though really just Blue in practice, into a robot is completely inaccessible to custom game modes. With the exception of calling SetCustomModelWithClassAnimations with a robot model and SetBloodColor(DONT_BLEED) in VScript, there is nothing that one can do to turn a team into robots, since that logic is completely locked behind hardcoded IsMannVsMachineMode calls. The result is that custom game types which want to use the robots have to do many extremely ugly hacks just to implement the rest of the robot details beyond the robot models and setting don't bleed to true, for instance, custom MVM missions where you play as Blue and fight robots on Red, which does not have any robot support from the game whatsoever, or something like Stop that Tank, or maybe a server owner just wants robots in casual and nothing else. But even with those hacks, a lot of things don't work, like robots still bleeding like a human when hit, and so on. Right now, the only way to accomplish this is to switch on the CNetworkVar m_bPlayingMannVsMachine but that enables a whole lot of extra Man vs Machine specific logic that you may not want, doesn't do anything in the case of when you want to turn a non\-Blue team into robots, and obviously cannot be switched on outside of actual Man vs Machine.

It would be helpful to custom modes and modders if this functionality is exposed somehow without all the extra Man vs Machine logic, for now just turning players into robots (Though enabling certain things like allowing the sapper to sap players but with a cooldown could be explored in a followup), ideally in a convenient "Setting this applies the changes across the whole game". This current implementation does this by making robot teams a ConVar (Though see below, I do not intend to merge this as is). Setting it to 0 means neither team are robots, 1 means both are, and a value matching TF\_TEAM\_BLUE and TF\_TEAM\_RED means the corresponding team will be robots.

The implementation here is only meant to get a discussion started on how best to do this. It is a complete wreck, and I do not intend to merge it in this state whatsoever. It is heavily broken and the robots do not make any sound at all, no speech, no damage sounds, not even death sounds (Their weapons and footsteps and impact sounds do work though), and when they gib the gibs don't appear at all, among other problems. I also don't know if I got everything either. The code in CTFPlayer::ForceRespawn is also just placeholder code so I can see the robot models during testing, I will leave whether that should be kept or whether it should be removed and modders should just call the available VScript functions up to discussion. I also am taking bikeshedding suggestions on whether to make the switch for enabling this a CNetworkVar, keep it as a ConVar or do something else, I don't know which way is best at the moment. I'm hoping to make the robot features easier for people and modders who want to access them in the future, try not to laugh at the terrible code :)

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -1244,7 +1244,7 @@ void C_TFRagdoll::OnDataChanged( DataUpdateType_t type )
 					CreateTFHeadGib();
 					EmitSound( "TFPlayer.Decapitated" );
 
-					bool bBlood = true;
+					bool bBlood = !( pPlayer && IsRobotTeam( pPlayer->GetTeamNumber() ) );
 					if ( TFGameRules() && ( TFGameRules()->UseSillyGibs() || 
 											( TFGameRules()->IsMannVsMachineMode() && pPlayer && pPlayer->GetTeamNumber() == TF_TEAM_PVE_INVADERS ) ) )
 					{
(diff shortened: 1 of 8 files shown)


GrubPL:
It's Mann VS machine
Not man

JoriKos:
> It's Mann VS machine 
> Not man

It's actually 'Mann vs. Machine'.

PR #1858, opened 2026-03-28
[TF2] Fix overheal particles revealing dead ringer feigns

kunring:
# Current behaviour

When a spy feigns while emitting overheal particles, emission of new overheal particles stop, but the existing particles follow the spy until their end of life. This is one of the most universal "dead ringer tells" in the game, since especially with the kunai, spies are quote often overhealed. Anyone who knows to look for it can see the spy's position for a second after the feign.

https://github.com/user-attachments/assets/3b6fe1b9-a63f-456d-b18c-f8f3746b6588

# Proposed behaviour

The spy's overheal particles are immediately destroyed upon feigning.

https://github.com/user-attachments/assets/aac0e7aa-54f0-4aa2-b74c-bb86d89003f0

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -10041,7 +10041,14 @@ void C_TFPlayer::UpdateOverhealEffect( void )
 	{
 		if ( m_pOverHealedEffect )
 		{
-			ParticleProp()->StopEmission( m_pOverHealedEffect );
+			if ( m_Shared.InCond( TF_COND_FEIGN_DEATH ) )
+			{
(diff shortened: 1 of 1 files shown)


Ashe-tf:
The healing particles aren't just killed after dying, they stop emitting but the crosses remain for a while

https://github.com/user-attachments/assets/43846c4b-a5f1-476e-a4fa-e08f71845e6f

kunring:
> The healing particles aren't just killed after dying, they stop emitting but the crosses remain for a while

This is a possible alternate behaviour, but in any case, if the spy dies to low damage while emitting heal particles, the enemy can tell it must have been a feign. Continuing to show the particles makes it easier to check this after the feign. I would assume the particles being killed will be generally hard to notice, so I don't think it matters that much. I think both ways are within reason.

PR #1860, opened 2026-03-28
Fix Scout LongDistanceRunner achievement not counting distance on round end

qequ:
# Description

The CAchievementTFScout_LongDistanceRunner achievement ("Race for the Pennant") has a bug where distance progress is lost when a round ends without the player dying.

The FireGameEvent_Internal handler already has a branch for teamplay_round_win, but the event was never registered in ListenForEvents()  making it dead code since the initial SDK commit. The m_fMetersRan counter resets on respawn (c_tf_player.cpp:7953), so any distance accumulated during a life that ends via round win/loss is silently discarded.

Fix
Added ListenForGameEvent( "teamplay_round_win" ) to ListenForEvents() so the existing handler is actually reached.


FlaminSarge:
If the round ends and then the player dies, does it end up double-counting the current life's distance?

PR #1861, opened 2026-03-29
[TF2] Fix Huntsman arrows not being extinguished in water

dilbertron2:
This PR makes ignited Huntsman arrows become extinguished when you go underwater, bringing this interaction in line with others of a similar nature. (Players being extinguished when going underwater as well as the Pyro's flamethrower not working underwater)

https://github.com/user-attachments/assets/44cde043-7f60-4b43-9b01-968713a12a77


a584-droid:
Should you then make Huntsman impossible to use underwater probably?

PR #1866, opened 2026-03-30
[TF2] Fix Sydney Sleeper extinguishing teammates on unscoped hits

dilbertron2:
# Description
This change brings consistency to how the Sydney Sleeper interacts with both friendlies and enemies alike.
As Jarate is only applied to enemies upon scoped hits it also makes sense for extinguishing to only take place under the same conditions.

# Video
https://github.com/user-attachments/assets/3e3a05c6-bf5b-48f4-af27-b7b6bff5487b

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -10348,7 +10348,7 @@ void CTFPlayer::FireBullet( CTFWeaponBase *pWpn, const FireBulletsInfo_t &info, 
 				// put out fire for burning teammate
 				if ( nCustomDamageType == TF_DMG_CUSTOM_PENETRATE_NONBURNING_TEAMMATE )
 				{
-					if ( pTargetPlayer && GetTeamNumber() == pTargetPlayer->GetTeamNumber() && pTargetPlayer->m_Shared.InCond( TF_COND_BURNING ) && ToTFPlayer(GetActiveWeapon()->GetOwner())->m_Shared.InCond( TF_COND_AIMING ))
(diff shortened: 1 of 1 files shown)


src/game/shared/tf/tf_player_shared.cpp
```diff
@@ -10348,7 +10348,7 @@ void CTFPlayer::FireBullet( CTFWeaponBase *pWpn, const FireBulletsInfo_t &info,
 				// put out fire for burning teammate
 				if ( nCustomDamageType == TF_DMG_CUSTOM_PENETRATE_NONBURNING_TEAMMATE )
 				{
-					if ( pTargetPlayer && GetTeamNumber() == pTargetPlayer->GetTeamNumber() && pTargetPlayer->m_Shared.InCond( TF_COND_BURNING ) )
+					if ( pTargetPlayer && GetTeamNumber() == pTargetPlayer->GetTeamNumber() && pTargetPlayer->m_Shared.InCond( TF_COND_BURNING ) && ToTFPlayer(GetActiveWeapon()->GetOwner())->m_Shared.InCond( TF_COND_AIMING ))
```
bakugo: The entire `ToTFPlayer(GetActiveWeapon()->GetOwner())->` part can be removed. Don't know why Valve also did that in the `ExtinguishPlayer` call instead of just using `this`.

dilbertron2: Noted, Thanks.

wgetJane:
i'd say this is a balance change rather than a fix

JoriKos:
I would argue it's a fix more than a balance change, as the Meet Your Match patch notes specifically states scoped shots extinguishing teammates. Unscoped shots are never mentioned anywhere to intentionally extinguish teammates. I personally think the weapon is better off as-is, though. It's not a great unlock and this bug and/or feature doesn't make it overpowered and isn't harmful.

PR #1869, opened 2026-04-01
[TF2] Add convar to prevent voice commands being printed in chat

ghost:
**ISSUE:** On certain gamemodes and servers, it can be quite annoying to constantly get spammed with voice commands being subtitled in chat. `(Voice) shitman gaming 123: Spy!`, etc. being done by people on the other side of the map or as a joke. It would be nice to give players the ability to prevent these messages from being printed to their chat.

**FIX:** Add in a **client side** convar named `cl_mute_voice_commands_text`.
This convar is set to 0 by default, allowing new players and less technically inclined people to still see those voice command subtitles in the chat.
When `cl_mute_voice_commands_text 1` is set, none of the voice command subtitles get printed into your chat, allowing you to experience more peace of mind.
When `cl_mute_voice_commands_text 2` is set, only calls for a MEDIC will be printed to chat. A special case is made for these as medic calls are useful for medics to determine where somebody is on the map (seeing both their name in chat and seeing the balloon pop up on your screen)
disabling-voice-commands-in-text-chat-v0-ddzsn4x8ktw11

### DIFF
--- a/src/game/client/hud_basechat.cpp
+++ b/src/game/client/hud_basechat.cpp
@@ -39,6 +39,7 @@ ConVar cl_showtextmsg( "cl_showtextmsg", "1", 0, "Enable/disable text messages p
 ConVar cl_chatfilters( "cl_chatfilters", "63", FCVAR_CLIENTDLL | FCVAR_ARCHIVE, "Stores the chat filter settings " );
 ConVar cl_chatfilter_version( "cl_chatfilter_version", "0", FCVAR_CLIENTDLL | FCVAR_ARCHIVE | FCVAR_HIDDEN, "Stores the chat filter version" );
 ConVar cl_mute_all_comms("cl_mute_all_comms", "1", FCVAR_ARCHIVE, "If 1, then all communications from a player will be blocked when that player is muted, including chat messages.");
+ConVar cl_mute_voice_commands_text("cl_mute_voice_commands_text", "0", FCVAR_ARCHIVE, "Disable voice commands being printed in chat when performed. 0 = All voice commands get printed, 1 = No voice commands get printed, 2 = Only allow MEDIC! calls to be printed.");
 ConVar cl_enable_text_chat( "cl_enable_text_chat", "1", FCVAR_ARCHIVE, "Enable text chat in this game" );
 
 const int kChatFilterVersion = 1;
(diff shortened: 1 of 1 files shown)


naydef:
It would be even better if there's in addition a server-side convar for this. (Not that there's a plugin blocking voice commands in chat already)

dilbertron2:
In my opinion a nice addition would be to have Spy! calls also be a part of `cl_mute_voice_commands_text 2` as Spy! and MEDIC! calls are similarly as important, in the heat of combat it can be easy to miss a Spy! callout if its not printed to chat.

Gavitro:
I think a more efficient solution would be an option to filter which specific voice commands are muted. I would argue "Sentry Ahead!", "Help!" and "Incoming!" are useful, but adding them all to cl_mute_voice_commands_text 2 might be cumbersome.

ghost:
> I think a more efficient solution would be an option to filter which specific voice commands are muted. I would argue "Sentry Ahead!", "Help!" and "Incoming!" are useful, but adding them all to cl_mute_voice_commands_text 2 might be cumbersome.

What do you think would be a better solution then? I kinda wanted to keep it simple to increase the chances of Volvo merging this, but if you have a better idea I'm willing to add it in

FlaminSarge:
This should also be integrated with the chat "Filters" button as "Voice Commands" before being merged, as an additional entry alongside e.g. Team Changes/Achievement announcements. The cvar should be a part of cl_chatfilters instead of its own cvar.

See https://github.com/ValveSoftware/source-sdk-2013/blob/3300848d8a25ef6403c91f82a4cd97d6daefbc06/src/game/client/hud_basechat.h#L63 and https://github.com/ValveSoftware/source-sdk-2013/blob/3300848d8a25ef6403c91f82a4cd97d6daefbc06/src/game/client/hud_basechat.cpp#L499, should be simple enough to add.

If necessary, two separate filters (one for non-critical and one for critical voice commands) can work.

ghost:
> This should also be integrated with the chat "Filters" button as "Voice Commands" before being merged, as an additional entry alongside e.g. Team Changes/Achievement announcements. The cvar should be a part of cl_chatfilters instead of its own cvar.
> 
> See
> 
> https://github.com/ValveSoftware/source-sdk-2013/blob/3300848d8a25ef6403c91f82a4cd97d6daefbc06/src/game/client/hud_basechat.h#L63
> and
> 
> https://github.com/ValveSoftware/source-sdk-2013/blob/3300848d8a25ef6403c91f82a4cd97d6daefbc06/src/game/client/hud_basechat.cpp#L499
> , should be simple enough to add.
> 
> If necessary, two separate filters (one for non-critical and one for critical voice commands) can work.

Thanks. I will close this PR and make a new one which does it like that instead if nobody else will do it before me

GrubPL:
They deleted their account I think?

I don't think we're getting a new PR from them any time soon
Especially not a new version of this PR, it'd be neat if someone made it, just so we'd even have a chance of it being merged

PR #1870, opened 2026-04-01
[HLDMS] Fix broken "impulse 76" command

ghost:
**ISSUE:** The `impulse 76` console command is broken as it tries to spawn in a `NPC_human_grunt` entity, which doesn't exist in Half life Deathmatch Source or Half Life Source

**FIX:** Change `NPC_human_grunt` into `monster_human_grunt`, which does exist


GrubPL:
I wonder why is it even a thing in the first place
Like the impulse itself, not the bug

tschumann:
Probably a testing thing from back before the original Half-Life was
released.

PR #1872, opened 2026-04-02
[TF2] Fix incorrect coloring on death animation when dying to friendly Cow Mangler

dilbertron2:
# Description
When a player dies to the Cow Mangler, their ragdoll enters a dissolving/fizzling animation using the color of the opposing team. This behavior also occurs even if the player is dying to friendly Cow Mangler shots. This is because the function uses the color of the player's opposing team without any further checks.

This PR creates a new variable called ~`m_bIsSuicide`~ `m_bIsFriendlyFireOrSuicide` and utilises the `player_death` GameEvent to determine whether or not the death was caused by friendly fire/suicide. This variable is then incorporated into the ragdoll dissolving function.

# Video
https://github.com/user-attachments/assets/5bb16b83-dbbc-4c09-90da-b31ef6cb9a3b

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -1598,29 +1598,16 @@ void C_TFRagdoll::DissolveEntity( CBaseEntity* pEnt )
 
 		Vector vColor;
 		C_TFPlayer *pLocalPlayer = C_TFPlayer::GetLocalTFPlayer();
+		bool bIsSuicide = pLocalPlayer ? pLocalPlayer->IsDeathSuicide() : NULL;
 		// If a player kills themselves with a weapon that dissolves ragdolls, set dissolving color to their own team
 		if ( m_iTeam == TF_TEAM_BLUE )
 		{
-			if (pLocalPlayer)
-			{
-				vColor = ( pLocalPlayer->IsDeathSuicide() ) ? TF_PARTICLE_WEAPON_BLUE_1 * 255 : TF_PARTICLE_WEAPON_RED_1 * 255;
-			}
-			else // Fallback to original code if pLocalPlayer is invalid
-			{
-				vColor = TF_PARTICLE_WEAPON_RED_1 * 255;
-			}
+			vColor = ( bIsSuicide ) ? TF_PARTICLE_WEAPON_BLUE_1 * 255 : TF_PARTICLE_WEAPON_RED_1 * 255;
 			pDissolve->SetEffectColor( vColor );
 		}
 		else
 		{
-			if (pLocalPlayer)
-			{
-				vColor = ( pLocalPlayer->IsDeathSuicide() ) ? TF_PARTICLE_WEAPON_RED_1 * 255 : TF_PARTICLE_WEAPON_BLUE_1 * 255;
-			}
-			else // Fallback to original code if pLocalPlayer is invalid
-			{
-				vColor = TF_PARTICLE_WEAPON_BLUE_1 * 255;
-			}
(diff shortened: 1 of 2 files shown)


dilbertron2:
Unsure if I have over-engineered this PR or not, there may be a simpler way to check if the player has killed themselves that I am unaware of.

src/game/client/tf/c_tf_player.h
```diff
@@ -675,6 +676,7 @@ class C_TFPlayer : public C_BasePlayer, public IHasAttributes, public IInventory
 	bool			m_bSaveMeParity;
 	bool			m_bOldSaveMeParity;
 	bool			m_bIsCoaching;
+	bool			m_bIsSuicide = false;
```
ficool2: Initialize in constructor to follow existing conventions

src/game/client/tf/c_tf_player.cpp
```diff
@@ -11182,6 +11199,23 @@ void C_TFPlayer::FireGameEvent( IGameEvent *event )
...
 		}
 	}
+	else if (FStrEq(event->GetName(), "player_death" ) )
+	{
+		m_bIsSuicide = false;
+		const int iAttacker = engine->GetPlayerForUserID(event->GetInt("attacker"));
+		C_TFPlayer* pAttacker = ToTFPlayer(UTIL_PlayerByIndex(iAttacker));
+
+		const int iVictim = engine->GetPlayerForUserID(event->GetInt("userid"));
+		C_TFPlayer* pVictim = ToTFPlayer(UTIL_PlayerByIndex(iVictim));
```
ficool2: Follow Valve's styling (spaces between parentheses)

src/game/client/tf/c_tf_player.cpp
```diff
@@ -1597,14 +1597,30 @@ void C_TFRagdoll::DissolveEntity( CBaseEntity* pEnt )
...
 		{
-			vColor = TF_PARTICLE_WEAPON_RED_1 * 255;
+			if (pLocalPlayer)
+			{
+				vColor = ( pLocalPlayer->IsDeathSuicide() ) ? TF_PARTICLE_WEAPON_BLUE_1 * 255 : TF_PARTICLE_WEAPON_RED_1 * 255;
+			}
+			else // Fallback to original code if pLocalPlayer is invalid
+			{
+				vColor = TF_PARTICLE_WEAPON_RED_1 * 255;
+			}
```
ficool2: Rather than writing it like this, you can make a bool like so:
bool bSuicide = pLocalPlayer ? pLocalPlayer->IsDeathSuicide() : NULL;

and then you can remove the else here and use that bool in the ternary

Permik:
How does this interact with the friendly fire option? Do teammates killed by friendly fire get fizzled with correct colour in this instance?

This seems like it should just be handled "correctly" by checking the associated team of the projectile.

dilbertron2:
> How does this interact with the friendly fire option? Do teammates killed by friendly fire get fizzled with correct colour in this instance?
> 
> This seems like it should just be handled "correctly" by checking the associated team of the projectile.

Completely forgot that friendly fire is a thing! I'll look into it..

FlaminSarge:
Might be better to call it m_bAttackerSameTeam or something similar, for brevity (since self is same-team too).

For the sake of mods that add more teams, it may even be better to just store the attacker's team number instead, so that mods can color the particles appropriately even if they're e.g. green team.

Storing the var on the player seems incorrect in general for this, however. Maybe a param on DissolveEntity instead?

PR #1880, opened 2026-04-09
Optimize UTIL_CreateScaledPhysObject

RaphaelIT7:
# Description
I made this back in like 2024, though it should be merged upstream for everyone's benefit :)
This optimizes UTIL_CreateScaledPhysObject by switching it to use CPhysPolysoup which is way faster.
I've suggested this change for GMod around a week ago and has been working great there on the dev branch.

This is an old performance test from when I made it since I **can't** test it with the sdk when the engine keeps crashing :/
Though even in GMod it was confirmed to be significantly faster.
image

Test with the same 100 PhysCollide:
Old took ~880ms
New takes ~140ms

### DIFF
--- a/src/game/server/props.cpp
+++ b/src/game/server/props.cpp
@@ -6044,7 +6044,7 @@ bool UTIL_CreateScaledPhysObject( CBaseAnimating *pInstance, float flScale )
 		physcollision->DestroyQueryModel( pQuery );
 
 		// Create a collision model from all the convexes
-		pNewCollide = physcollision->ConvertPolysoupToCollide( pPolySoups, true );
+		pNewCollide = physcollision->ConvertPolysoupToCollide( pPolySoups, false );
 		if ( pNewCollide == NULL )
 			return false;
 	}
--- a/src/game/server/props.cpp
+++ b/src/game/server/props.cpp
@@ -6016,7 +6016,7 @@ bool UTIL_CreateScaledPhysObject( CBaseAnimating *pInstance, float flScale )
 
 		// Create a container to hold all the convexes we'll create
 		const int nNumConvex = pQuery->ConvexCount();
-		CPhysConvex **pConvexes = (CPhysConvex **) stackalloc( sizeof(CPhysConvex *) * nNumConvex );
+		CPhysPolysoup *pPolySoups = physcollision->PolysoupCreate();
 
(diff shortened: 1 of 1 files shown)


PiMoNFeeD:
`ConvertPolysoupToCollide`'s 2nd argument should be `false` instead, because `ConvertConvexToCollide` does not use MOPP. Also, pretty sure MOPP path isn't even compiled in the binary?
Having it "true" instead technically works for regular VPhysics (because its unused, looking at decomp), but breaks VPhysics Jolt which doesn't implement it and returns NULL instead for that path

robotboy655:
After a trial by fire in recent Garry's Mod update, I cannot recommend this change, as it seems to cause severe stability issues, namely infinite loops in the physics engine post-scale, as well as possibly severe performance degradations with scaling certain props. (related?)

Scaling itself seems to be very error prone when scaling models above 10x.
`models/props_c17/statue_horse.mdl` from Half-Life 2 for example crashes `vphysics` on collision with scales above 2x on Windows with this change.

On Linux SRCDS within Garry's Mod, [gm_sentimental98v1](https://steamcommunity.com/sharedfiles/filedetails/?id=3308701371) enters infinite loop in the physics simulation with this code present.

RaphaelIT7:
Turns out- in like 1 month of testing, I don't think anyone encountered any issues with it? (there were some on 64x GMod though there an very old vphysics version was used, which was assumed to be the reason for it not working), and 1 day after the update release, many found weird issues 😭
As Rubat mentioned and as seen with how it went with GMod I'll make this PR a draft until vphysics becomes stable enough or some solution for those issues is found. This definitely has potential yet in its current state it also brings a lot of destructive issues.
And as I can't test it with this SDK state anyways, it'll be annoying to figure anything out and back then when I had tested it, it also was with a different vphysics version which probably contibuted to this becoming such an failure.

PiMoNFeeD:
Can confirm after my own testing that this change causes performance issues: if applied to static props, when a player moves into a scaled static prop, FPS drops by at least 10 times, VPROF explodes so much that both "Prediction" and "Player" don't go below full bars at all, and FPS drops from 300+ down to ~20, and VS profiling shows that it's somewhere in IVP collision trace code

I wonder what polysoup breaks interally so badly..?

PR #1881, opened 2026-04-10
[TF2] Prevent resistance buffs from stopping Half-Zatoichi duel instant kill

FlaminSarge:
# Description
Rework of #1691
This adds checks to prevent damage reduction (from Buff Banner or Powerups) from preventing the Honorbound duel insta-kill.
It does this by creating a new damage type flag `DMG_IGNORE_RESIST_BUFFS` (=`DMG_PHYSGUN`), setting it when the Katana instant kill is handled and reading it where conditions apply their damage resistances.
I used `DMG_PHYSGUN` because it was the least likely to cause side effects (melee damage sets its damage forces on CTakeDamageInfo manually, so Damage_NoPhysicsForce won't take effect even if DMG_PHYSGUN is present in the damage info).
I avoided `DMG_DIRECT` because it is also used by EntityFlame, so using that would cause map-created legacy fire damage (torches on Egypt, as an example) to no longer be affected by resistances, which is technically a regression. I'll leave it to Valve to determine of DMG_DIRECT is a better representation/usage here and if that regression is acceptable, though.
I avoided reusing `DMG_DONT_COUNT_DAMAGE_TOWARDS_CRIT_RATE` as I expected mods to make use of that damage type much more often than other types.

Tested against the case where the victim had Resist powerup + Battalion's Backup buff.

Originally based off changes by @JoriKos

Note that the change in tf_shareddefs is due to Sourcetree very much disliking the special characters in that comment and refusing to commit until they were removed, for some reason.

### DIFF
--- a/src/game/shared/tf/tf_gamerules.cpp
+++ b/src/game/shared/tf/tf_gamerules.cpp
@@ -6488,9 +6488,7 @@ bool CTFGameRules::ApplyOnDamageModifyRules( CTakeDamageInfo &info, CBaseEntity 
 		}
 
 		// Reduce damage if not resist-piercing weapon or honorbound duel
-		if ( !iAttackIgnoresResists && !( pTFAttacker && pWeapon
-			&& pWeapon == pTFAttacker->GetActiveTFWeapon() && pWeapon->IsHonorBound()
-			&& pVictim->GetActiveTFWeapon() && pVictim->GetActiveTFWeapon()->IsHonorBound() ) )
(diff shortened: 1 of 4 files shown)


FlaminSarge:
@JoriKos Your implementation with the new DamageCustom value was pretty close, tbh. Adding a damage type flag was way cleaner than doing the Honorbound victim+attacker checks everywhere.

JoriKos:
Yeah this makes more sense, though it is not something I would've probably figured out on my own due to a lack of experience.

PR #1889, opened 2026-04-14
[TF2] Fix Pomson 6000 projectile offset

hankjordan:
# Description

The Pomson 6000's projectile offset is defined incorrectly, causing the projectile to be fired below the crosshair. This PR modifies the offset so the projectile works as expected.

### Before

https://github.com/user-attachments/assets/74625163-7920-46f4-882c-0647dcda7534

### After

https://github.com/user-attachments/assets/88daeffc-6422-4039-a27d-c89b5d196c48


kvog-git:
How does this compare to #1247 ?

Also, I think your AI assistant accidentally created a git submodule in your PR:
image

hankjordan:
> How does this compare to #1247 ?
@kvog-git 

Thanks! That's my fault, I must have been working from a dirty branch.

It doesn't differ significantly from #1247 other than the fact that the first person angle is slightly different (and there's 10 lines less churn). As mentioned in that PR, this is consistent with rocket launcher behavior. I figure this diff will be less controversial.

PR #1891, opened 2026-04-17
[ALL] Expose VPhysics related functions

ALIEN31ITA:
# Description
This Pull Request goal is to make Physics functions exposed to vscript, allowing mappers and scripters to experiment with any entity they wish, by turning them Physical, override their material properties, changing mass, making them float on water and so on.

All the functions should also work on brush entities too! (tested with CreatePhysics, and Setmass for now

### API
| Function | Info |
| :--- | :--- |
| `void CreatePhysics( int nSolidType, int nSolidFlags, bool asleep )` | Create the entity Physics. |
| `void DestroyPhysics()` | Destroy the entity Physics. |
| `bool HasPhysics()` | Check if the entity has Physics. |
| `int GetSurfaceProperty()` | Returns the ID of the entity's surfaceprop |
| `string GetSurfacePropertyName(string name)` | Returns the ID of the name of the entity's surfaceprop |
| `void SetSurfaceProperty( int surfaceprop )` | Sets the ID of the entity's surfaceprop |
| `void SetSurfacePropertyByName( string name )` | Sets the ID of the entity's surfaceprop by name |
| `void GetMass()` | N/A |
| `float SetMass( float mass )` | N/A |
| `float SetBuoyancy( float buoyancy )` | Sets the Buoyancy of the Entity, on a 0-1 ratio, 0 = sinks, 1 = fully float |
| `void GetElasticity()` | N/A |
| `float SetElasticity( float elasticity )` | N/A |
| `void ToggleCollisionsOn( handle entity, bool Enable )` | Toggles collisions between 2 objects like logic_collision_pair, must use false, true only for colliding them again. |
| `void SetPhysicsFlag( int PhysFlag )` | Sets a Physics Flag to the entity |
| `bool HasPhysicsFlag()` | checks a Physics Flag on the entity, usefull for detecting if 2 props are stuck into each other for example. |
| `void RemovePhysicsFlag( int PhysFlag )` | Remove a Physics Flag on the entity. |

# Showcase (of the Most important ones)
**sv_turbophysics was turned off for demonstrating this in TF2.**
It's kinda hard showcasing demostrating most of them, but i recorded the ones i find to be the most important, and a must have. 
## MakePhysics(6,0,false)
This is really usefull if you want to convert stuff to have physics, could be an NPC, or specific objects you wish, without requiring you to spawn a prop_physics (or anything physical) and parent it, should work with most entities.

https://github.com/user-attachments/assets/53666bd6-5978-4455-bf39-d0de096a47a6
## GetMass & SetMass

https://github.com/user-attachments/assets/255c2baa-6293-4bb5-8ba7-f52ec044f5dd
## ToggleCollisionsOn( used a loop in the example )
The Closest we have is using a logic_collision_pair, or setting the entities collisiongroup to COLLISION_GROUP_BREAKABLE_GLASS, but this gives us filtering, so we decide what to collide with or not (may not work on some entities).

https://github.com/user-attachments/assets/f3e17a17-78fd-4983-a06c-4f9bd43863f5
## GetSurfacePropertyByName & SetSurfacePropertyByName
Let's you override their physical properties without the need of recompiling existing models.

https://github.com/user-attachments/assets/34ac9050-6e98-4300-b09f-456bcef1a85c

I will try Expose more functions, i only found these to be high priority, and easier to make accessible, if anyone has suggestions, or functions that SHOULD be exposed, or want to improve it, let me know.


src/game/server/baseentity.h
```diff
@@ -1340,6 +1345,180 @@ class CBaseEntity : public IServerEntity
...
+	void ScriptSetElasticity( float flElasticity ) 
+	{ 
+		SetElasticity( flElasticity );
+	}
+
+	float ScriptGetElasticity ( void ) const
+	{
+		return m_flElasticity; 
+	}
+
```
practicemedicine, line 1521: Although it is fine but you don't really need to make an wrapper for `GetElasticity` and `SetElasticity` since you can just directly define them as VScript functions. Just saying.

practicemedicine (approved):
I made a little comment around L1512-L1521 regarding how VScript functions are defined, but overall this looks good to me.

src/game/server/baseentity.h
```diff
@@ -1340,6 +1345,180 @@ class CBaseEntity : public IServerEntity
...
+	int ScriptGetSurfaceProperty( void ) const
+	{
+		IPhysicsObject * vPhys = VPhysicsGetObject();
+		if ( vPhys )
+		{
+			return vPhys->GetMaterialIndex();
+		}
+		else
+		{
+			Log_Warning( LOG_VScript, "Entity (%s) has no VPhysics, make sure it does, or use the CreatePhysics function\n" );
```
ChaseTownsend: Missing GetDebugName for format

SpookyToad2:
Definitely a must

PR #1895, opened 2026-04-18
[TF2] Fix Demoman Charge Meter

ThatFriendlyWolf:
Currently, in live TF2, the charge meter for Demoman does not factor in the stat that makes it only mini crit (for the Tide Turner).
Alongside this it does not properly reflect the color range properly. 

This pull request tries to fix this so that it is a lot more reliable to use.
Changing it so that the charge meter does not display the crit colors for the Tide Turner, while also fixing the percentage to line up with what is actually used.

shield fix

### DIFF
--- a/src/game/client/tf/tf_hud_demomanpipes.cpp
+++ b/src/game/client/tf/tf_hud_demomanpipes.cpp
@@ -66,7 +66,7 @@ CHudDemomanPipes::CHudDemomanPipes( const char *pElementName ) : CHudElement( pE
 
 	SetHiddenBits( HIDEHUD_MISCSTATUS | HIDEHUD_PIPES_AND_CHARGE );
 
-	vgui::ivgui()->AddTickSignal( GetVPanel(), 100 );
+	vgui::ivgui()->AddTickSignal( GetVPanel(), 50 );
 
 	m_bChargeMode = false;
 	m_flOldProgress = 1.f;
@@ -186,9 +186,6 @@ void CHudDemomanPipes::OnTick( void )
 				{
 					m_pChargeMeter->SetFgColor( Color( 153, 255, 153, 255 ) );
 				}
-				
-				// We halve the update ticks so that the meter is mostly reliably synced for the colors
-				vgui::ivgui()->AddTickSignal( GetVPanel(), 50 );
 			}
 			else
 			{
@@ -207,8 +204,6 @@ void CHudDemomanPipes::OnTick( void )
 				{
 					m_flOldProgress = flProgress;
 				}
-
-				vgui::ivgui()->AddTickSignal(GetVPanel(), 100);
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


ThatBapGuy:
Forgot to add the third caption to the GIF, whoops. The third one is the video of the fixed Tide Turner charge.

src/game/client/tf/tf_hud_demomanpipes.cpp
FlaminSarge, line 192: While it's recharging, do you think it's worth setting the color to grey so that it has a visible color change when it's full (grey to white)?

ThatFriendlyWolf: Seeing how the other meters do not have this type of effect _(other than the vaccinator charge which is a entirely different meter)_ I believe it would be best to keep it as is to keep it consistent with the other meters that are used.

FlaminSarge: Yeah, it should probably be made a feature for _all_ meters at once, rather than just this one, if it were to be changed.

SolarLightTF2:
█▀▀░█░█░█▀█░█▀▀░█▀█
▀▀█░█░█░█▀▀░█▀░░█▀▄
▀▀▀░▀▀▀░▀░░░▀▀▀░▀░▀
█▄░▄█░█▀▀░█▀▀▀░ █▀█
█░█░█░█▀░░█░▀█░█▀▀█
▀░▀░▀░▀▀▀░▀▀▀▀░▀░░▀
█░░█░█░▀▀█▀▀░█▀█░ █▀█
█░░█░█░░░█░░░█▀▄░█▀▀█
▀▀▀▀░▀▀▀░▀░░░▀░▀░▀░░▀
█▀▀░█▀█░█░█▀▀
█▀░░█▀▀░█░█
▀▀▀░▀░░░▀░▀▀▀
░░░░░░██▄
░░░░░░█░█
░░░░░█░░█
█▄▄▄█░░░██
█░░ Valve ░░░█
█░ Add This ░░█
█░ Please! ░░░█
█░ And Give ░░█
█░ me one ! ░░█
███████████

FlaminSarge:
Separately, I wonder if there's merit in having some way to tune what percentage/charge duration the crit stages advance. E.g. allowing longer charge durations to still get crits at the default timings (via cvar or something).

misyltoad:
I should look at this when off holiday. I have opinions and I need to write them up =)

torzod:
> Separately, I wonder if there's merit in having some way to tune what percentage/charge duration the crit stages advance. E.g. allowing longer charge durations to still get crits at the default timings (via cvar or something).

if this is to be added, it is important to note that charge duration can be extended by fully refilling the charge meter mid-charge if you are holding right-click when that happens. this is possible by hitting a resupply cabinet brush mid-charge, or by picking up metal with the persian persuader mid-charge.

this would indirectly slightly buff the persian persuader, as often when charging into a skirmish you will pick up a dead player's ammo pack, bringing your charge meter out of the crit range, and reducing your damage at the last second.

PR #1896, opened 2026-04-18
[TF2] Fix Shields Not Restoring Charge With Pickups When Mid Charge

### LINKED ISSUE #4722: [TF2] [Feature Request] Disable picking up ammo kits granting charge while charging for the persian persuader.
Currently a Demoman equipped with the Persian Persuader can pick up ammo kits while charging with a shield weapon, this is an issue as the Persian persuaders stat of granting charge for picking up ammo kits interferes with the crit bonus gained by charging by reseting the bonus timer.

ThatFriendlyWolf:
The stat that gives shield charge with ammo pickups (_Persian Persuader_) do not accurately restore the charge meter, as it essentially does nothing if you are mid charge when collecting. In the end, causing you to lose the charge you would have gained. This PR fixes this alongside makes it align accurately if you charge into a resupply locker.

This PR basically adjusts the clock on the charge from when you get a pickup or resupply.

Resolves https://github.com/ValveSoftware/Source-1-Games/issues/4722

### DIFF
--- a/src/game/server/tf/entity_ammopack.cpp
+++ b/src/game/server/tf/entity_ammopack.cpp
@@ -111,7 +111,7 @@ bool CAmmoPack::MyTouch( CBasePlayer *pPlayer )
 				// Update the shield charge condition to reflect the ammo pack increase
 				if ( pTFPlayer->m_Shared.InCond( TF_COND_SHIELD_CHARGE ) )
 				{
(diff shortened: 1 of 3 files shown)


torzod:
this does remove nuance in the persian persuader's gameplay, and is a slight balance change. as mentioned in #1895, the persian persuader *can* extend your charge, your charge meter just needs to hit 100% full and you need to be holding mouse2 at the moment that happens. this gives the player more control over their charge than other swords, as if there is metal on the ground, you can plan a charge that has much longer than normal range or choose for it to end at its normal length, depending on which would be preferred in a given situation.

PR #1904, opened 2026-04-28
[TF2] building damage feedback fixes

sopranoguitar:
# Description

fixes an issue where health sent in damage feedback on buildings is rounded down. this would cause damage feedback to incorrectly play the last hitsound on buildings on occasion. this also fixes an issue where the damage feedback sound would not be played if the entity was removed on the client before the event was handled (though the visible damage number still remains unfixed in this regard)

unfixed:

https://github.com/user-attachments/assets/ac73d02a-9fd3-4cb1-bd1a-b5b3fb2a54ba

fixed:

https://github.com/user-attachments/assets/923d6cf8-bab8-4e4f-a2b6-5fabaaba4314


Platina6978:
The killsound playing incorrectly happens quite a lot if you regularly use continuous fire weapons (such as a minigun) on buildings that are in the process of being built. Speaking from experience here.

Very nice fix!

PR #1906, opened 2026-05-02
[ANY?] Double the hud channels 6 -> 12

naydef:
# Description

Allowing to show more than 6 simultaneous hud text (as long as this is the right code i'm changing here).

Mainly intended for Sourcemod plugins.

This was not tested nor verified to compile without issues.


pizzaperp95:
did you really just submit an untested pr?? LMAOOO

naydef:
Probably not the first nor the last time someone doesn't test changes.
I guess I'll get back to this at later point when I can test it (this PR is lower priority).

PR #1907, opened 2026-05-02
[TF2] Allow forcing TF2 style hud messages even if cl_hud_minmode is 1

naydef:
# Description

These look nice, but cannot be shown to players in min mode. 

This necessitate a fallback code in SM plugins or repeating the same information with the other options (chat, hudText).

So make it possible to force them visible irrespective of value of `cl_hud_minmode` convar.

### DIFF
--- a/src/game/client/tf/tf_hud_notification_panel.cpp
+++ b/src/game/client/tf/tf_hud_notification_panel.cpp
@@ -101,7 +101,7 @@ void CHudNotificationPanel::MsgFunc_HudNotify( bf_read &msg )
 	// Ignore notifications in minmode
 	if ( !bForceShow )
 	{
-		ConVarRef cl_hud_minmode( "cl_hud_minmode", true );
+		static ConVarRef cl_hud_minmode( "cl_hud_minmode", true );
 		if ( cl_hud_minmode.IsValid() && cl_hud_minmode.GetBool() )
 			return;
 	}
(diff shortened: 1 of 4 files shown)


mastercoms:
Better to add a force parameter.

naydef:
> Better to add a force parameter.

I've considered doing it the way HudMsg works, but lack of experience.
But I guess i could make it that way:
```c++
	int iBackgroundTeam = msg.ReadByte();
	
	bool bForceShow = false;
	if ( msg.GetNumBytesLeft() > 0 )
	{
		bForceShow = msg.ReadByte();
	}

	// Ignore notifications in minmode
	if ( !bForceShow )
	{
		ConVarRef cl_hud_minmode( "cl_hud_minmode", true );
		if ( cl_hud_minmode.IsValid() && cl_hud_minmode.GetBool() )
			return;
	}

	SetupNotifyCustom( szText, szIcon, iBackgroundTeam );
```

mtxfellen:
With this approach, `game_text_tf` would benefit from a keyvalue to control force showing.

naydef:
> With this approach, `game_text_tf` would benefit from a keyvalue to control force showing.

Hopefully I did it right.

naydef:
I made some searching and there's a reported issue https://github.com/ValveSoftware/Source-1-Games/issues/4498 that this PR should fix.

Also I found that [mastercoms](https://github.com/mastercoms) already did something similar in https://github.com/mastercomfig/tf2-patches-old/commit/86dd47ab15f3691b45484d4ca551a10e11cb39fd. I think I'll embed some of the changes there.

PR #1910, opened 2026-05-03
[TF2] Fix Thermal Thruster ignoring `tf_rocketpack_cost` when checking IsRocketPackReady

ChaseTownsend:
If you modify the `tf_rocketpack_cost` convar to change how much charge a launch uses, the `IsRocketPackReady` function never updates, and always checks for if the value is enough for a default charge


ChaseTownsend:
i have no idea how i closed this

PR #1912, opened 2026-05-03
[TF2] Update Default FOV from 75 -> 90 for freshly generated configurations.

blackletum:
This is a resubmission of a [previous PR](https://github.com/ValveSoftware/source-sdk-2013/pull/1000), per request.

Original PR text:

# Description
Self explanatory but most new player guides almost immediately suggest changing the fov from 75 -> 90 as it is a direct benefit to do so.

### DIFF
--- a/src/game/shared/tf/tf_gamerules.h
+++ b/src/game/shared/tf/tf_gamerules.h
@@ -1036,7 +1036,7 @@ private:
 
 	void ChooseNextMapVoteOptions();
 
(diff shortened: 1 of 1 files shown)


Pygargue00fr:
If this is ever actually changed, as every tf2 player's been requested more than a decade, it would be nice for the weapon fov to also be changed to 70 by default, which is what most players also use.

PR #1913, opened 2026-05-04
[TF2] Restore code for the Flame Rockets

BreavyTF2:
Restores code for the Flame Rockets, and let's the Flamethrower shoot Rockets if the projectile override is set to Flame Rockets. Code is reverse-engineered from the June 2007 prototype, with minor alterations to make it work in modern TF2.

https://github.com/user-attachments/assets/1e403a49-4a65-4f27-96e7-1d59f90a3772

Requires the Flame Rocket trail particles to be added.
[incendiaryrockettrail.zip](https://github.com/user-attachments/files/27367251/incendiaryrockettrail.zip)


GrubPL:
Is this like based off of leaked/reverse engineered code?

Edit: oh yeah it is... I don't think valve would be happy with that

PR #1914, opened 2026-05-05
Add `_rt_FullFrameDepth` as `IMAGE_FORMAT_NV_INTZ`

Akabenko:
### Description
INTZ colored hardware depth can be as analogue of Resolved Depth buffer from `SSAO_DepthPass` for moders. Without extra rendering of scene.
We can see scene on our screen.
Image
And hardware colored INTZ depth on `_rt_FullFrameDepth` render target.
Image

### Explanation of code
- [`PushRenderTargetAndViewport` ](https://github.com/ValveSoftware/source-sdk-2013/blob/3300848d8a25ef6403c91f82a4cd97d6daefbc06/src/public/materialsystem/imaterialsystem.h#L1461) allow to write Hardware depth as second arg, but we need to input both args. First as colored rendertarget and second depth render target. We can't skip first argument.

- When we run `PushRenderTargetAndViewport` with colored and depth args, scene will gone from screen. So we need to return that, and we call `pRenderContext->CopyTextureToRenderTargetEx(0, pColor, NULL, NULL);` to write image to `BackBuffer` from colored texture.


Akabenko:
But there is still a problem.

No 3D skybox
image

Activated 3D skybox
image

Akabenko:
This code helps, but I think there must be a more elegant solution.
```cpp
if (g_CurrentViewID == VIEW_MAIN) {
	pRTColor = GetFullscreenTexture();
	pRTDepth = GetFullFrameDepthTexture();
}
```

[viewrender.cpp#L4994](https://github.com/ValveSoftware/source-sdk-2013/blob/3300848d8a25ef6403c91f82a4cd97d6daefbc06/src/game/client/viewrender.cpp#L4994)
```cpp
void CSkyboxView::Draw()
{
	VPROF_BUDGET( "CViewRender::Draw3dSkyboxworld", "3D Skybox" );

	ITexture *pRTColor = NULL;
	ITexture *pRTDepth = NULL;

	if (g_CurrentViewID == VIEW_MAIN) { // extra code
		pRTColor = GetFullscreenTexture();
		pRTDepth = GetFullFrameDepthTexture();
	}

	if( m_eStereoEye != STEREO_EYE_MONO )
	{
		pRTColor = g_pSourceVR->GetRenderTarget( (ISourceVirtualReality::VREye)(m_eStereoEye-1), ISourceVirtualReality::RT_Color );
		pRTDepth = g_pSourceVR->GetRenderTarget( (ISourceVirtualReality::VREye)(m_eStereoEye-1), ISourceVirtualReality::RT_Depth );
	}

	DrawInternal(VIEW_3DSKY, true, pRTColor, pRTDepth );
}
```

https://github.com/user-attachments/assets/ac976957-072e-4a9c-a260-de2270c095a1

Akabenko:
Seems INTZ kill MSAA

HyperionCoding:
What's the motivation for this?

Akabenko:
> What's the motivation for this?

This allow to get depth buffer without second render of scene like as do Resolved Depth Buffer in SSAO_DepthPass function. This can be useful for modders on Source SDK 2013 for deferred renderer. Thats why INTZ kill MSAA.

https://www.dmitrex.com/publications/Enhancing%20the%20looks%20of%20Source%20Engine%20in%20Military%20Conflict%20Vietnam.pdf
image

neobenedict:
Just change the pyrovision check instead so SSAO_DepthPass always runs if there's something in the scene that needs it. Instead of hardcoding to only have pyrovision on?

Akabenko:
But its not a SSAO_DepthPass. INTZ depth is cheaper to get, than Resolved Depth.

neobenedict:
I know, I was suggesting since the feature's already implemented, it could be made more "available", which is potentially a lower bar to clear than this.

Also it's only 8-bit in the red buffer, other channels should be used to encode more data...

Akabenko:
@neobenedict you can run GetRenderTargetEx in view.cpp `void CViewRender::Init( void )` for _rt_ResolvedFullFrameDepth with IMAGE_FORMAT_R32F. And then your depth will be as 32 bit. Then use it as depth * 4000 = projPosW in shaders. or as 1/depth/4000 as depth in range 0...1

Akabenko:
We can modify DepthWrite to writes WorldNormals in other channels, but here no Normals info in meshes in Brushes. Its as optimization of render of shadow maps and ssao depth pass. U can get Normals of props and reconstruct Normals for World.
533283244-bf424ad7-28bd-4f43-ab3b-a6a31ea5bfac

So you can get normals from Models and make 3Tap reconstruction for world.

PR #1917, opened 2026-05-11
[TF2] Fix Escape handling in item selection dialogs

SanyaKor:
# Description

While playing TF2 casual, I noticed an issue with backpack combo overlay dialogs. Here is a video of the issue:

https://github.com/user-attachments/assets/cbb47f0e-f9d0-430a-9d02-f763f67e496e

And if u play around with that it can gets even worse. 

https://github.com/user-attachments/assets/ecc68888-1574-4fb9-aaba-ddfaf1ffd75e

The Select Style window did not have complete Escape handling, especially for the combo box dropdown state. 

I compared this behavior with `CConfirmDialog` in `game/client/econ/confirm_dialog.cpp`. Like `CComboBoxBackpackOverlayDialogBase`, `CConfirmDialog` is also based on `vgui::EditablePanel`. `CConfirmDialog` handles Escape by overriding the `EditablePanel` keyboard input method, `OnKeyCodeTyped`, and routing it through its normal cancel command:

```cpp
void CConfirmDialog::OnKeyCodeTyped( vgui::KeyCode code )
{
	if ( code == KEY_ESCAPE )
	{
		OnCommand( "cancel" );
	}
}
```

`CComboBoxBackpackOverlayDialogBase` did not fully handle the same case. Adding Escape handling to the base overlay fixes the case where the overlay itself has focus. However, `vgui::ComboBox` opens an internal `vgui::Menu` for its dropdown, and that menu takes focus while the dropdown is open. Because of that, Escape never reaches the parent overlay while the dropdown is active.

This fix handles both focus paths:

- `CComboBoxBackpackOverlayDialogBase` now maps Escape to `OnCommand("cancel")`, similar to `CConfirmDialog`.
- The combo box dropdown uses a small custom menu that catches Escape while the dropdown has focus.
- That menu posts a `ComboBoxEscape` action signal back to the parent overlay.
- The parent overlay handles `ComboBoxEscape` by calling the same `OnCommand("cancel")` path.

Here is the result after the fix:

https://github.com/user-attachments/assets/6b02d05c-fc25-4f63-8c07-6dee98e513bf

Now Escape consistently closes the Select Style dialog whether focus is on the overlay itself or inside the opened combo box dropdown. The existing style selection/apply logic is unchanged.

This base combo overlay is not only used by Select Style (`CStyleSelectDialog`). It is also used by the refurbish/restoration dialog (`CRefurbishItemDialog`) and by the strange part restriction selection dialog (`CSelectStrangePartToRestrictDialog`). The refurbish dialog can hit the same issue: for example, try restoring/removing a customization from a Strange weapon that has more than one Strange part, open the dropdown, and press Escape. The dropdown can close while the parent overlay remains open, which can leave the UI in the same awkward state.

Example of issue with restoration

https://github.com/user-attachments/assets/ede6e3f8-e329-4de9-b789-144452c9d615

I did not find an in-game issue with the strange part restriction flow (`CSelectStrangePartToRestrictDialog`) during testing, but it shares the same base dropdown behavior. I do not expect forwarding Escape through the normal cancel path to break it. If a specific derived dialog ever needs different Escape behavior, it can override the relevant handler with an empty or custom implementation.

### DIFF
--- a/src/game/client/econ/item_style_select_dialog.cpp
+++ b/src/game/client/econ/item_style_select_dialog.cpp
@@ -20,6 +20,36 @@
 #include <tier0/memdbgon.h>
 
 
+
+
+
+class CStyleComboMenu : public vgui::Menu
+{
+	DECLARE_CLASS_SIMPLE(CStyleComboMenu, vgui::Menu);
+
+public:
+	CStyleComboMenu(vgui::Panel* pParent, const char* pszName)
+		: BaseClass(pParent, pszName)
+	{
+	}
+
+	virtual void OnKeyCodeTyped(vgui::KeyCode code)
+	{
+		if (code == KEY_ESCAPE)
+		{
+			SetVisible(false);
+			PostActionSignal(new KeyValues("ComboBoxEscape"));
+			return;
+		}
+
+		BaseClass::OnKeyCodeTyped(code);
+	}
+};
+
+
+
+
+
 //-----------------------------------------------------------------------------
 // Purpose:
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


SanyaKor:
Note: I also found another related input issue. If you open Select Style or the refurbish/restoration dropdown, press Escape to return to gameplay, and then press the chat/team chat key (`U` in my test), the UI input state can break badly. After that, backpack, loadout, and team selection inputs may stop responding, and in the worst case the game cannot be closed normally without Alt+F4. This looks related to the same missing/incorrect Escape handling leaving the UI in an inconsistent focus state.


https://github.com/user-attachments/assets/66a69998-02ba-47d2-8ad9-64d233dafb47

PR #984, opened 2025-03-10
fix: Fix the sandman voice lines not playing.

Trico-Everfire:
# Description
The Scout has voice lines related to stunning another player with the sandman ball.
These lines no longer trigger as Scout no longer stuns the enemy player's controls which it was specifically checked for. This was originally triggered by [tf_player_shared.cpp](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/game/shared/tf/tf_player_shared.cpp#L9805)'s StunPlayer function. The trigger for the voice lines has now been moved into the `ApplyBallImpactEffectOnVictim` function to not break any existing logic related to the original StunPlayer function.

| Before | After |
| -------- | ------- |
| <video src="https://github.com/user-attachments/assets/c6b19e6c-5941-4af5-8642-f3f14328aba9"> | <video src="https://github.com/user-attachments/assets/24d26692-46f8-4d3e-906b-781a3e339dd2"> |

This would fix issue:
https://github.com/ValveSoftware/Source-1-Games/issues/4467

PR #1919, opened 2026-05-12
[TF2] Re-add stun effects for The Sandman (BONK! icon, conc stars, voicelines)

### LINKED ISSUE #4467: [TF2] Special voice-lines refuse to play upon getting stunned by the Sandman
https://user-images.githubusercontent.com/98559994/198858998-233c8584-e0fc-41c2-aa0a-19566323cc2a.mp4

Before the Jungle Inferno update, a special voice-line would play upon stunning somebody with the Sandman's ball. 
However, ever since it got reworked in said update, voice-lines don't play at all when you stun (slow) an enemy.

haaksirikko:
# Description
Restores the effects (BONK! icon, conc stars, voicelines) of the Sandman that made it iconic without changing the weapon's functionality otherwise.

Adds a new define `TF_STUN_EFFECTS` for the unused stun flag `TF_STUN_DODGE_COOLDOWN` (once used for the original Bonk's post-effect slowdown), which now makes the effects appear

Supersedes #984, fixes https://github.com/ValveSoftware/Source-1-Games/issues/4467

Vanilla:

https://github.com/user-attachments/assets/0a6c0d9d-4898-4442-b840-756fe93e85bc

This PR (attacker):

https://github.com/user-attachments/assets/11bc39f0-ffa2-4477-a819-6249bc08b3f1

Victim:

https://github.com/user-attachments/assets/db97405b-288e-4b20-85be-fca679523fb2

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -7340,41 +7340,42 @@ void CTFPlayerShared::OnAddOverhealed( void )
 //-----------------------------------------------------------------------------
 void CTFPlayerShared::OnAddStunned( void )
 {
-	if ( IsControlStunned() || IsLoserStateStunned() )
+	if ( GetActiveStunInfo() )
 	{
+		int iStunFlags = GetStunFlags();
 #ifdef CLIENT_DLL
-		if ( GetActiveStunInfo() )
+		if ( !m_pOuter->m_pStunnedEffect && !( iStunFlags & TF_STUN_NO_EFFECTS ) &&
+			 ( iStunFlags & ( TF_STUN_CONTROLS | TF_STUN_LOSER_STATE | TF_STUN_EFFECTS ) ) )
 		{
-			if ( !m_pOuter->m_pStunnedEffect && !( GetActiveStunInfo()->iStunFlags & TF_STUN_NO_EFFECTS ) )
+			if ( ( iStunFlags & TF_STUN_BY_TRIGGER ) || InCond( TF_COND_HALLOWEEN_BOMB_HEAD ) )
 			{
-				if ( ( GetActiveStunInfo()->iStunFlags & TF_STUN_BY_TRIGGER ) || InCond( TF_COND_HALLOWEEN_BOMB_HEAD ) )
-				{
-					const char* pEffectName = "yikes_fx";
-					m_pOuter->m_pStunnedEffect = m_pOuter->ParticleProp()->Create( pEffectName, PATTACH_POINT_FOLLOW, "head" );
-				}
-				else
-				{
-					const char* pEffectName = "conc_stars";
-					m_pOuter->m_pStunnedEffect = m_pOuter->ParticleProp()->Create( pEffectName, PATTACH_POINT_FOLLOW, "head" );
-				}
+				const char* pEffectName = "yikes_fx";
+				m_pOuter->m_pStunnedEffect = m_pOuter->ParticleProp()->Create( pEffectName, PATTACH_POINT_FOLLOW, "head" );
+			}
+			else
+			{
+				const char* pEffectName = "conc_stars";
+				m_pOuter->m_pStunnedEffect = m_pOuter->ParticleProp()->Create( pEffectName, PATTACH_POINT_FOLLOW, "head" );
 			}
 		}
 #endif
-
-		// Notify our weapon that we have been stunned.
-		CTFWeaponBase* pWpn = m_pOuter->GetActiveTFWeapon();
-		if ( pWpn )
+		if ( iStunFlags & ( TF_STUN_CONTROLS | TF_STUN_LOSER_STATE ) )
 		{
-			pWpn->OnControlStunned();
-		}
+			// Notify our weapon that we have been stunned.
+			CTFWeaponBase* pWpn = m_pOuter->GetActiveTFWeapon();
+			if ( pWpn )
+			{
+				pWpn->OnControlStunned();
+			}
 
-		if ( InCond( TF_COND_SHIELD_CHARGE ) )
-		{
-			SetDemomanChargeMeter( 0 );
-			RemoveCond( TF_COND_SHIELD_CHARGE );
-		}
+			if ( InCond( TF_COND_SHIELD_CHARGE ) )
+			{
+				SetDemomanChargeMeter( 0 );
+				RemoveCond( TF_COND_SHIELD_CHARGE );
+			}
 
-		m_pOuter->TeamFortress_SetSpeed();
+			m_pOuter->TeamFortress_SetSpeed();
+		}
 	}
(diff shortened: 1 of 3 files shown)


antielectronbomb:
Good for adding visual and audio recognition, sometimes I can't really tell if someone got hit by the ball or not.

src/game/shared/tf/tf_weapon_bat.cpp
```diff
@@ -732,12 +732,13 @@ void CTFStunBall::ApplyBallImpactEffectOnVictim( CBaseEntity *pOther )
...
 	{
 		bool bMax = flLifeTimeRatio >= 1.f;
-		int iStunFlags = ( bMax ) ? TF_STUN_SPECIAL_SOUND | TF_STUN_MOVEMENT : TF_STUN_SOUND | TF_STUN_MOVEMENT;
+		int iStunFlags = TF_STUN_MOVEMENT | TF_STUN_SOUND | TF_STUN_STUNBALL;
 		float flStunAmount = 0.5f;
 		float flStunDuration = Max( 2.f, tf_scout_stunball_base_duration.GetFloat() * flLifeTimeRatio );
 		if ( bMax )
 		{
 			flStunDuration += 1.0;
+			iStunFlags |= TF_STUN_SPECIAL_SOUND;
```
FlaminSarge, line 741: Worth noting for future readers that SPECIAL_SOUND fully overrides the regular sound in StunSound, so having both flags here won't be an issue.

FlaminSarge (commented):
Something seems clunky with the way stun flags are being used in general (not from this PR, but in existing logic). Might be worth trying to clean up/streamline?
Also, it might be worth piggybacking off of one of the other flags instead of adding a new flag e.g. TF_STUN_SOUND. That _will_ cause it to appear after a Scout's drink wears off, but you could argue that's thematic.

haaksirikko:
> Something seems clunky with the way stun flags are being used in general (not from this PR, but in existing logic). Might be worth trying to clean up/streamline?

Did cleanup of stun flag checking in StunPlayer.

>Also, it might be worth piggybacking off of one of the other flags instead of adding a new flag e.g. TF_STUN_SOUND. That _will_ cause it to appear after a Scout's drink wears off, but you could argue that's thematic.

Not sure if that's the way to go, although TF_STUN_SOUND is only used in the Bonk slowdown and Sandman so maybe it wouldn't matter much. In Bonk's case TF_STUN_NO_EFFECTS could be used to hide the effects. ~If I decide to proceed this way then the flag needs to be renamed to e.g. TF_STUN_EFFECTS~

haaksirikko:
> Also, it might be worth piggybacking off of one of the other flags instead of adding a new flag

Come think of it, `TF_STUN_DODGE_COOLDOWN` is a flag that was used for the original Bonk's post-effect slowdown but has gone unused since its removal and its reintroduction in Jungle Inferno does not use the flag

haaksirikko:
Now the effects appear on `TF_STUN_SOUND` and Bonk has been reworked to use `TF_STUN_DODGE_COOLDOWN`, which is reworked to only play the bonk sound effect. Removed `TF_STUN_STUNBALL`

FlaminSarge:
> Now the effects appear on `TF_STUN_SOUND` and Bonk has been reworked to use `TF_STUN_DODGE_COOLDOWN`, which is reworked to only play the bonk sound effect. Removed `TF_STUN_STUNBALL`

For clarity you can even add a new define TF_STUN_(whatever it should be named for the bonk effect) as
#define TF_STUN_(new_value) TF_STUN_DODGE_COOLDOWN

Then it won't be confusing to read in code but will still use the unused value instead of making a new value.

src/game/shared/tf/tf_player_shared.cpp
```diff
@@ -7340,41 +7340,41 @@ void CTFPlayerShared::OnAddOverhealed( void )
...
+				const char* pEffectName = "conc_stars";
+				m_pOuter->m_pStunnedEffect = m_pOuter->ParticleProp()->Create( pEffectName, PATTACH_POINT_FOLLOW, "head" );
 			}
 		}
 #endif
-
-		// Notify our weapon that we have been stunned.
-		CTFWeaponBase* pWpn = m_pOuter->GetActiveTFWeapon();
-		if ( pWpn )
+		if ( IsControlStunned() || IsLoserStateStunned() )
```
FlaminSarge: Is it worth optimizing IsControlStunned/IsLoserStateStunned? Do they only check stunflags?

src/game/shared/tf/tf_player_shared.cpp
```diff
@@ -7340,41 +7340,41 @@ void CTFPlayerShared::OnAddOverhealed( void )
...
-				}
-				else
-				{
-					const char* pEffectName = "conc_stars";
-					m_pOuter->m_pStunnedEffect = m_pOuter->ParticleProp()->Create( pEffectName, PATTACH_POINT_FOLLOW, "head" );
-				}
+				const char* pEffectName = "yikes_fx";
+				m_pOuter->m_pStunnedEffect = m_pOuter->ParticleProp()->Create( pEffectName, PATTACH_POINT_FOLLOW, "head" );
+			}
+			else if ( iStunFlags & ( TF_STUN_CONTROLS | TF_STUN_LOSER_STATE | TF_STUN_SOUND ) )
```
FlaminSarge: Is this intended to be that if _any_ of the 3 flags are set, the particle appears? Or was it meant to be if _all_ of them are present? (Loserstate is the one I'm a little concerned about, I don't recall whether those historically had the dizzy stars).

haaksirikko: Intended if any of the flags are set. The block should only be entered if the expression evaluates as nonzero.

Also yes, loserstate triggered the stars

src/game/shared/tf/tf_player_shared.cpp
haaksirikko: It just checks if the stun condition is present and if the respective stun flags are present. I originally decided to leave this alone as to not modify code more than necessary but the diff ended up being quite big here anyway, so I might optimize this

batbrain55:
It'd be neat if the Scout's voice lines also played when you hit someone with the Wrap Assassin's bauble.

haaksirikko:
Reworked again by using `TF_STUN_DODGE_COOLDOWN` as a new define `TF_STUN_EFFECTS`

> It'd be neat if the Scout's voice lines also played when you hit someone with the Wrap Assassin's bauble.

Out of scope for this PR.

SomeBloodyMinger:
This is long overdue. The sandman at the least deserves its old iconic voice-lines and BONK! indicator. Both have basically been unheard of since the disastrous patch in Jungle Inferno; except, funnily enough, MvM when a Spy disguises as Scout and saps a robot. The modern sandman has no soul anymore.

PR #1922, opened 2026-05-16
[TF2] Fix countdown timer skipping from 12 -> 10

asrichesson:
# Description

Fix bug where the countdown timer skips from 14..13..12..10..9.. in casual mode. 

Using `ceil` to make the timer consistent with the `"restart_timer_time"` client event that also uses `ceil`. See [`CTeamplayRoundBasedRules::Update( float frametime )`](https://github.com/ValveSoftware/source-sdk-2013/blob/3300848d8a25ef6403c91f82a4cd97d6daefbc06/src/game/shared/teamplayroundbased_gamerules.cpp#L3530)

Without `ceil` here, the final countdown will start too early. 10.9 seconds gets truncated to 10 seconds, while the ui still is showing 11 seconds (because it's rounding up using ceil). When the final countdown condition is met, it sets the remaining time to 10 seconds, causing the time to jump from 10.9 ->10, causing the ui to jump to 11 to 10 seconds in just a couple frames.

Instead, the final contdown should wait until the remaining seconds drops below 10 seconds remaining. 

Tested locally. See here for fix example [video](https://drive.google.com/file/d/1gvgx4oCgM9kEshMPP8Ya5s9GVBcKMhx7/view)

Broken/Current behavior:
CountdownBug

### DIFF
--- a/src/game/shared/tf/tf_gamerules.cpp
+++ b/src/game/shared/tf/tf_gamerules.cpp
@@ -2250,7 +2250,7 @@ bool CTFGameRules::BInMatchStartCountdown() const
 	if ( IsCompetitiveMode() )
 	{
 		float flTime = GetRoundRestartTime();
-		if ( ( flTime > 0.f ) && ( (int)( flTime - gpGlobals->curtime ) <= mp_tournament_readymode_countdown.GetInt() ) )
+		if ( ( flTime > 0.f ) && ( (int)ceil( flTime - gpGlobals->curtime ) <= mp_tournament_readymode_countdown.GetInt() ) )
 		{
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


Pandaptable:
How was this tested? I was going to submit a PR months ago but this isn't properly testable. You can't just say "tested locally" with no backing.

mastercoms:
There's a lot of obscure timing hacks with comparing to 10.1, 10.5, etc all throughout the codebase that determine explicit stages to the countdown when approaching the point of no return. In fact, one of your changes breaks this already, not counting the other forms of desync that you did not touch.

As you can see, when you modify bStartFinalCountdown to use ceil, it will only become true when the timer is 9.000001 to 10.0, when before it would be true when the timer is 10.0 to 10.99999. Then, in the if statement that uses bStartFinalCountdown, you can notice that it compares to flDropDeadTime, which will be 10.1. So, you effectively changed the drop dead time from starting at 10.1, to starting at 10.0. This has consequences for various parts of the code and can cause desync by extending the time window before drop dead time.

I have my own changes for this in my own repo, that I have not properly split out yet, including changes from this commit and others: https://github.com/mastercomfig/tc2/commit/26df685dc536ac58c1a924527b8db758233d7d07

asrichesson:
Thank you for the feedback. @Pandaptable, I've uploaded a video of me testing on a local server and added it to the description. You're right that it is very tricky to test. I changed the ui code that checks for the casual/competitive match group, specifically in `tf_hud_match_status.cpp` and `tf_hud_tournament.cpp`. Not saying this is ideal. But it was enough to track down the problematic code.

PR #1924, opened 2026-05-17
Fix html motd window reloading whenever any other panels are displayed.

bottiger1:
# Description

When you are viewing an HTML motd and if you spawn/die/round end or any situation where new panels are shown, the motd will be forcefully refreshed, causing flashing and anyone that scrolled down or clicked on another link would lose their navigation state.

Why this happens:

When any panel is shown, UpdateAllPanels() is called which then calls Update() on all panels including the motd. 

How this is fixed:

We record if the panel is showing html and the url. If upon calling Update() it is html and the same url, we don't redraw it.

### DIFF
--- a/src/game/client/game_controls/vguitextwindow.cpp
+++ b/src/game/client/game_controls/vguitextwindow.cpp
@@ -88,6 +88,7 @@ CTextWindow::CTextWindow(IViewPort *pViewPort) : Frame(NULL, PANEL_INFO	)
 	m_szTitle[0] = '\0';
 	m_szMessage[0] = '\0';
 	m_szMessageFallback[0] = '\0';
+	m_szLastURL[0] = '\0';
 	m_nExitCommand = TEXTWINDOW_CMD_NONE;
 	m_bShownURL = false;
 	m_bUnloadOnDismissal = false;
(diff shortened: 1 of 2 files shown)


src/game/client/game_controls/vguitextwindow.cpp
FlaminSarge, line 191: Do you need to empty out m_szLastURL when the client disables html motds, so that a client doesn't get stuck never showing an html motd if they toggle the setting twice (while still having the panel open)? Unsure if that's a case that can be hit.

bottiger1: no

cl_disablehtmlmotd only hides the window 

https://github.com/bottiger1/source-sdk-2013/blob/f9ec4155d7250d2c46391f45d92e4ca09dc4e4da/src/game/client/game_controls/vguitextwindow.cpp#L187

m_pHTMLMessage->SetVisible( true ); is always called if it is 0.

https://github.com/bottiger1/source-sdk-2013/blob/f9ec4155d7250d2c46391f45d92e4ca09dc4e4da/src/game/client/game_controls/vguitextwindow.cpp#L195

chaorace:
I experience this bug multiple times per session during Highlander scrim days. It's extremely common to be looking at a log in the MOTD panel while other players are goofing around killing each other (or farming me for strange kills).

Under current behavior, getting killed resets the scroll position, flashes a white background during page loading, then finally takes another 1-2 seconds after to finish reflowing. It's... actually extremely annoying. For me, certainly, but also certainly the several hundred other players who actively partake in league play on a regular basis. This fix would make a _lot_ of people's lives slightly happier.

PR #1928, opened 2026-05-23
[TF2] Fix missing pilot light on dropped stock flamethrower

num-get:
# Description
All stock weapons are assigned with INVALID_ITEM_ID by default. Consequently, the current code for dropped weapon will skip over all the stock weapons when looking up attachment. As a result, the pilot light only shows up on non-stock flamethrower.
This PR fixes this.

Before:

After:

### DIFF
--- a/src/game/shared/tf/tf_dropped_weapon.cpp
+++ b/src/game/shared/tf/tf_dropped_weapon.cpp
@@ -150,114 +150,111 @@ void CTFDroppedWeapon::OnDataChanged( DataUpdateType_t updateType )
 	if ( updateType == DATA_UPDATE_CREATED )
 	{
 		// if its startrak attach a model to it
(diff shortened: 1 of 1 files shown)


ZzZombo:
> All stock weapons are assigned with INVALID_ITEM_ID by default.

What's the reason for this? I'm curious.

JoriKos:
Alternatively, it would be neat if all of them were disabled when on the ground instead to signify them having 'turned off'.

FlaminSarge:
> Alternatively, it would be neat if all of them were disabled when on the ground instead to signify them having 'turned off'.

I agree with this, while other attachments may make sense to show on dropped weapons, the pilot light being on kinda doesn't, from a 'disabled weapon' perspective. Non-stock Flamethrowers should instead turn off their pilot lights when dropped.

PR #1929, opened 2026-05-24
fix [tf2]: Prevent wearables from leaking when weapon is dropped

### LINKED ISSUE #8021: [TF2] you can spawn infinite botkiller viewmodel attachments (tf_wearable and tf_wearable_vm) entities with the most recent update regarding spy disguises (and crash servers, with enough effort)
Image

Reproduction steps:
- someone who has a botkiller (can be on either team)
- disguise as them
- mash "Last Disguise" bind with revolver/sapper (primary, secondary if demoman/medic disguise) to cause the game to just create a botkiller attachment that tries to glue itself to your viewmodel.
- the more you mash it, the more botkiller attachment viewmodels are made (these are created on a delay, so they won't show up instantly. seems to be on a roughly 1 second delay?)
- oh by the way, each of these takes up 2 edict slots. so of course...
Image

- swapping disguises clears the botkiller attachments, and frees up the edict slots used.

seanmcgeehan:
# Description

Weapon extra wearables are orphaned during CTFWeaponBase::Drop due to RemoveExtraWearables being called after the drop. This is only exposed due to DetermineDisguiseWeapon actually processing Extra wearables. Viewmodel removal is moved before the drop to correctly clean up view models. World models are capped at 5 (3 for wearables, 1 for action slot + 1 extra wearable) to prevent a leak from accumulating. This enables disguises to correctly keep extrawearables while disguised as the same target. These wearables are correctly cleaned up on disguise removal. 

Tested in itemtest with record_entities and bot_mirror. 

closes https://github.com/ValveSoftware/Source-1-Games/issues/8021

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -8524,8 +8524,25 @@ void CTFPlayerShared::DetermineDisguiseWeapon( bool bForcePrimary )
 			m_hDisguiseWeapon->m_iState = WEAPON_IS_ACTIVE;
 			m_hDisguiseWeapon->m_bDisguiseWeapon = true;
 			m_hDisguiseWeapon->SetContextThink( &CTFWeaponBase::DisguiseWeaponThink, gpGlobals->curtime + 0.5, "DisguiseWeaponThink" );
+			m_hDisguiseWeapon->RemoveExtraWearables();
 
-			m_hDisguiseWeapon->UpdateExtraWearables();
+			// Cap accumulated disguise wearables. Each disguise-weapon swap
+			// intentionally orphans the prior weapon's world extras (so banners
+			// etc. stay visible across swaps); skip recreation once we're at
+			// the cap so they can't grow unbounded under rapid cycling.
+			// fully cleaned up after disguise removal.
+			const int kMaxDisguiseWearables = 5;
+			int nDisguiseWearableCount = 0;
+			for ( int i = 0; i < m_pOuter->GetNumWearables(); ++i )
+			{
+				CTFWearable *pWearable = dynamic_cast< CTFWearable * >( m_pOuter->GetWearable( i ) );
+				if ( pWearable && pWearable->IsDisguiseWearable() )
+					nDisguiseWearableCount++;
+			}
+			if ( nDisguiseWearableCount < kMaxDisguiseWearables )
+			{
+				m_hDisguiseWeapon->UpdateExtraWearables();
+			}
 
 			// Ammo/clip state is displayed to attached medics
 			m_iDisguiseAmmo = 0;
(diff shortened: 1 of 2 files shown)


RosneTF2:
Epic ai generated code! Love to see it 😀

num-get:
Wouldn't the max number of wearables be 6 because soldier, pyro and sniper can equip both a botkiller weapon and a backpack at the same time?

bakugo:
Gotta be honest, "fixing" this by adding a hard cap on the number of disguise wearables smells like spaghetti.

Wouldn't it make more sense to update `CTFWeaponBase::UpdateExtraWearables()` so that, if it finds an existing orphan disguise wearable with the right model, it re-associates itself with it instead of creating a new one?

seanmcgeehan:
@bakugo I was worried about edgecases that could be missed. I wouldn't object to revisiting this but felt this was the correct way to quickly fix this given the issue.

num-get:
two more problems, the botkiller model won't be removed when changing disguise


Second problem is that the wearable is capped at 4 instead of the intended 5 since you use < instead of <=
https://github.com/ValveSoftware/source-sdk-2013/blob/c98767b329f07c086281d787cc0e1c4d9a6b1410/src/game/shared/tf/tf_player_shared.cpp#L8534
https://github.com/ValveSoftware/source-sdk-2013/blob/c98767b329f07c086281d787cc0e1c4d9a6b1410/src/game/shared/tf/tf_player_shared.cpp#L8542

Missing botkiller model when equipping 3 cosmetics + 1 backpack + 1 botkiller weapon (I did change disguise weapon to make sure)


missing contracker when equipping 3 cosmetics + 1 backpack + 1 contracker

seanmcgeehan:
@num-get Botkiller removal on weapon change is an edge case due to how extra wearables are programmed. You'd have to hard code in a check for botkillers. I view this as extra curricular to the bugs solved. 

I can take a look at the less then 5 case down the line (I need some space after the stress and harrassment i got).

num-get:
From my further testing, action slot item seems to refuse to show up on disguise at all.

seanmcgeehan:
thats interesting, Missing action slot apparently was broken pre #1824. (just tested on older commits and its never worked.) I will take a look at that.

seanmcgeehan:
@num-get the < sign is logically correct, we only update extra wearables when we have 4 or less wearables. It’s possible you have an extra cosmetic somehow.

num-get:
> @num-get the < sign is logically correct, we only update extra wearables when we have 4 or less wearables. It’s possible you have an extra cosmetic somehow.

Apparently it only goes over the limit when botkiller is involved.
3 cosmetics + 1 backpack + 1 botkiller = missing jetpack (probably because jetpack is technically a weapon.)

2 cosmetics + 1 backpack + 1 botkiller = no problem

VladislavEkimtcov:
> Epic ai generated code! Love to see it 😀

AI generated code is an improvement to the Source codebase

PR #1931, opened 2026-05-24
[TF2] Fix disguising as a player without primary weapon

leemustache:
## Before
before

## After
after

## Reproduction steps
1. Disguise as any class
2. Disguise as a Demoman without a primary weapon (booties)
3. Your disguise is now holding previous disguise's weapon

## Explanation
([this video](https://youtu.be/TJM0AiMhqDg) explains it)

`DetermineDisguiseWeapon` gets called when Spy disguises or changes disguise weapon (the b key).
This function, among other things, checks if there's a "previously held weapon" to fallback to when attempting to switch to an inaccessible weapon slot (primary slot for Demoman with boots, secondary slot for soldier with parachute etc).

This works well when guarding against changing disguise weapon to an invalid slot with the b key. But when changing disguise the `m_hDisguiseWeapon` doesn't get cleared, so the "previous weapon" travels through different disguises.

That makes it so if the primary slot is not available it tries to pick the previously held valid weapon. Which in the scenario described in the reproduction steps is the previous disguise's weapon.

This change clears the `m_hDisguiseWeapon` field every time a spy changes disguise, fixing the issue.

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -8316,6 +8316,8 @@ void CTFPlayerShared::CompleteDisguise( void )
 	m_pOuter->UpdateModel();
 	m_pOuter->ClearExpression();
 
+	RemoveDisguiseWeapon();
+
 	FindDisguiseTarget();
 
 	if ( GetDisguiseTarget() )


seanmcgeehan:
Can you validate that clean up happens correctly with report_entities? This has a similar pattern to #1824  where i'd be worried about orphaning. Otherwise I think this makes sense.

leemustache:
@seanmcgeehan 
image

Seems fine to me.

This is after a couple of disguises from various classes to a demoman with booties.
The setup is on itemtest with a single spy bot and myself.
The entity count seems steady and not rising.

seanmcgeehan:
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/8216

PR #1933, opened 2026-05-24
lightmappedgeneric: add $forcetangentspace material param (fixes #1916)

### LINKED ISSUE #1916: tangentSpaceTranspose matrix in lightmappedgeneric should be correct when TANGENTSPACE parameter is set to true and the material has no envMap or bumpMap
I'm currently working on adding an interior mapping shader to a source engine mod, using cubemaps, and tangent space. The interior cubemap is added as an extra texture to the lightmappedgeneric_dx9 pixel shader. To correctly translate our pixelToEye vector from world space to tangent space, I need the tangentSpaceTranspose calculated in the vertex shader. The vertex shader has a static parameter TANGENTSPACE, which when 1, calculates this matrix.

InteriorMapped brushes with the TANGENTSPACE parameter set to true and hasEnvMap bool in lightmappedgeneric_dx9_helper.cpp set to false.

> InteriorMapped brushes with the TANGENTSPACE parameter set to true and hasEnvMap bool in lightmappedgeneric_dx9_helper.cpp set to false.

For the longest time i could not get this transformation to work, until I used the components of the matrix as colours and found out that the tangent and binormal are both 0 0 0. After some more experimenting I worked out that either passing a dummy bumpmap or dummy envmap (or setting hasBump or hasEnvMap in lightmapgeneric_dx9_helper.cpp to true, with the latter producing no errors far as I can tell), the tangent and binormal are finally correctly passed into the vertex shader.

InteriorMapped brushes with the TANGENTSPACE parameter set to true and hasEnvMap bool in lightmappedgeneric_dx9_helper.cpp set to true.

> InteriorMapped brushes with the TANGENTSPACE parameter set to true and hasEnvMap bool in lightmappedgeneric_dx9_helper.cpp set to true.

I think this quirk in how the vertex shader does not receive the tangent and binormal when neither a bumpmap or envmap is present (or some other parameter I don't know about), and thus TANGENTSPACE calculates an incorrect tangentSpaceTranspose, could be better documented, either as a comment in the pixel and vertex shader, or via some new static parameter that sets the tangent and binormal correctly even without a bumpmap or envmap.

The shader was made with the help of this [YouTube tutorial](https://www.youtube.com/watch?v=2qVSV5dhdb0&list=PL78XDi0TS4lFHEvvUXpOoTzV-IdCs1hNk&index=2) by Ben Cloward.

ianherzing:
Fixes #1916.

When a `lightmappedgeneric` material has neither `$bumpmap` nor `$envmap`, the vertex shader's `TANGENTSPACE` static combo is forced off and the vertex format omits `VERTEX_TANGENT_S | VERTEX_TANGENT_T`, so tangent and binormal arrive at the vertex shader as zero. This breaks custom shader paths (e.g. interior mapping via cubemaps) that need tangent-space transforms but don't carry a normal map or envmap.

This PR adds an opt-in material parameter `$forcetangentspace` (default `0`). When set, the helper:

1. ORs `bForceTangentSpace` into the gate at `lightmappedgeneric_dx9_helper.cpp:452`, so the vertex format gets `VERTEX_TANGENT_S | VERTEX_TANGENT_T | VERTEX_NORMAL`.
2. ORs it into the `TANGENTSPACE` static combo selector at line 536 so the vertex shader actually consumes the stream.

The new struct field defaults to `-1` via the existing `memset(this, 0xFF, sizeof(...))` in `LightmappedGeneric_DX9_Vars_t`, so any caller (including `worldvertextransition.cpp`, which shares the struct) that doesn't set the new param is bit-for-bit unchanged. Materials that don't declare `$forcetangentspace` are bit-for-bit unchanged.

No shader recompile required — the change only adjusts which inputs feed the existing `TANGENTSPACE` combo.

**On testing:** I did not set up a full SDK build + custom shader to visually confirm tangent/binormal arrival. The reporter of #1916 already demonstrated empirically (screenshots in the issue) that forcing `hasEnvMap=true` in the helper — which flips exactly the same two gates this PR widens — makes tangent and binormal arrive correctly in the vertex shader. This PR exposes that same toggle as an opt-in material parameter without altering the underlying mechanism. Happy to do a build + capture if a reviewer wants extra confirmation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/materialsystem/stdshaders/lightmappedgeneric_dx9.cpp
+++ b/src/materialsystem/stdshaders/lightmappedgeneric_dx9.cpp
@@ -73,6 +73,8 @@ BEGIN_VS_SHADER( LightmappedGeneric,
 		SHADER_PARAM( OUTLINESTART1, SHADER_PARAM_TYPE_FLOAT, "0.0", "inner start value for outline")
 		SHADER_PARAM( OUTLINEEND0, SHADER_PARAM_TYPE_FLOAT, "0.0", "inner end value for outline")
 		SHADER_PARAM( OUTLINEEND1, SHADER_PARAM_TYPE_FLOAT, "0.0", "outer end value for outline")
+
+		SHADER_PARAM( FORCETANGENTSPACE, SHADER_PARAM_TYPE_BOOL, "0", "Force tangent-space vertex stream even without $bumpmap/$envmap" )
 END_SHADER_PARAMS
 
 	void SetupVars( LightmappedGeneric_DX9_Vars_t& info )
@@ -134,6 +136,8 @@ END_SHADER_PARAMS
 		info.m_nOutlineStart1 = OUTLINESTART1;
 		info.m_nOutlineEnd0 = OUTLINEEND0;
 		info.m_nOutlineEnd1 = OUTLINEEND1;
+
+		info.m_nForceTangentSpace = FORCETANGENTSPACE;
 	}
 
 	SHADER_FALLBACK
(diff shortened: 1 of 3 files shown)


misyltoad:
The shader wouldn't use a cubemap unless you set it as $envmap. You can just make your own vertex shader that exposes this if you want it.

PR #1935, opened 2026-05-24
dt_utlvector_recv: log diagnostic before RCE-prevention crash (fixes #1511)

### LINKED ISSUE #1511: [All] Explain client crash reason when malicious server sends CUtlVector length not in allowed range
~Due to off by one error above if server sends `m_Int == pExtra->m_nMaxElements` as length it will be accepted by the client.
The problem is `CUtlVector` indexes lie in the range `[0...length)` so it is classic off by one error.~

~Corrected segment (notice added `Error` statement so client has a chance to understand why his game crashed):~

We have a silent crash when malicious server has send CUtlVector length not in allowed range:
```
void RecvProxy_UtlVectorLength( const CRecvProxyData *pData, void *pStruct, void *pOut )
{
	CRecvPropExtra_UtlVector *pExtra = (CRecvPropExtra_UtlVector*)pData->m_pRecvProp->GetExtraData();
	if ( pData->m_Value.m_Int < 0 || pData->m_Value.m_Int > pExtra->m_nMaxElements )
	{
		// If this happens we're most likely talking to a malicious server.
		// Protect against remote code execution by crashing ourselves.
		// A malicious server can send an invalid lengthprop attribute and cause the below code
		// to "successfully" resize the vector to -1, which eventually translates into a call to realloc(0)
		// due to integer math overflow.
		// Then the remaining payload ( the actual elements of the vector ) can be used
		// to write arbitrary data to out of bounds memory.
		// There isn't much we can do at this point - we're deep in the networking stack, it's hard to recover
		// gracefully and we shouldn't be talking to this server anymore.
		// So we crash.
		*(int *) 1 = 2;
	}
```

What I suggest is to add `Error` statement just before crash so player can understand why his client crashed.
```
void RecvProxy_UtlVectorLength( const CRecvProxyData *pData, void *pStruct, void *pOut )
{
	CRecvPropExtra_UtlVector *pExtra = (CRecvPropExtra_UtlVector*)pData->m_pRecvProp->GetExtraData();
	if ( pData->m_Value.m_Int < 0 || pData->m_Value.m_Int > pExtra->m_nMaxElements )
	{
		// If this happens we're most likely talking to a malicious server.
		// Protect against remote code execution by crashing ourselves.
		// A malicious server can send an invalid lengthprop attribute and cause the below code
		// to "successfully" resize the vector to -1, which eventually translates into a call to realloc(0)
		// due to integer math overflow.
		// Then the remaining payload ( the actual elements of the vector ) can be used
		// to write arbitrary data to out of bounds memory.
		// There isn't much we can do at this point - we're deep in the networking stack, it's hard to recover
		// gracefully and we shouldn't be talking to this server anymore.
		//
		// So we notify client.
		Error("Server send utlvector length value %d which is not in range [%d...%d]. Crashing client to prevent RCE...\n",
		    pData->m_Value.m_Int, 0, pExtra->m_nMaxElements);
		// And crash.
		*(int *) 1 = 2;
	}
```

ianherzing:
Fixes #1511.

`RecvProxy_UtlVectorLength` in `src/public/dt_utlvector_recv.cpp` deliberately crashes the client when a server sends an out-of-range `lengthprop`, on the grounds that we're deep in the networking stack and any non-crash recovery would risk arbitrary memory writes (the existing comment block explains the RCE vector). The deliberate crash is correct.

The problem #1511 raises is that the crash is *silent* — the user gets a hard `*(int *) 1 = 2;` segfault with no message, no log line, no clue what happened. From an end-user / support perspective this looks indistinguishable from a random crash, and there's nothing for them to report.

This PR adds a single `Warning()` line immediately before the crash, describing the offending lengthprop value and the allowed range. The deliberate crash is unchanged.

```diff
+		// Log the diagnostic before crashing so the user knows why their client died.
+		Warning( "RecvProxy_UtlVectorLength: server sent lengthprop %d outside allowed range [0, %d]. "
+			"Crashing client to prevent possible remote code execution.\n",
+			pData->m_Value.m_Int, pExtra->m_nMaxElements );
 		// So we crash.
 		*(int *) 1 = 2;
```

**Why `Warning()` instead of `Error()`:** The reporter suggested `Error()`. I chose `Warning()` because the existing comment is emphatic that we want to crash *immediately* without doing anything else ("we're deep in the networking stack, it's hard to recover gracefully"). `Error()` on Source can route through assert handlers, dialogs, and spew sinks, which is more pre-crash work than we want here. `Warning()` is the lightest log call (printf-equivalent with a "Warning: " prefix) and matches the "log and crash" pattern used elsewhere in the engine. Open to swapping if reviewers prefer `Error()` or `Msg()`.

**On testing:** Did not stand up a local SDK build — the change is a strictly additive log line in a code path whose existing crash behavior is preserved bit-for-bit. The new `Warning()` call uses the standard `tier0/dbg.h` printf-style signature and the format specifiers match the int types being logged. Happy to add a build verification if a reviewer wants one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/public/dt_utlvector_recv.cpp
+++ b/src/public/dt_utlvector_recv.cpp
@@ -36,6 +36,10 @@ void RecvProxy_UtlVectorLength( const CRecvProxyData *pData, void *pStruct, void
 		// to write arbitrary data to out of bounds memory.
 		// There isn't much we can do at this point - we're deep in the networking stack, it's hard to recover
 		// gracefully and we shouldn't be talking to this server anymore.
+		// Log the diagnostic before crashing so the user knows why their client died.
+		Warning( "RecvProxy_UtlVectorLength: server sent lengthprop %d outside allowed range [0, %d]. "
+			"Crashing client to prevent possible remote code execution.\n",
+			pData->m_Value.m_Int, pExtra->m_nMaxElements );
 		// So we crash.
 		*(int *) 1 = 2;
 	}


KGBTF2:
More slop please!

ChaseTownsend:
I feel like valve could have just used the Error function to cause a fatal error with a message rather than just setting 1 to 2

chocochocotorta:
The person literally mass-submitted PRs without even testing them (+ AI generated).

sapphonie:
I normally don't get mean on PRs but I think you should be blocked from contributing to this and other Valve repos for this slop

misyltoad:
This warning would never be seen because the game crashes intentionally after.
Ironically, Error is actually better here, which the magical nonsense AI decided against for some reason.

PR #1939, opened 2026-05-29
[TF2] Fix botkiller related disguise issue

num-get:
# Description
Botkiller disguise has two issues. The spy could clone up to 5 botkiller models when switching to botkiller weapon, and the botkiller model staying after the spy switches to other disguise weapon.
This PR fixes this by looking up items that have item definition "extra_wearable_vm" which is currently exclusively used by botkiller weapons, and then remove the botkiller model.

Before:

https://github.com/user-attachments/assets/fc468719-139d-4de9-8563-867726d0814e

After:

https://github.com/user-attachments/assets/eed68c31-a1f2-4c0a-a202-2e52f6cce275

Edit:
Currently this PR will cause edict leak, I will evaluate a better solution later.

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -8490,6 +8490,12 @@ void CTFPlayerShared::DetermineDisguiseWeapon( bool bForcePrimary )
 
 	if ( strDisguiseWeapon )
 	{
+		// Remove botkiller related wearables
+		if ( m_hDisguiseWeapon && m_hDisguiseWeapon->GetAttributeContainer()->GetItem()->GetStaticData()->GetExtraWearableViewModel() )
+		{
+			m_hDisguiseWeapon->RemoveExtraWearables();
+		}
+
 		// Remove the old disguise weapon, if any.
 		RemoveDisguiseWeapon();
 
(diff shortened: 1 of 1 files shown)


seanmcgeehan:
As discussed this leaks and causes the issue in https://github.com/ValveSoftware/Source-1-Games/issues/8021

I would also be worried that you are also moving these lines to after the drop, which might lead to a leak.

```
		// Remove botkiller related wearables
		if ( m_hDisguiseWeapon && m_hDisguiseWeapon->GetAttributeContainer()->GetItem()->GetStaticData()->GetExtraWearableViewModel() )
		{
			m_hDisguiseWeapon->RemoveExtraWearables();
		}


```

The drop happens here.

https://github.com/ValveSoftware/source-sdk-2013/pull/1929/changes#diff-aa946fb2c369973e2c7604d6b538442821f1a81a88f2f9eede45915759ea6118R1137


My suggestion is to leave the cap in place and try to move these lines to right before the drop. That likely fixes the issue. RemovingExtraWearbles entirely in this function will lead to a regression in the battalions backup + quickfix + vaccinator cases, so that cant be done. You have to specifically target botkillers with hard coding. 

This is why i went with the cap solution, its ugly, but it prevents the leak and preserves the banner + medigun cases. 

In any case you *must* test battalions backup + quickfix + vaccinator to avoid regression.

seanmcgeehan:
Just reread, and you are actually before the drop but i still think this should go in CTFWeaponBase::Drop before line 1137.

PR #1940, opened 2026-05-29
Stop splash damage from being able to go through walls

Goennigoegoe:
Fixes a bug that lets splash damage go through walls in certain cases.
The bug is caused due to plane normal being added to explosion pos and explosion damage not being checked correctly for starting in a solid.

It has 2 fixes, one that adds the check for starting in a solid and another that removes the plane normal addition.
They are togglable via the ConVars: "tf_splash_fix" and "tf_splash_fix_extreme".
The extreme version might cause issues with splash damage being eaten by things it shouldn't be. It does fix a bug where you are able to shoot around corners if aiming close enough to the corner.
The normal version doesn't cause issues as far as I've tested.

Showcase: https://www.youtube.com/watch?v=OJ7SZjltdbw

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -502,4 +502,3 @@ FodyWeavers.xsd
 
 # JetBrains Rider
 *.sln.iml
-/game
--- a/.gitignore
+++ b/.gitignore
@@ -502,3 +502,4 @@ FodyWeavers.xsd
 
 # JetBrains Rider
 *.sln.iml
+/game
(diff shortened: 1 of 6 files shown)


GrubPL:
I don't exactly feel like a PR related to fixing splash dmg through walls should include edits to .gitignore or add in a tf specific build script

Goennigoegoe:
Oh sorry, it was probably due to my visual studio stuff, I will try removing them.

Goennigoegoe:
I have reverted them, it still has extra commits for it but eh, it doesn't add / change the files anymore.

Goennigoegoe:
Here's the test position since I forgot to add it into the description: setpos 749.287903 1829.719727 -187.167358;setang -23.551832 11.992801 0.000000

RileyKun:
I personally hope this gets merged as this exploit is abused too much by malicious clients.

PR #1941, opened 2026-05-30
Added `$frame` params for `screenspace_general`

Akabenko:
Here 4 new params of frame for each texture.


PiMoNFeeD:
`$frame` is a default parameter, you must not add it to the shader yourself

Akabenko:
> `$frame` is a default parameter, you must not add it to the shader yourself
Thanks

PR #1942, opened 2026-05-30
[TF2] Fix 2 bugs related to `melee_cleave_attack`

BreavyTF2:
Makes it so dead players can't be hit and fixes a crash if the target becomes null.


FlaminSarge:
How do non-cleave attacks handle this, out of curiosity? Is it built into the raytrace filter handling (where cleave doesn't use those at all).

PR #1951, opened 2026-06-05
[TF2] Fix Thermal Thruster passives not being removed if it's unequipped during use (or player is a Medic) - [Resubmission]

### LINKED ISSUE #4731: [TF2] Thermal Thruster retains passive stats when it shouldn't
If you pick up a secondary weapon while using a Thermal Thruster charge, the passive stats of +50% knockback, -75% less fall damage, as well as the ability to stomp people are retained until death/switching class.

rust-src:
First and foremost: unless it is a discussion regarding the code itself, this Pull Request is _not_ the appropriate place for discussions regarding this bug or weapon balance. 

I'm also resubmitting this of my own volition. **Do not harass the original PR author**.

If you wish to view a (now closed) discussion regarding this bug, you can view that here:

https://github.com/ValveSoftware/Source-1-Games/issues/7654

If you wish to add to the discussion, the following thread is still active and open (PLEASE do so constructively!):

https://github.com/ValveSoftware/Source-1-Games/issues/7802

Now that is out of the way, a few thoughts here.

This is a reimplementation of a PR that was previously merged, then later reverted. You can find it [here](https://github.com/ValveSoftware/source-sdk-2013/pull/1490).

From my understanding, it was reverted due to strong feedback from some players who preferred the bugged behavior.

While I personally believe known bugs should be fixed regardless of player preference, I recognize that opinions on this may differ greatly, particularly around whether certain weapons (such as the Detonator) should receive balance adjustments to compensate. However, this is neither here nor there. 

This PR is strictly intended as a bug fix for an issue that has existed for a long time. Any discussion around weapon balance is best handled in their own appropriate threads, not in this PR. Thank you!

### Original PR text:

Fixes https://github.com/ValveSoftware/Source-1-Games/issues/4731

The `TF_COND_ROCKETPACK` cond (stomp, lower fall damage, etc) from the Thermal Thruster is supposed to be removed as soon as the player touches the ground, but the code responsible does not do this if the player does not have a Thermal Thruster equipped.

As a result, players can stop the cond from being removed and keep it until they die by unequipping the Thermal Thruster after using it but before touching the ground (such as by flying into a resupply cabinet after changing their loadout). Additionally, Medics with the Quick Fix will always keep the cond until they die if they heal a Pyro as he uses the Thermal Thruster, because they inherit the cond from the Pyro but don't have the weapon themselves, so the cond is never removed from them.

Fixed by removing `TF_COND_ROCKETPACK` if a player touches the ground without the Thermal Thruster equipped.

### DIFF
--- a/src/game/shared/tf/tf_player_shared.cpp
+++ b/src/game/shared/tf/tf_player_shared.cpp
@@ -3160,6 +3160,11 @@ void CTFPlayerShared::ConditionThink( void )
 #endif
 				}
 			}
+			else
+			{
+				// RocketPack not found, may have been unequipped, or we may be a Medic who inherited the cond
+				RemoveCond( TF_COND_ROCKETPACK );
+			}
 		}
 	}
 


JoriKos:
Couldn't you implement a ConVar to toggle this behaviour, then it can be disabled on Casual servers while community servers can determine it themselves?

Luxzhv:
Waiting for the same exact thing to happen where the pyro mains moan about this. Thank you for this bug fix!

megascatterbomb:
> Couldn't you implement a ConVar to toggle this behaviour, then it can be disabled on Casual servers while community servers can determine it themselves?

Community servers with custom weapon requirements can reimplement this behaviour using any of the custom weapon frameworks available for sourcemod.

This bug should not be in Valve servers nor should it be an officially supported feature via cvar. Having two implementations (one of which is intentionally left buggy) will just make future maintenance more tedious and prone to additional bugs.

JoriKos:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/1951#issuecomment-4640914927

Submitting the same fix that caused a controversy and expecting it to be accepted is really naive though. Clearly Valve doesn't want this fix in the game as it was presented at the time. A ConVar is much easier to work with than having to install Sourcemod for just a single weapon change. As I mentioned in my comment, it would be disabled on Casual servers (so it would not be possible to do in Valve servers) and there's no reason to not have the option for server owners to have it enabled.

There's also a precedent for turning bugged behaviour into a ConVar. Take a command like `tf_allow_taunt_switch` which lets you survive the Equalizer taunt with the Rocket Jumper (and even explicitly mentions it being used to replicate bugged behaviour). Or commands like `tf_allow_sliding_taunt` and `tf_overtime_nag` which were also created to replicate bugged behaviour. Just asking for the same thing to be implemented with 0 changes to appease those that complained about it is pretty much knowingly creating a PR that'll never be accepted, at least with a ConVar people can toggle it on or off if they wish to but it (rightfully) won't be allowed on official servers and you can argue if they want this kind of behaviour they'll have to find a server that enables it.

As it stands, this bug just won't be fixed and refusing any compromise that'll make it effectively fixed except on a few servers that 99% of the playerbase doesn't play on isn't going to help improve the game.

UberMedicFully:
I will never understand those who try and excuse a oversight in the code, to call it a feature when it clearly was not meant to be a thing ever

slender100:
While we're at it, we need to also patch out the spycrab as well. Think of all of the times that headshots were missed because of spy's arm getting in the way of his head, due to the animation bug.

PR #1953, opened 2026-06-06
[TF2] Implement unused teleporter activation sound for MvM

CookieCat45:
Implements the unused Teleporter activation sound for MvM into the game: `mvm/mvm_tele_activate.wav`

https://github.com/user-attachments/assets/63f08176-6bce-4e4d-a5c7-6e9470d65174

### DIFF
--- a/src/game/server/tf/tf_obj_teleporter.cpp
+++ b/src/game/server/tf/tf_obj_teleporter.cpp
@@ -445,7 +445,7 @@ void CObjectTeleporter::OnGoActive( void )
 		}
 	}
 
-	if ( TFGameRules()->IsMannVsMachineMode() )
+	if ( TFGameRules()->IsMannVsMachineMode() && GetTeamNumber() == TF_TEAM_PVE_INVADERS )
 	{
 		EmitSound( "mvm/mvm_tele_activate.wav" );
 	}
--- a/src/game/server/tf/tf_obj_teleporter.cpp
+++ b/src/game/server/tf/tf_obj_teleporter.cpp
@@ -444,6 +444,11 @@ void CObjectTeleporter::OnGoActive( void )
 			UpdateMaxHealth( pMatch->GetMaxHealth() );
 		}
 	}
+
+	if ( TFGameRules()->IsMannVsMachineMode() )
+	{
+		EmitSound( "mvm/mvm_tele_activate.wav" );
+	}
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


JoriKos:
You may want to have a better commit message than 'almost forgot'.

FlaminSarge:
> You may want to have a better commit message than 'almost forgot'.

+1 Valve tends to merge the commits as-is instead of squash merge, so the commit history will end up with word-for-word what you put as the commit message. Best to make them as descriptive as you can (though you'll need to force push over your current branch's history, sadly).

src/game/server/tf/tf_obj_teleporter.cpp
```diff
@@ -482,6 +487,7 @@ void CObjectTeleporter::Precache()
 	PrecacheScriptSound( "Building_Teleporter.SpinLevel1" );
 	PrecacheScriptSound( "Building_Teleporter.SpinLevel2" );
 	PrecacheScriptSound( "Building_Teleporter.SpinLevel3" );
+	PrecacheSound("mvm/mvm_tele_activate.wav");
```
FlaminSarge, line 490: This should most likely be a script sound like the other sounds, and set up in game_sounds_mvm.txt. You can request for changes to those files in the PR description, and set up the code as if those files were updated.

Also this precache should probably happen alongside the other MVM sounds.

src/game/server/tf/tf_obj_teleporter.cpp
```diff
@@ -444,6 +444,11 @@ void CObjectTeleporter::OnGoActive( void )
 			UpdateMaxHealth( pMatch->GetMaxHealth() );
 		}
 	}
+
+	if ( TFGameRules()->IsMannVsMachineMode() && GetTeamNumber() == TF_TEAM_PVE_INVADERS )
+	{
+		EmitSound( "mvm/mvm_tele_activate.wav" );
+	}
```
FlaminSarge, line 451: Instead of here, this logic should live in `C_TFBotHintEngineerNest::StartEffect` in `src\game\client\tf\bot_npc\map_entities\c_tf_bot_hint_engineer_nest.cpp` alongside the particle effect logic, or in its serverside code `CTFBotHintEngineerNest::HintTeleporterThink` in `src\game\server\tf\bot\map_entities\tf_bot_hint_engineer_nest.cpp`.

There's already examples of EmitSound being called on hint entities like this, e.g. `CTFBotMvMEngineerTeleportSpawn::Update` calls EmitSound on its hint entity. However, if you want the noise to be global, you can use TFGameRules()->BroadcastSound (which that class also does).

(You could probably get away with emitting it on the clientside object, honestly).

Pygargue00fr:
This is honeslty such a simple yet meaningfull change, without this sound there's now ways to know if a teleporter is active other than checking the wave status, which people often don't in the middle of battle.

A welcomed change for the mvm update :)

FrozenDragon0:
I'm genuinely curious as to why this was never used considering having an audio cue when a tele activates would be super helpful.

EricS-Valve:
We've implemented this in a different way than the pull request, so we're going to close this. Thanks for the report.

PR #1954, opened 2026-06-07
[TF2] Fix players being stuck in teleporter when being revived in MVM

loporlp:
If a player's reanimator is dropped inside a friendly teleporter, they will be trapped upon being revived. This occurs because friendly buildings are added to a filter when checking if the respawn location is valid. This PR fixes this by not ignoring friendly buildings when checking respawn location.

Before:

https://github.com/user-attachments/assets/d605d86d-3dad-4eeb-94d0-55a1e6c3ba3c

After:

https://github.com/user-attachments/assets/808a84aa-63e2-4a1c-b6eb-ec862ebd3a57

### DIFF
--- a/src/game/shared/tf/tf_revive.cpp
+++ b/src/game/shared/tf/tf_revive.cpp
@@ -380,7 +380,7 @@ bool CTFReviveMarker::ReviveOwner( void )
 	// See if their marker is clear
 	Vector vecTeleportPos = GetAbsOrigin();			
 	trace_t tr;
-	CTraceFilterIgnoreTeammatesAndTeamObjects filter( m_hOwner, COLLISION_GROUP_NONE, m_hOwner->GetTeamNumber() );
+	CTraceFilterSimple filter( m_hOwner, COLLISION_GROUP_NONE );
 	UTIL_TraceHull( vecTeleportPos, vecTeleportPos, VEC_HULL_MIN_SCALED( m_hOwner ), VEC_HULL_MAX_SCALED( m_hOwner ), ( MASK_SOLID | CONTENTS_PLAYERCLIP ), &filter, &tr );
 		
 	// If not, try the medic's location


FlaminSarge:
This would trigger the fix if a reanimator is inside a sentry or dispenser as well. While that's probably safer in general, any idea if there's a way to make this happen strictly for teleporters? To that end, it looks like CTraceFilterIgnoreTeammatesAndTeamObjects also has some logic around ignoring spies and stuff, so you may have also fixed an issue where a disguised/cloaked Spy standing on a reanimator would still allow the rez to occur in that spot. Would be interested to see if that's the case.

loporlp:
I think we could make this teleporter-specific by creating a new trace filter or modifying CTraceFilterIgnoreTeammatesAndTeamObjects to selectively collide with teleporters while still ignoring other friendly buildings.

That said, I’m leaning toward the current approach because revive placement feels safer when we don’t ignore any entities during the collision check. While sentries/dispensers clipping into a reanimator is probably uncommon, this also protects against future cases involving other collidable entities/buildings without needing additional special cases.

As for a disguised/cloaked spy the fix is triggered, so this should resolve any issues with getting stuck in a spy.

https://github.com/user-attachments/assets/59e0bcf7-02cb-4c88-8735-09951ceeed05

FlaminSarge:
> As for a disguised/cloaked spy the fix is triggered, so this should resolve any issues with getting stuck in a spy.

Nice. I'd be interested to see what happens in the Spy case without the fix (if they have some other failsafe in place), if you have that recording on hand.

JoriKos:
> While sentries/dispensers clipping into a reanimator is probably uncommon

Happens more often than you think, though generally only to Engineers as they get killed by their own Sentry when it's shooting at something behind them or when they get backstabbed by a Spy.

PR #1964, opened 2026-06-26
Fix client crash by skipping bad overlays when compiling the bsp

bottiger1:
In VBSP, skip overlays that bind to an entity, they crash the client.

Client side code to ignore overlays requires access to engine code.

# Description


SirYodaJedi:
Does the check properly occur after func_detail, func_ladder, etc. are collapsed to the world? I tried to check the code myself, but overlay creation (in EmitFace() in writebsp.cpp) and entity handling (in CMapFile::LoadEntityCallback() in map.cpp) happen so far apart from each other that I was having difficulty telling which happens first.

PR #1968, opened 2026-07-02
[TF2] Move two of Helltower's Soul Gargoyle locations to in-bounds locations

JoriKos:
As title says. Two Soul Gargoyle locations are out of bounds, both on the cliff side. A similar PR already exists under under #1290, except this one doesn't remove the locations but instead picks two more appropriate locations near the tracks. I considered putting them near the incline, but since that is a place often filled with Sentry Guns it did not seem a fair location, and I considered putting it more near the cliff, but there was not much space to make it look good and putting it on the tracks risks the Payload cart blocking the Soul Gargoyle (not necessarily the hitbox, but definitely visually).

Below are two images showing what the locations look like (they are as centered as I could get them):

Hightower 1

Hightower 2

Same location as the second image from a different angle, better showing how centred it is:

Hightower 3


JoriKos:
Closing now that the locations have been removed entirely

PR #1970, opened 2026-07-07
[TF2] Updated Spy's cigarette to extinguish on milk/gas

JoriKos:
As title says, the Spy's cigarette is 'extinguished' (no smoke particle) when covered in milk or gas. This behaviour already exists when covered in Jarate, and also when dripping with water, but not when covered in milk (would extinguish a cigarette) or gas (safety hazard).


universe-breaker:
since when was mad milk, milk?

JoriKos:
I did consider making the comment 'milk' (in quotation marks), or non-milk substance. It would have been funny for a joke, but it may also be a bit vague as to what it means (or be seen as the most obvious joke everyone's been making, even though I would be referencing the item type).

PR #1977, opened 2026-07-21
[TF2] Add support for Robot models in 3D playerhud while disguised in MvM

rabscootle:
This PR adds support to draw robot models in the 3D playerhud while disguised in Mann vs Machine mode.

A brief delay has also been added when updating disguises (MvM and non-MvM) to ensure that the disguise weapon has been properly updated before refreshing the model.

https://github.com/user-attachments/assets/e6b3d439-d7db-492d-b435-5cf3416b58e3


Pygargue00fr:
Why this isn't a thing yet is one of life's many questions we will never find an awnser to.

Chickenman456:
A huge no-brainer

PR #1991, opened 2026-08-03
[TF2] Fix multiple class menu related issues

num-get:
# Description
Based on mastercoms implementation of class menu fix.
This PR addresses two issues regarding class selection menu. First is missing sound effect, second is facial animation not resetting at the start of the animation.

Before:

https://github.com/user-attachments/assets/0b30393e-a82c-482b-bd44-da4da8679cdb

After:

https://github.com/user-attachments/assets/43b42e43-4771-4c2e-be47-973b57731e08

In addition, Spy has faulty class_select.vcd which causes Spy to have choppy class select animation at the start. The .vcd will also cause Spy to repeatedly huffing his cigarette instead of playing his idle animation. After this PR is applied, the faulty .vcd will cause Spy to play idle animation for 0.1 second at the start of his class select animation.
In order to avoid all the issues listed above, this PR must be accompanied with .vcd fix. Valve will need to place class_select.vcd inside tf/scenes/player/spy/low and then rebuild scenes.image.

Edit: Restore missing facial animation where Spy blows the cigarette smoke.
[class_select.zip](https://github.com/user-attachments/files/30703242/class_select.zip)

Before:

https://github.com/user-attachments/assets/d2d6a904-2aed-4de0-8ff4-4360e3f4f8b5

After without .vcd fix:

https://github.com/user-attachments/assets/deded8fe-6af7-4216-b3a8-f3e743b68a7f

After with .vcd fix:

https://github.com/user-attachments/assets/e49779d0-f0d6-47dd-84fc-d06d112df077

Lastly, the class menu used to have spotlight that was absent since the MVM update. The player model looks much darker compared to how it was before 2012. The method to fix this was mentioned by https://github.com/ValveSoftware/Source-1-Games/issues/4121
Valve needs to put this fixed classselection.res inside tf/resource/ui

[classselection.zip](https://github.com/user-attachments/files/30666918/classselection.zip)

Before:

After:
Screenshot 2026-08-03 23-28-49

If all the suggestions are implemented, this related issue can be closed for good.
https://github.com/ValveSoftware/Source-1-Games/issues/4121

### DIFF
--- a/src/game/client/tf/vgui/tf_playermodelpanel.cpp
+++ b/src/game/client/tf/vgui/tf_playermodelpanel.cpp
@@ -486,6 +486,7 @@ void CTFPlayerModelPanel::PlayVCD( const char *pszVCD, const char *pszWeaponEnti
 	m_bVCDFileNameOnly = bFileNameOnly;
 }
 
+extern ConVar _cl_classmenuopen;
 //-----------------------------------------------------------------------------
 // Purpose: 
 //-----------------------------------------------------------------------------
@@ -508,6 +509,25 @@ void CTFPlayerModelPanel::FireEvent( const char *pszEventName, const char *pszEv
 		{
 			m_aMergeMDLs[nWeaponIndex].m_bDisabled = false;
 		}
+	}
+	else if ( V_strcmp( pszEventName, "AE_CL_PLAYSOUND" ) == 0 || V_strcmp( pszEventName, "CL_EVENT_SOUND" )
+			|| V_strcmp( pszEventName, "CL_EVENT_FOOTSTEP_LEFT" ) || V_strcmp( pszEventName, "CL_EVENT_FOOTSTEP_RIGHT" ) )
+	{
+		if ( !engine->IsInGame() || !_cl_classmenuopen.GetBool() )
+			return;
+
+		soundlevel_t iSoundlevel = SNDLVL_NONE;
+
+		EmitSound_t es;
+		es.m_nChannel = CHAN_STATIC;
+		es.m_flVolume = 1;
+		es.m_SoundLevel = iSoundlevel;
+		es.m_flSoundTime = gpGlobals->curtime;
+		es.m_bEmitCloseCaption = false;
+		es.m_pSoundName = pszEventOptions;
+
+		C_RecipientFilter filter;
+		C_BaseEntity::EmitSound(filter, SOUND_FROM_UI_PANEL, es);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


JoriKos:
I think the Spy continually smoking his cigarette looks much more natural, the proposed fix looks quite awkward.

num-get:
> I think the Spy continually smoking his cigarette looks much more natural, the proposed fix looks quite awkward.

The animation was like that before the MVM update and on console TF2. <del>From what I understand, the class_select.vcd that handles the animation was broken since day 1 but it was unused up until 2012.</del>

Edit: I double check the 2007 build of TF2. Apparently the class_select.vcd is in use for Spy, but the .vcd is broken in a different way.
Spy will play his class select animation once normally, play his idle animation for about 12 seconds as normal. After that, the animation will skip to the part where Spy taking off the cigarette from his mouth twice, and then loop his animation where he's repeatedly huffing his cigarette like current TF2. It's broken in a way that nobody will notice because nobody stares at the class menu that long.
Interestingly, 2007 build has additional facial animation where Spy blows the cigarette smoke out of his mouth which is absent in current TF2. I will add that to the .vcd file later.

Chickenman456:
I've been waiting for a fix like this for years. Holding out hope this gets merged

Pygargue00fr:
This NEEDS to be added to the game. Presentation is important and this had been bugging me for a decade.

mastercoms:
I am a bit unhappy about my fix to the class menu so far and wanted to do more work on it, which is why I didn't submit it yet. This PR is also missing the subsequent fixes to the class menu I had done, including stopping the select sounds when the class menu is closed.

num-get:
> including stopping the select sounds when the class menu is closed.

I added an extra check to avoid that.

mastercoms:
Yes, but sounds that start while selecting a class, will continue playing after you select the class, causing errant foley sounds to play for up to a few seconds while in-game.

num-get:
> Yes, but sounds that start while selecting a class, will continue playing after you select the class, causing errant foley sounds to play for up to a few seconds while in-game.

Like I said, I added an extra check to avoid that.

mastercoms:
The check you do prevents playing new sounds to start playing when the class menu is closed, it does not stop already playing sounds.

num-get:
> https://github.com/ValveSoftware/source-sdk-2013/pull/1991#issuecomment-5206650122

I don't see the point.
https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/game/client/game_controls/basemodelpanel.cpp
The original 2007 class menu code didn't have any specific function to interrupt already played sound effect.

JoriKos:
It would be an improvement over what we used to have then. Don't see a reason to not have it.

PR #1997, opened 2026-08-15
[All] fix trigger_wind brush rotation

SirYodaJedi:
## Description
Fixes "angles" KV conflict, having been used for both wind direction and brush rotation (as discussed on VDC page). 
* A new "WindAngle" KV has been added, which maps directly to the preexisting `m_nDirBase` variable. Wind direction is modified by brush angles if using the new method (ensures yaw-rotated instances work correctly).
* A new "SetWindAngle" input has been added, which properly updates the wind angle.
* `m_nDirBase` now defaults to `-1` if absent from the entity lump, which reverts to the old method.

## Legalese
All code in this PR was written by me based upon existing Source SDK code. Permission is granted to use it as is laid out in the LICENSE and CONTRIBUTING files.

### DIFF
--- a/src/game/server/triggers.cpp
+++ b/src/game/server/triggers.cpp
@@ -3821,6 +3821,7 @@ public:
 	// Input handlers
 	void	InputEnable( inputdata_t &inputdata );
 	void	InputSetSpeed( inputdata_t &inputdata );
+	void	InputSetWindAngle( inputdata_t &inputdata);
 
 private:
 	int 	m_nSpeedBase;	// base line for how hard the wind blows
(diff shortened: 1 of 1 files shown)


src/game/server/triggers.cpp
```diff
@@ -3866,18 +3867,38 @@ BEGIN_DATADESC( CTriggerWind )
...
 
 	DEFINE_INPUTFUNC( FIELD_INTEGER, "SetSpeed", InputSetSpeed ),
+	DEFINE_INPUTFUNC( FIELD_INTEGER, "SetWindAngle", InputSetWindAngle ),
 
 END_DATADESC()
 
+//------------------------------------------------------------------------------
+// Purpose: Constructor for trigger_wind
+//------------------------------------------------------------------------------
+CTriggerWind::CTriggerWind()
```
Bitl: CTriggerWind() is not defined in the header, causing an error to be generated when compiling.

add CTriggerWind(); under DECLARE_DATADESC(); on line 3802

details: https://github.com/BitlDevelopmentStudios/source-sdk-2013/commit/a788c74261f0782c20bec0a1fbb3e68bc06fe231

PR #2005, opened 2026-08-19
[TF2] Fix Half-Zatoichi being invisible when held by disguised Spy

Piogre:
Currently, if a Spy is disguised as a Soldier or Demoman with the Half-Zatoichi equipped, and has their disguise weapon set to the melee slot, the Half-Zatoichi model will not appear to enemy players, leaving the disguised Spy holding nothing (a major giveaway that they are a Spy).

This is because the Half-Zatoichi is (as far as I'm aware) the only weapon to make use of model_player_per_class in its item definition, and the code to fetch the world model for the weapon just calls for it based on the owner's class (which in this case is Spy, not Soldier or Demoman).

This PR fixes this issue, so that a disguised Spy holding the Half-Zatoichi as the disguise weapon will load the correct katana model for the disguise's class.  It also forces the Demoman disguise to use the correct animation for the Half-Zatoichi.

sollybefore
sollyafter
demobefore
demoafter

### DIFF
--- a/src/game/shared/tf/tf_weapon_sword.cpp
+++ b/src/game/shared/tf/tf_weapon_sword.cpp
@@ -582,6 +582,14 @@ int CTFKatana::GetActivityWeaponRole() const
 		// demo should use act table item1
 		return TF_WPN_TYPE_ITEM1;
 	}
+	else if ( pPlayer && pPlayer->GetPlayerClass()->GetClassIndex() == TF_CLASS_SPY )
+	{
+		CTFPlayer* pDisguiseTarget = pPlayer->m_Shared.GetDisguiseTarget();
+		if ( pDisguiseTarget && pDisguiseTarget->GetPlayerClass()->GetClassIndex() == TF_CLASS_DEMOMAN ) {
+			// demo should use act table item1
+			return TF_WPN_TYPE_ITEM1;
+		}
+	}
 
 	return BaseClass::GetActivityWeaponRole();
 }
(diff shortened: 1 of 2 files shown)


Ashe-tf:
Disguised Demo is holding the katana wrong

Chickenman456:
> Disguised Demo is holding the katana wrong

That's just how he uses it in-game I'm pretty sure

Piogre:
EDIT: see below, the animation is now fixed as well

> That's just how he uses it in-game I'm pretty sure

no, no it's not.

in-game he holds it two-handed like the eyelander.  I didn't catch it because there's a separate issue that causes him to hold it one-handed in the lower-left panel (along with loadout screen, see https://github.com/ValveSoftware/Source-1-Games/issues/1971) -- I didn't catch it because it matched the lower-left panel so nothing seemed inconsistent

technically a separate issue but it may be that both are fixable; I'll take a look

Chickenman456:
> > That's just how he uses it in-game I'm pretty sure
> 
> no, no it's not.
> 
> in-game he holds it two-handed like the eyelander. I didn't catch it because there's a separate issue that causes him to hold it one-handed in the lower-left panel (along with loadout screen, see [ValveSoftware/Source-1-Games#1971](https://github.com/ValveSoftware/Source-1-Games/issues/1971)) -- I didn't catch it because it matched the lower-left panel so nothing seemed inconsistent
> 
> technically a separate issue but it may be that both are fixable; I'll take a look

Ah I see. I had no idea that issue was exclusive to the loadout screens!

Piogre:
> Disguised Demo is holding the katana wrong

added a fix for this, good catch ty

PR #2008, opened 2026-08-21
[TF2] Restore Huntsman Taunt Loop Bug

antielectronbomb:
# Description
This PR restores the ["Staring Contest"](https://wiki.teamfortress.com/wiki/Arrow_Stab#cite_note-1) infinite stun loop bug with the Huntsman or the Fortified Compound as mentioned by the TF2 Wiki: 
> Because of this, two Snipers wielding the Huntsman or Fortified Compound can create an infinite stun loop without killing each other, commonly known as a "Staring Contest".[[1]](https://wiki.teamfortress.com/wiki/Arrow_Stab#cite_note-1)

The TF2 Wiki page for the Arrow Stab taunt kill does not mention that this bug has been fixed, and was a beloved part of the community back in the day as shown by the number of videos about it such as this: https://www.youtube.com/watch?v=b7XDbzxpGJM

To perform the Staring Contest bug, two players each from opposing teams equipped with the Huntsman/Fortified Compound must be present. Next, both players taunt at around the same time with the bow. After both players are stunned, the taunt key is immediately and repeatedly pressed quickly over and over until one player eventually tires out from repeatedly pressing the taunt button and loses the Staring Contest.

If you try to do the Staring Contest bug in live TF2, both Snipers do get stunned by each other but the loop does not happen. You can test this with a puppet bot with zero ping and no matter how much you try to mash the taunt key, you are not able to do the Staring Contest bug.

I believe this bug was inadvertently fixed sometime around 2013 (or 2017?) based on patch notes, video comments, and historical videos related to the taunt. I still have yet to confirm when exactly was the bug fixed. What I am sure of is an unrelated change fixed it.

This PR tries to recreate the Staring Contest taunt loop bug by setting m_flTauntNextStartTime to the current time, so that the player is able to do the taunt and stun another player again even if stunned by another player doing the same thing.

Video demonstration of the taunt bug restoration: https://youtu.be/9UMqfykEsP4
Before:

https://github.com/user-attachments/assets/418354c6-e82f-4611-9f76-c028aeb4771e

After:

https://github.com/user-attachments/assets/7476b89d-48d1-4a52-aa45-bcfa4c4bac3c

Credits to VerdiusArcana for the bug restoration via the SourceMod weapon reverts plugin

### DIFF
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -18081,6 +18081,7 @@ void CTFPlayer::Taunt( taunts_t iTauntIndex, int iTauntConcept )
 		if ( !V_stricmp( szResponse, "scenes/player/sniper/low/taunt04.vcd" ) )
 		{
 			m_flTauntAttackTime = gpGlobals->curtime + 0.85f;
+			m_flTauntNextStartTime = gpGlobals->curtime;
 			m_iTauntAttack = TAUNTATK_SNIPER_ARROW_STAB_IMPALE;
 		}
 	}
--- a/src/game/server/tf/tf_player.cpp
+++ b/src/game/server/tf/tf_player.cpp
@@ -297,6 +297,8 @@ extern ConVar mp_developer;
 extern ConVar bot_mimic;
 #endif // _DEBUG || STAGING_ONLY 
 
+ConVar tf_compound_bow_taunt_loop_bug("tf_compound_bow_taunt_loop_bug", "1", FCVAR_NOTIFY, "Restore the compound bow's old taunt attack loop bug behavior when enabled");
+
 extern CBaseEntity *FindPickerEntity( CBasePlayer *pPlayer );
 extern bool CanScatterGunKnockBack( CTFWeaponBase *pWeapon, float flDamage, float flDistanceSq );
 extern bool IsCustomGameMode();
@@ -18081,7 +18083,6 @@ void CTFPlayer::Taunt( taunts_t iTauntIndex, int iTauntConcept )
 		if ( !V_stricmp( szResponse, "scenes/player/sniper/low/taunt04.vcd" ) )
 		{
 			m_flTauntAttackTime = gpGlobals->curtime + 0.85f;
-			m_flTauntNextStartTime = gpGlobals->curtime;
 			m_iTauntAttack = TAUNTATK_SNIPER_ARROW_STAB_IMPALE;
 		}
 	}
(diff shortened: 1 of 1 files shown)


GrubPL:
Why?

Professor-Heavy:
This bug spawned a finalist entry for Funniest Replay in the First Annual Saxxy Awards. Given how completely inconsequential this bug actually was I'd also like to see it restored.

It's not really pressing of an issue, there are other issues and PRs that take priority, but would still be neat to see in.

antielectronbomb:
> Why?

Because it used to be fun and did not really affect gameplay back then. It was also part of the Saxxy Awards too.
It got fixed where nobody noticed it.

Bitl:
If you absolutely must reintroduce a bug, it probably would be better to add a ConVar that toggles it on and off for server hosters. It should also be disabled by default.

look at tf_overtime_nag for an example

JoriKos:
If there is a ConVar I do think it should be enabled in Casual Mode by default too.

FlaminSarge:
Rather than setting next start taunt time when the taunt is performed, it may be better to reset next start taunt time on the victim when they're hit by the stun from the other player's taunt instead.
The way you currently have it, if the player has a way to instantly interrupt their taunt, they will be able to spam the Huntsman taunt, which shouldn't have that many side effects but may be safer to prevent (the next start taunt time var was added to prevent spawning a bunch of objects by interrupting the Shred Alert/other entity-spawning taunts, so you're giving the Huntsman taunt a pass to be able to do that again, which shouldn't matter since it doesn't spawn anything but better to be defensive in the behavior).

antielectronbomb:
Added a cvar called tf_compound_bow_taunt_loop_bug that allows this behavior. I am not sure if this is the best name for such esoteric bug. m_flTauntNextStartTime has also been set onto the victim instead of the attacker.

For anyone who thinks restoring this bug as "risky" (for lack of better word), the m_flTauntNextStartTime method in restoring the bug has been practically play tested and used for [months](https://github.com/rsedxcftvgyhbujnkiqwe/castaway-plugins/pull/338) in server-side SourceMod Weapon Revert community servers in TF2.

src/game/server/tf/tf_player.cpp
FlaminSarge, line 18524: While it would also catch the Engineer taunt, I think it's safer to only set the victim start taunt time if the StunPlayer actually gets called.
Also allows you to rename the cvar to account for "let taunt stuns interrupt the next taunt timer", so to speak. I don't see any reason the other taunt stuns shouldn't also gain the behavior. Could potentially have different values of the cvar affect which taunts get it, too. Not sure whether Ubersaw should get it since it's a different code branch though.

antielectronbomb: I would do it for all taunts that have a similar stun effect like the Engineer's Gunslinger taunt if I can confirm it with historical sources.

From my memory I am not sure if it was even possible with other taunts back then.

FlaminSarge: The taunt loop itself wasn't necessarily possible due to timings, but stuns interrupting the taunt and allowing a second taunt were definitely possible.

neobenedict:
+1 i miss this every day

PR #2009, opened 2026-08-21
[TF2] Restore sticky detonation while taunting

num-get:
# Description
Sticky detonation while taunting was originally an intentional feature added in <a href=https://store.steampowered.com/oldnews/2529>Sniper VS Spy update</a>. Around 2014-2015, in order to accommodate Mannrobic taunt that triggers special animation via primary/secondary fire key, CTFPlayer::CanAttack function blocks the ability to attack while taunting. This in turn removes Demoman's ability to detonate his stickies during a taunt. This removal was never explicitly addressed in any of the patch note.
I have added a special attack flag TF_CAN_ATTACK_FLAG_PIPEBOMBLAUNCHER that restores Demo's ability to use his secondary fire key while taunting.

https://github.com/user-attachments/assets/9a7ee880-d593-4e3c-9a20-5d7daece2d8d

### DIFF
--- a/src/game/client/tf/c_tf_player.cpp
+++ b/src/game/client/tf/c_tf_player.cpp
@@ -6661,6 +6661,16 @@ bool C_TFPlayer::CreateMove( float flInputSampleTime, CUserCmd *pCmd )
 			VectorCopy( angMoveAngle, pCmd->viewangles );
 		}
 
+		// Re-add IN_ATTACK2 if player is Demoman with sticky launcher. This is done so they can detonate stickies while taunting.
+		if ( (nCurrentButtons & IN_ATTACK2) && IsPlayerClass( TF_CLASS_DEMOMAN ) )
+		{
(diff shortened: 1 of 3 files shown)


JoriKos:
This would make infinite taunts like the High Five give even more of an advantage than they already do.

Whurrhurr:
make it so that non-stock taunts/partner taunts (ones you have to equip in order to use in the case of "non-stock taunts") can't let demoman detonate his stickies then

JoriKos:
That would just be a weird inconsistency, it's better to just leave the behaviour as-is.

dilbertron2:
Tying this behavior to a ConVar that is disabled by default (Disabling means it keeps the current behavior) would be a good substitute.

ElDous1:
Really should only be possible on stock taunts. Or either just not possible in looping taunts, it was removed because you could see through corners with a looping taunt. rembember you can cancel inmediatly looping taunts too

Issue #1135, opened 2025-03-21
[TF2] Game crashes after damaging Robot Destruction robots


Tkain:
On a freshly-built copy of the TF2 SDK, when damaging robots in the Robot Destruction game mode, the game will crash.

The particular error that causes the crash has code `0x80000003`. (This is usually triggered by choosing to break after an assertion failure, but no assertion failure dialog is shown before this error.)

The crash seems to occur within engine code that is called at the following line:

https://github.com/ValveSoftware/source-sdk-2013/blob/a62efecf624923d3bacc67b8ee4b7f8a9855abfd/src/game/client/cdll_client_int.cpp#L268

nvidiacorp:
Hello I have invested alot of time on this since this is my favourite gamemode. It's fixed on my copy I think it was caused by the empty network table please try my changes here that setup a dummy variable:
https://github.com/nvidiacorp/tf2-rd-fix/tree/master

Please let me know if this solves the heap corruption for you too.

PR #2011, opened 2026-08-22
[TF2] Fix Robot Destruction heap corruption in source sdk 2013

nvidiacorp:
This pull request adds a boolean network variable to entity_bonuspack that solves the heap corruption caused on robot destruction for the tf2 sdk. This changes makes it function as expected on my end in testing.


nvidiacorp:
This solves the issue mentioned in https://github.com/ValveSoftware/source-sdk-2013/issues/1135

bakugo:
While this does indeed prevent the heap corruption, it doesn't address the underlying cause.

I believe I have a proper fix, but I still need to investigate a bit more.

nvidiacorp:
Absolutely fair, I made this a pull request in hopes of having more eyeballs on the issue as well as giving people out there a solution for the time being.

PR #2022, opened 2026-08-25
[TF2] Fix voice muting causing the user to be kicked for sending too many commands

### LINKED ISSUE #8237: [TF2] Muting a large number of players causes the user to be kicked from the server
There's an issue with the "Mute Players" UI in TF2 where if you select multiple users, the muting will be done one-by-one by sending individual commands to the server. If you're playing in a large server with e.g. 64 players, this will consistently cause you to get kicked for sending too many commands to the server.

Steps to reproduce:
- Join a large server
- Open the Mute Players UI
- Select every user and press "Mute". You will get kicked instantly.

Screenshot_20260825_122803

rockbruno:
Currently there's an issue with the "Mute Players" UI in TF2 where if you select multiple users, the muting will be done one-by-one by sending individual commands to the server. If you're playing in a large server with e.g. 64 players, this will consistently cause you to get kicked for sending too many commands to the server.

This PR fixes this logic by adding a variant to `SetPlayerBlockedState` that receives a list of users, sending the update only on the last entry rather than for every user. The reason this works is because `UpdateServerState` always sends the info for every user, so there's no need to call it in-between.

How to reproduce the issue:
- Join a large server (or, if testing locally, change mute_player_dialog.cpp to not filter bots)
- Open the Mute Players UI
- Select every user and press "Mute"
- Result: You will get kicked instantly. This PR fixes this.

I was able to validate this fix by temporarily changing the code to allow bots to be muted.

Screenshot_20260825_122803

Fixes https://github.com/ValveSoftware/Source-1-Games/issues/8237

### DIFF
--- a/src/game/shared/voice_status.cpp
+++ b/src/game/shared/voice_status.cpp
@@ -546,33 +546,59 @@ bool CVoiceStatus::IsLocalPlayerSpeaking( void )
 
 //-----------------------------------------------------------------------------
 // Purpose: blocks/unblocks the target client from being heard
-// Input  : playerID - 
-// Output : Returns true on success, false on failure.
+// Input  : playerID, block state
 //-----------------------------------------------------------------------------
 void CVoiceStatus::SetPlayerBlockedState(int iPlayer, bool blocked)
 {
-	if (voice_clientdebug.GetInt())
-	{
-		Msg( "CVoiceStatus::SetPlayerBlockedState part 1\n" );
-	}
+	CUtlVector<VoiceBlockState_t> playerState( 1 );
+	VoiceBlockState_t state;
+	state.playerIndex = iPlayer;
+	state.blocked = blocked;
+	playerState.AddToTail( state );
+	SetPlayersBlockedState( playerState );
+}
 
-	player_info_t pi;
-	if ( !engine->GetPlayerInfo( iPlayer, &pi ) )
+//-----------------------------------------------------------------------------
+// Purpose: blocks/unblocks the target client(s) from being heard
+// Input  : vector of playerID, block state
+//-----------------------------------------------------------------------------
+void CVoiceStatus::SetPlayersBlockedState( const CUtlVector<VoiceBlockState_t> &playerStates )
+{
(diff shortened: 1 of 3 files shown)


src/game/shared/voice_status.cpp
```diff
@@ -546,33 +546,59 @@ bool CVoiceStatus::IsLocalPlayerSpeaking( void )
...
+	SetPlayersBlockedState( playerState );
+}
 
-	player_info_t pi;
-	if ( !engine->GetPlayerInfo( iPlayer, &pi ) )
+//-----------------------------------------------------------------------------
+// Purpose: blocks/unblocks the target client(s) from being heard
+// Input  : vector of playerID, block state
+//-----------------------------------------------------------------------------
+void CVoiceStatus::SetPlayersBlockedState( const CUtlVector<VoiceBlockState_t> &playerStates )
```
rockbruno, line 565: Note: You may notice the logic ignores the blocked field. The original also did that. I considered removing the field from the API entirely, but that would imply changing all callers and I wasn't sure that was a good idea for this PR

src/game/shared/voice_status.cpp
```diff
@@ -546,33 +546,59 @@ bool CVoiceStatus::IsLocalPlayerSpeaking( void )
...
+		}
+
+		int iPlayer = playerStates[i].playerIndex;
+		player_info_t pi;
+		if ( !engine->GetPlayerInfo( iPlayer, &pi ) )
+			continue;
+
+		if (voice_clientdebug.GetInt())
+		{
+			Msg( "CVoiceStatus::SetPlayerBlockedState part 2\n" );
```
kvog-git, line 585: Did you intend to include this?

rockbruno: Yes, the original code also had them, it just looks weird on the diff view because of the added indentation. https://github.com/rockbruno/source-sdk-2013/blob/c22529f725414dbb857ace0f170013ab32d35fff/src/game/shared/voice_status.cpp#L552-L576

PR #2023, opened 2026-08-25
ALU optimization by approximating weighting functions in Bicubic Lightmaps

JohnMega:
We can remove the two divisions from the `h0` and `h1` functions by approximating them, thus optimizing the ALU. This will have no impact on the visual quality.

Reference: [Real-Time Samurai Cinema](https://www.glowybits.com/talks/real-time_samurai_cinema/real-time_samurai_cinema.pdf)

Details are in the Desmos graph (www.desmos.com/calculator/7dcech96ju).

image


JohnMega:
@misyltoad Since you're familiar with this code, could you take a look? It's just two lines here.

PR #2024, opened 2026-08-26
[TF2] Fix team color of attachments and Ubersaw vial in inventory inspect and backpack icons

Piogre:
Currently, the behavior of weapon skin team color for generated backpack icons and right click -> inspect in the inventory is as follows:

-If the weapon is a Decorated/War Painted weapon whose paint has different team-colored textures, the preview uses the team color selected via the buttons in the preview screen (and defaults to red team in the backpack icons).
-Otherwise, the preview uses the weapon's skin for the team the player is currently assigned.  If the player is not assigned a team (including if they are not connected to a server), this defaults to red team.

This behavior has a couple small bugs within it.  First of all, attachments such as the stat clock and festivizer do not get properly assigned to the correct team-colored skin, instead always showing as red in these contexts, even if the weapon they're attached to is displaying blue team textures.  Second, the vial of the Ubersaw, unlike the weapon base texture (which gets overwritten by the war paint compositor) retains its per-team coloring for many war paints, which works correctly in-game but can result in either red textures with a blue vial or blue textures with a red vial when previewed or seen in a backpack icon (depending on the player's current team).

The PR retains the team-color-skin prioritization described above (selected paintkit skin (defaulting red for icons), otherwise current team, otherwise red), but fixes these holes in the preview, properly setting the team skin for attachments and the Ubersaw vial.

inventorybefore
inventoryafter
flarebefore
flareafter
smgbefore
smgafter
nutuberbefore
nutuberafter
sniperbefore
sniperafter
teamuberbefore
teamuberafter

Please note, the right-click menu in the Inventory is disabled entirely by default in the SDK, via a single return statement at the top of CBackpackPanel::OpenContextMenu() in src\game\client\econ\backpack_panel.cpp -- testing this code change for right click -> inspect within the SDK requires commenting out that line.

### DIFF
--- a/src/game/client/econ/item_model_panel.cpp
+++ b/src/game/client/econ/item_model_panel.cpp
@@ -17,6 +17,7 @@
 #include "vgui_controls/Label.h"
 #include "vgui_controls/Button.h"
 #include "econ_item_system.h"
+#include "econ_paintkit.h"
 #include "ienginevgui.h"
 #include "VGuiMatSurface/IMatSystemSurface.h"
 #include "renderparm.h"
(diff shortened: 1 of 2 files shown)


Piogre:
Someone pointed out on another platform that it might be preferable to just always use red team for uniformity with static backpack icons.  This might be a valid decision, but it's a design decision that I didn't decide to make for this PR whose purpose was to fix inconsistencies within a single weapon's preview.

If Valve wanted, they could change the one line 412 in src/game/client/econ/item_model_panel.cpp:

`int iTeam = GetLocalPlayerTeam();`

to

`int iTeam = TEAM_UNASSIGNED;`

which would make the icons ALWAYS red, and inspect panel be red unless it's team-colored paintkit in which case it's selectable (and all elements follow the selection)

src/game/client/econ/item_model_panel.cpp
```diff
@@ -408,6 +409,52 @@ void CEmbeddedItemModelPanel::SetItem( CEconItemView *pItem )
 					}
 				}
 
+				int iTeam = GetLocalPlayerTeam();
```
Piogre, line 412: Changing this to:

`int iTeam = TEAM_UNASSIGNED`

Would make generated icons ALWAYS be red, and right click -> inspect would be red unless it's a team-colored paintkit (in which case it's selectable).  attachments and ubersaw vial would still all match the selection.

This might be preferable as it makes festivized weapon icons always consistent with the static non-generated backpack icons, and makes team-colored decorated weapon icons consistent with their non-team-colored counterparts (for the purposes of attachments etc).  It's a design decision which I elected not to make for this PR but it remains an option to improve visual consistency even more.

(a chunk of other code below could also be cut out in this case but just changing this line should still work)

PR #2025, opened 2026-08-26
[ALL] Add deletion of downloaded manifest files from Servers

chocochocotorta:
When connecting to a Community Server, this one can provide downloads of any type (in this case, .pcf files for particles and a txt inside maps that contains a manifest for particles or level sounds in a specific map).
While this is not an issue in custom maps that already packs it's own manifest file inside because it takes priority over the downloaded one, a server could maliciously send lots of particles manifest files for all the base game maps with a pcf that floods the "ParticleEffectNames" string table on precache (making some of the default particles from the base game not being precached and display the default error particle), mute/replace certain sounds, or crash the client when creating a Local Server.

The PR checks for files ending in particles.txt or level_sounds.txt inside download/maps and deletes them on game exit (like tf_delete_temp_files).
Tested both in TF2 & HL2:DM.

### DIFF
--- a/src/game/client/clientmode_shared.cpp
+++ b/src/game/client/clientmode_shared.cpp
@@ -409,6 +409,8 @@ void ClientModeShared::VGui_Shutdown()
 //-----------------------------------------------------------------------------
 void ClientModeShared::Shutdown()
 {
+	RemoveDownloadedMapManifests( "particles" );
+	RemoveDownloadedMapManifests( "level_sounds" );
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


neobenedict:
If a malicious server forces download of multiple particle files this will only remove the first?

Better is just to reject "_particles.txt" suffix in the client download check but that's in engine.dll.

chocochocotorta:
> If a malicious server forces download of multiple particle files this will only remove the first?

In case the server does this, the client will just delete all files containing _particles.txt inside download/maps/.

> Better is just to reject "_particles.txt" suffix in the client download check but that's in engine.dll.

That could also work, but i'm not sure how good would be for Valve to add that specific case inside the engine.
I made it this way so that servers could precache particles to the client and to prevent the client issue of having to deal with a server that sends so many particles (i based it the same way as the ConVar tf_delete_temp_files that got added on Scream Fortress XIII, if needed, i could make the logic only remove on Shutdown instead of server join/changelevel or server leave).

KaelaSavia:
I recommend also doing same for following files:
```
_level_sounds.txt
```
For example:
```
cp_manor_event_level_sounds.txt
```
As these files can also be abused.

I'm personally against blocking things in engine.dll, as there are legitimate uses for uploading _level_sounds.txt or _particles.txt that is not packed into map.  Removing these files just on disconnect would be most flexible in my opinion.

chocochocotorta:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/2025#issuecomment-5481357993

I'll make the PR as draft and implement the same thing for level sounds too, it'll take me a bit of time since I'm outside.

chocochocotorta:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/2025#issuecomment-5481357993

Added to also delete _level_sounds.txt when closing the game (i tried to rename the branch and it accidentally got deleted from the PR).

PR #2028, opened 2026-08-27
Add sv_rough_landing_effects

kusabana:
# Description
Many community servers in CS:S, especially those movement oriented (like bhop, surf, kz, trikz) disable rough landing effects on the server via plugins, but this results in the client predicting it and then snapping back.

I opted to make it a convar to disable all the effects, while most servers only disable the view punch, I'd be interested to hear if the community has any strong opinions about that.

### DIFF
--- a/src/game/shared/gamemovement.cpp
+++ b/src/game/shared/gamemovement.cpp
@@ -3984,31 +3984,36 @@ void CGameMovement::CheckFalling( void )
 	player->m_Local.m_flFallVelocity = 0;
 }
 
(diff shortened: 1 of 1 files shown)


JoriKos:
> I opted to make it a convar to disable all the effects, while most servers only disable the view punch

Couldn't you make the convar have multiple values? 0 disables everything, 1 enables everything, 2 only enables everything but the view punch, etc.

kusabana:
> > I opted to make it a convar to disable all the effects, while most servers only disable the view punch
> 
> Couldn't you make the convar have multiple values? 0 disables everything, 1 enables everything, 2 only enables everything but the view punch, etc.

Definitely could, it was more a matter of if people wanted it or not. Personally I'm just interested in getting the view punch difference predicted, and don't care if it plays the loud footstep sound.

PR #2029, opened 2026-08-27
Fix trigger_push prediction

LilinhoOGatinho:
Removes CPrediction::CheckMovingGround.

After the addition of CheckMovingGround to client code in the sdk update, push triggers have had their prediction broken. Before simulating player movement, this function will check whether FL_BASEVELOCITY isn't set, and if that is so, it will apply an impulse of one second plus half a frame of the player's current base velocity, set base velocity to 0 and remove FL_BASEVELOCITY. This is fine on the server, because after player movement the trigger_push will reapply the base velocity and FL_BASEVELOCITY flag, making it so that the CheckMovingGround code won't run until the player leaves the push trigger, but on the client side there's nothing to set the player's base velocity and add the flag. So after CheckMovingGround resets base velocity for the first time, the client will simulate movement as if the player has no base velocity until it receives another update from the server, resulting in prediction errors.

With CheckMovingGround:

https://github.com/user-attachments/assets/9da110f1-be24-42a1-821d-e16fcf7e4225

Without CheckMovingGround:

https://github.com/user-attachments/assets/2948aef2-5416-4b75-aeac-4a8442c757b4

### DIFF
--- a/src/game/client/prediction.cpp
+++ b/src/game/client/prediction.cpp
@@ -836,44 +836,6 @@ void CPrediction::RunPostThink( C_BasePlayer *player )
 }
 
 //-----------------------------------------------------------------------------
-// Purpose: Checks if the player is standing on a moving entity and adjusts velocity and 
-//  basevelocity appropriately
-// Input  : *player - 
-//			frametime - 
-//-----------------------------------------------------------------------------
-void CPrediction::CheckMovingGround( C_BasePlayer *player, double frametime )
-{
-#if 0
-	CBaseEntity	    *groundentity;
-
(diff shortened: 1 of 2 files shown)


speedvoltage:
I was going to open a PR with a similar outcome -- predicting trigger_push, but opening it now would conflict with this PR. At any rate, for comparison, what I had: https://github.com/speedvoltage/hl2dm-fixes/commit/15dfd147fb35e0f1fd8127c08127e89df4fb28d5

LilinhoOGatinho:
Your fix looks a lot better, so I'll close this one and you can submit it @speedvoltage

PR #2031, opened 2026-08-28
[ALL] Implement prediction for forced and disabled player input masks

speedvoltage:
Issue:

The server can force or disable player buttons without telling the owning client. The client then predicts movement using inputs that the server changes.

Fix:

Send the button masks to the owning client and apply them to the existing prediction command context before client-side simulation.

### DIFF
--- a/src/game/client/prediction.cpp
+++ b/src/game/client/prediction.cpp
@@ -197,7 +197,6 @@ void CPrediction::CheckError( int commands_acknowledged )
 	}
 #endif
 }
-
 //-----------------------------------------------------------------------------
 // Purpose: 
 //-----------------------------------------------------------------------------
@@ -397,7 +396,6 @@ bool CPrediction::ShouldDumpEntity( C_BaseEntity *ent )
 	{
 		if ( cl_predictionentitydumpbyclass.GetString()[ 0 ] == 0 )
 			return false;
-
 		if ( !FClassnameIs( ent, cl_predictionentitydumpbyclass.GetString() ) )
 			return false;
 	}
@@ -597,7 +595,6 @@ void CPrediction::PostNetworkDataReceived( int commands_acknowledged )
 	}
 #endif
 #endif
-
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


src/game/client/prediction.cpp
```diff
@@ -888,6 +888,11 @@ void CPrediction::RunCommand( C_BasePlayer *player, CUserCmd *ucmd, IMoveHelper
 	Q_snprintf( sz, sizeof( sz ), "runcommand%04d", ucmd->command_number );
 	PREDICTION_TRACKVALUECHANGESCOPE( sz );
 #endif
+	CUserCmd effectiveCmd( *ucmd );
+	effectiveCmd.buttons |= player->m_afButtonForced;
```
JustYourOrdinaryNormie: why create a whole new usercmd for this?

speedvoltage: The copy was to avoid permanently modifying the `CUserCmd` stored in `CInput`'s command history, since `buttons` is part of its verified checksum. Prediction needs downstream code and `GetCurrentCommand()` to see the server's effective forced/disabled buttons, while the original command still needs to remain intact. If the underlying assumption behind that question is that copying the entire `CUserCmd` is unnecessary, you are probably right. Saving `ucmd->buttons`, applying the masks in place for the duration of `RunCommand`, then restoring it after `FinishCommand()` should accomplish the same thing without copying the whole command. I have pushed the changes and obtained the same results overall.

JustYourOrdinaryNormie: the checksum verification happens during writeusercmddeltatobuffer before prediction so the usercmd will have already been sent off. the only verification that occurs again is in getusercmd and that only compares the sequence number to the usercmd in that slot. thats why restoring buttons is kinda pointless

speedvoltage: Yep, agreed. I traced it further and `RunCommand()` is already operating on `ctx->cmd`, which is the prediction copy. The current branch now just applies the masks directly to that command with no additional copy or restoration.

PR #2038, opened 2026-08-29
Add cvar tf_contract_mvm_show

kvog-git:
Hi. This gives the player the option to hide the contract progress widget in MvM, where there are currently no contracts.


kvog-git:
I thought about it more and this is useless so I'm going to close it.

PR #2046, opened 2026-09-01
[TF2] Revert Tomislav's Firing Sounds Pitch

### LINKED ISSUE #1674: [TF2] Tomislav Firing Sounds Pitch is Low-Pitched
Before the Love & War patch, the Tomislav used to have higher pitched sounds for firing and deploying. The sounds changed because the pitch ended up getting controlled by the 20% slower firing speed.

For comparison, the original sounds for the Tomislav sounded like this:

[Pre-Love and War Tomisalv Sounds](https://www.youtube.com/watch?v=2YgirY6n-vc)

And for the Tomislav today:
[Current Tomislav](https://www.youtube.com/watch?v=qHfbDjv2qvc)
Notice how the current Tomislav has a lower pitch for its deploying, firing, and stopping sounds.

I believe this is a bug because this change was undocumented (as noted in the [Official TF2 Wiki](https://wiki.teamfortress.com/wiki/June_18,_2014_Patch)), and the fact that the original sound files for the Tomislav is actually higher pitched than the sounds we currently hear in-game.

### LINKED ISSUE #4734: [TF2] Tomislav firing sound is inconsistent with it's actual fire rate.

antielectronbomb:
**_Works in Mann vs. Machine_**

## YouTube Video Explanation
(note: video shows outdated devlogs in the upper left corner of the screen, see commits instead)
[Fixing the Tomislav's Sounds in TF2](https://www.youtube.com/watch?v=27k3fU-jiKs)

# How to Fix:
See https://github.com/ValveSoftware/source-sdk-2013/pull/2046#issuecomment-5557285053 for more context. 

These must be added into items_game.txt for this PR to work. 

Make a new attribute called `mult_minigun_pitch_scale`, add it to the Tomislav's attribute fields, and set `mult_minigun_pitch_scale` value to 1.2.

_**4000 is an example attribute number, change the attribute number as needed**_

```
	"4000"
	{
		"name"	"minigun pitch scale"
		"attribute_class"	"mult_minigun_pitch_scale"
		"description_string"	"unused"
		"description_format"	"value_is_percentage"
		"hidden"	"1"
		"effect_type"	"neutral"
		"stored_as_integer"	"0"
	}
```
In the weapon_tomislav field, add "minigun pitch scale" with a value of 1.2:

```
		"weapon_tomislav"
		{
			...
			"attributes"
			{
				"minigun spinup time decreased"
				{
					"attribute_class"	"mult_minigun_spinup_time"
					"value" "0.8"
				}
				"minigun no spin sounds"
				{
					"attribute_class"	"minigun_no_spin_sounds"
					"value" "1"
				}
				"fire rate penalty"
				{
					"attribute_class"	"mult_postfiredelay"
					"value"	"1.2"
				}
				"weapon spread bonus"
				{
					"attribute_class"	"mult_spread_scale"
					"value"	"0.8"
				}
				"minigun pitch scale"
				{
					"attribute_class"	"mult_minigun_pitch_scale"
					"value" "1.2"
				}
			}
		}
```

# Description
Before the Love & War patch, the Tomislav used to have higher pitched sounds for firing and deploying. The sounds changed because the pitch ended up getting controlled by the 20% slower firing speed.

Because of this change, this makes the Tomislav's sounds in-game not in-sync with the actual ammo drain, as well as having an incorrect pitch.

For comparison, the original sounds for the Tomislav sounded like this:

[Pre-Love and War Tomisalv Sounds](https://www.youtube.com/watch?v=2YgirY6n-vc)

And for the Tomislav today:
[Current Tomislav](https://www.youtube.com/watch?v=qHfbDjv2qvc)
Notice how the current Tomislav has a lower pitch for its deploying, firing, and stopping sounds.

I believe this is a bug because this change was undocumented (as noted in the [Official TF2 Wiki](https://wiki.teamfortress.com/wiki/June_18,_2014_Patch)), and the fact that the original sound files for the Tomislav is actually higher pitched than the sounds we currently hear in-game.

# Solution
This PR resolves this problem by correcting the pitch of the Tomislav to its intended pitch (and any other Minigun using the same mult_postfiredelay attribute as the Tomislav) while also preserving firing sound pitch adjustment when the firing speed is adjusted such as the firing speed upgrade in MvM.

This is done by setting flPitch back to 100. Previously, flPitch was set at 92.0 which resulted in sound de-synchronization with the ammo drain.

# Video Clips

**_Wearing headphones is a must!_**

## Before
Firing sounds:

https://github.com/user-attachments/assets/e54e1e3d-e363-4e96-84b5-754151989398

With MVM Firing speed upgrades:

https://github.com/user-attachments/assets/107a9044-60ad-48ef-bab1-b0fdf4e340c5

## After
Firing sounds:

https://github.com/user-attachments/assets/1f697b79-b569-4ec1-ba21-48adc992629c

With MVM Firing speed upgrades:

https://github.com/user-attachments/assets/555b0b77-073f-492a-a453-4a4ea09501cb

Fixes https://github.com/ValveSoftware/Source-1-Games/issues/4734 
Fixes https://github.com/ValveSoftware/source-sdk-2013/issues/1674

### DIFF
--- a/src/game/shared/tf/tf_weapon_minigun.cpp
+++ b/src/game/shared/tf/tf_weapon_minigun.cpp
@@ -1441,6 +1441,13 @@ void CTFMinigun::WeaponSoundUpdate()
 	if ( flSpeed != 1.0f )
 	{
 		flPitch = RemapValClamped( flSpeed, 1.5f, 0.5f, 80.f, 120.f );
+		// Tomislav - Fixes Tomislav's base firing sound pitch sounding lower than intended due to its base 20% slower firing speed penalty
+		if ( GetAttributeContainer()->GetItem()->GetItemDefIndex() == 424 )
+		{
+			//DevMsg( "Tomislav pitch correction: flSpeed BEFORE = %f, flPitch = %f\n", flSpeed, flPitch );
+			flPitch = RemapVal( flSpeed, 1.2f, 0.6f, 100.f, 120.f );
+			//DevMsg( "Tomislav pitch correction: flSpeed AFTER = %f, flPitch = %f\n", flSpeed, flPitch );
+		}
 	}
 
 	if ( m_bRageDraining )
--- a/src/game/shared/tf/tf_weapon_minigun.cpp
+++ b/src/game/shared/tf/tf_weapon_minigun.cpp
@@ -156,9 +156,6 @@ void CTFMinigun::WeaponReset( void )
 	m_iMinigunSoundCur = -1;
 	m_flMinigunSoundCurrentPitch = 1.0f;
 
-	m_flBasePostFireDelay = 1.0f;
-	CALL_ATTRIB_HOOK_FLOAT( m_flBasePostFireDelay, mult_postfiredelay );
-
 	StopMuzzleEffect();
 	StopBrassEffect();
 #endif
@@ -1443,16 +1440,7 @@ void CTFMinigun::WeaponSoundUpdate()
 	float flSpeed = ApplyFireDelay( 1.0f );
 	if ( flSpeed != 1.0f )
 	{
-		if ( m_flBasePostFireDelay != 1.0f )
-		{
-			// Pitch correction. Adjusts base firing sound pitch with the base firing speed.
-			// Fixes Tomislav's base firing sound pitch sounding lower than intended due to its base 20% slower firing speed penalty
-			// DevMsg( "Pitch correction: flSpeed BEFORE = %f, m_flBasePostFireDelay = %f, flPitch = %f\n", flSpeed, m_flBasePostFireDelay, flPitch );
-			flSpeed /= m_flBasePostFireDelay;
-		}
-
 		flPitch = RemapValClamped( flSpeed, 1.5f, 0.5f, 80.f, 120.f );
-		// DevMsg( "Pitch correction: flSpeed AFTER = %f, m_flBasePostFireDelay = %f, flPitch = %f\n", flSpeed, m_flBasePostFireDelay, flPitch );
 	}
 
 	if ( m_bRageDraining )
--- a/src/game/shared/tf/tf_weapon_minigun.cpp
+++ b/src/game/shared/tf/tf_weapon_minigun.cpp
@@ -1438,16 +1438,17 @@ void CTFMinigun::WeaponSoundUpdate()
 	float flPitch = 1.0f;
 
 	float flSpeed = ApplyFireDelay( 1.0f );
+
+	float flPitchScale = 1.0f;
+	CALL_ATTRIB_HOOK_FLOAT( flPitchScale, mult_minigun_pitch_scale );
+	if ( flPitchScale != 1.0f )
+	{
+		flSpeed /= ( flPitchScale );
+	}
+
 	if ( flSpeed != 1.0f )
 	{
 		flPitch = RemapValClamped( flSpeed, 1.5f, 0.5f, 80.f, 120.f );
-		// Tomislav - Fixes Tomislav's base firing sound pitch sounding lower than intended due to its base 20% slower firing speed penalty
-		if ( GetAttributeContainer()->GetItem()->GetItemDefIndex() == 424 )
-		{
-			//DevMsg( "Tomislav pitch correction: flSpeed BEFORE = %f, flPitch = %f\n", flSpeed, flPitch );
-			flPitch = RemapVal( flSpeed, 1.2f, 0.6f, 100.f, 120.f );
-			//DevMsg( "Tomislav pitch correction: flSpeed AFTER = %f, flPitch = %f\n", flSpeed, flPitch );
-		}
 	}
 
 	if ( m_bRageDraining )
--- a/src/game/shared/tf/tf_weapon_minigun.cpp
+++ b/src/game/shared/tf/tf_weapon_minigun.cpp
@@ -156,6 +156,9 @@ void CTFMinigun::WeaponReset( void )
 	m_iMinigunSoundCur = -1;
 	m_flMinigunSoundCurrentPitch = 1.0f;
 
+	m_flBasePostFireDelay = 1.0f;
+	CALL_ATTRIB_HOOK_FLOAT( m_flBasePostFireDelay, mult_postfiredelay );
+
 	StopMuzzleEffect();
 	StopBrassEffect();
 #endif
@@ -1440,7 +1443,16 @@ void CTFMinigun::WeaponSoundUpdate()
 	float flSpeed = ApplyFireDelay( 1.0f );
 	if ( flSpeed != 1.0f )
 	{
+		if ( m_flBasePostFireDelay != 1.0f )
+		{
+			// Pitch correction. Adjusts base firing sound pitch with the base firing speed.
+			// Fixes Tomislav's base firing sound pitch sounding lower than intended due to its base 20% slower firing speed penalty
+			// DevMsg( "Pitch correction: flSpeed BEFORE = %f, m_flBasePostFireDelay = %f, flPitch = %f\n", flSpeed, m_flBasePostFireDelay, flPitch );
+			flSpeed /= m_flBasePostFireDelay;
+		}
+
 		flPitch = RemapValClamped( flSpeed, 1.5f, 0.5f, 80.f, 120.f );
+		// DevMsg( "Pitch correction: flSpeed AFTER = %f, m_flBasePostFireDelay = %f, flPitch = %f\n", flSpeed, m_flBasePostFireDelay, flPitch );
 	}
 
 	if ( m_bRageDraining )
--- a/src/game/shared/tf/tf_weapon_minigun.h
+++ b/src/game/shared/tf/tf_weapon_minigun.h
@@ -167,7 +167,6 @@ private:
 	CSoundPatch		*m_pSoundCur;				// the weapon sound currently being played
 	int				m_iMinigunSoundCur;			// the enum value of the weapon sound currently being played
 	float			m_flMinigunSoundCurrentPitch;
-	float			m_flBasePostFireDelay;
 
 #ifdef GAME_DLL
 	float	m_flAegisCheckTime;
--- a/src/game/shared/tf/tf_weapon_minigun.h
+++ b/src/game/shared/tf/tf_weapon_minigun.h
@@ -167,6 +167,7 @@ private:
 	CSoundPatch		*m_pSoundCur;				// the weapon sound currently being played
 	int				m_iMinigunSoundCur;			// the enum value of the weapon sound currently being played
 	float			m_flMinigunSoundCurrentPitch;
+	float			m_flBasePostFireDelay;
 
 #ifdef GAME_DLL
 	float	m_flAegisCheckTime;


src/game/shared/tf/tf_weapon_minigun.cpp
```diff
@@ -1440,7 +1443,16 @@ void CTFMinigun::WeaponSoundUpdate()
 	float flSpeed = ApplyFireDelay( 1.0f );
 	if ( flSpeed != 1.0f )
 	{
+		if ( m_flBasePostFireDelay != 1.0f )
+		{
+			// Pitch correction. Adjusts base firing sound pitch with the base firing speed.
+			// Fixes Tomislav's base firing sound pitch sounding lower than intended due to its base 20% slower firing speed penalty
+			// DevMsg( "Pitch correction: flSpeed BEFORE = %f, m_flBasePostFireDelay = %f, flPitch = %f\n", flSpeed, m_flBasePostFireDelay, flPitch );
+			flSpeed /= m_flBasePostFireDelay;
+		}
```
FlaminSarge: I don't see how this isn't equivalent to just setting flPitch to 1.0 when no other fire delay effects are in play.
That is: I don't see how this works to fix the Tomislav without also making all other mult_postfiredelay no longer affect the pitch. Is that what you intend to do here, making mult_postfiredelay not affect the pitch at all (regardless of which direction it goes, up or down)?

It would also be smart to guard against m_flBasePostFireDelay=0, but in general something doesn't seem quite right about how this is implemented.

antielectronbomb: Thanks. 
Made a new implementation where flPitch is remapped via flSpeed. I am not sure if this is the best method to do this. I made sure that this will only apply to the Tomislav (weapon id is 424)
```
	// Get the pitch we should play at
	float flPitch = 1.0f;

	float flSpeed = ApplyFireDelay( 1.0f );
	if ( flSpeed != 1.0f )
	{
		flPitch = RemapValClamped( flSpeed, 1.5f, 0.5f, 80.f, 120.f );
		// Tomislav - Fixes Tomislav's base firing sound pitch sounding lower than intended due to its base 20% slower firing speed penalty
		if ( GetAttributeContainer()->GetItem()->GetItemDefIndex() == 424 )
		{
			//DevMsg( "Tomislav pitch correction: flSpeed BEFORE = %f, flPitch = %f\n", flSpeed, flPitch );
			flPitch = RemapVal( flSpeed, 1.2f, 0.6f, 100.f, 120.f );
			//DevMsg( "Tomislav pitch correction: flSpeed AFTER = %f, flPitch = %f\n", flSpeed, flPitch );
		}
	}
```

Tkain:
Instead of hardcoding the Tomislav's ID and fire rate difference, I think it would be more robust to just divide `flSpeed` by the value of the `mult_postfiredelay` attribute. That way `flSpeed` is entirely relative to the weapon's normal firing speed and nothing needs to be hardcoded. Something like this (feel free to copy):

```cpp
	// Get the pitch we should play at
	float flPitch = 1.0f;

	float flSpeed = ApplyFireDelay( 1.0f );

	float flDelayMult = 1.0f;
	CALL_ATTRIB_HOOK_FLOAT( flDelayMult, mult_postfiredelay );
	flSpeed /= flDelayMult;

	if ( flSpeed != 1.0f )
	{
		flPitch = RemapValClamped( flSpeed, 1.5f, 0.5f, 80.f, 120.f );
	}
```

antielectronbomb:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/2046#issuecomment-5555944350

This works, however in Mann vs. Machine the pitch does not scale with the Firing Speed Upgrade because both flSpeed and mult_postfiredelay gets changed in real time.

A better solution would be to make a separate hidden attribute solely for adjusting the base pitch of a Minigun, where mult_minigun_pitch_scale is set to 1.2 from items_game.txt.

```c++
	// Get the pitch we should play at
	float flPitch = 1.0f;

	float flSpeed = ApplyFireDelay( 1.0f );

	float flPitchScale = 1.0f;
	CALL_ATTRIB_HOOK_FLOAT( flPitchScale, mult_minigun_pitch_scale );
	if ( flPitchScale != 1.0f )
	{
		flSpeed /= ( flPitchScale );
	}

	if ( flSpeed != 1.0f )
	{
		flPitch = RemapValClamped( flSpeed, 1.5f, 0.5f, 80.f, 120.f );
	}
```

However this will require work from Valve's side, since the new attribute will need to be added by Valve. So in items_game.txt (attrib 4000 is an example number), we add this new attribute (4000 is just an example attribute number):

```
	"4000"
	{
		"name"	"minigun pitch scale"
		"attribute_class"	"mult_minigun_pitch_scale"
		"description_string"	"unused"
		"description_format"	"value_is_percentage"
		"hidden"	"1"
		"effect_type"	"neutral"
		"stored_as_integer"	"0"
	}
```

And then we add the new attribute to the Tomislav's attributes:
```
		"weapon_tomislav"
		{
			...
			"attributes"
			{
				"minigun spinup time decreased"
				{
					"attribute_class"	"mult_minigun_spinup_time"
					"value" "0.8"
				}
				"minigun no spin sounds"
				{
					"attribute_class"	"minigun_no_spin_sounds"
					"value" "1"
				}
				"fire rate penalty"
				{
					"attribute_class"	"mult_postfiredelay"
					"value"	"1.2"
				}
				"weapon spread bonus"
				{
					"attribute_class"	"mult_spread_scale"
					"value"	"0.8"
				}
				"minigun pitch scale"
				{
					"attribute_class"	"mult_minigun_pitch_scale"
					"value" "1.2"
				}
			}
		}
```

By doing it this way, this makes it so that only the Tomislav is affected (since currently it is the only Minigun whose sounds are incorrectly pitched) and also makes it so that if ever the Tomislav's stats gets updated, it would be a simple matter of just updating the attribute values and works nicely with game server modifications.

If anyone wants to test this implementation, replace `mult_minigun_pitch_scale` with `mult_spread_scale` and then multiply `flPitchScale` by 1.5x such that it looks like the following:
```c++
	float flPitchScale = 1.0f;
	CALL_ATTRIB_HOOK_FLOAT( flPitchScale, mult_spread_scale);
	if ( flPitchScale != 1.0f )
	{
		flSpeed /= ( flPitchScale * 1.5f );
	}
```

Tkain:
> This works, however in Mann vs. Machine the pitch does not scale with the Firing Speed Upgrade because both flSpeed and mult_postfiredelay gets changed in real time.

Good catch, I didn't know MvM changed attributes in real time. I wonder if there's a way to get the weapon's base attributes separately? In any case, this is about as non-hardcoded as you could get otherwise, so good work.

antielectronbomb:
> Good catch, I didn't know MvM changed attributes in real time. I wonder if there's a way to get the weapon's base attributes separately? In any case, this is about as non-hardcoded as you could get otherwise, so good work.

Yep. Having a separate attribute for adjusting Minigun pitch is probably the most non-hardcoded way to fix the Tomislav's sound pitch while also working nicely with Mann vs. Machine. Valve will need to add a separate attribute from their end since the SDK doesn't have its own copy of items_game.txt.

When I first thought of trying to fix the Tomislav's weapon sound pitch, I had the following methods in mind:
1. Manually change the pitch of the Tomislav's sounds via game_sounds_weapons.txt
2. Manually change the pitch via tf_weapon_minigun.cpp
3. Use a new separate attribute for adjusting Minigun pitch

Both methods 1 and 2 are essentially hardcoded fixes. 2. only requires changes to just the Source SDK code, while 1 and 3 requires changes to the Source SDK code and needs Valve's input.

**_This should be the last alteration to this PR._**

FlaminSarge:
If you give me a few days, I'm going to doublecheck what all could be affected if you just use the original implementation (making mult_postfiredelay not affect the pitch).
It's possible that it'll mess with community mods not sounding right for the minigun (or for other weapons) that use that attribute, and if that's the case, then it may just be better to fix the Tomislav's sound itself rather than making the changes here.

JoriKos:
Is this really a bug? It looks completely fine to me, and the Wiki noting it as an undocumented change doesn't automatically mean it was unintended just that it was a change Valve didn't mention. There have been plenty of fully-intended changes that Valve hasn't mentioned, like the recent L'Étranger rename and the subsequent rename of the Tzar Athlete to the Tsar Athlete. Two intentional changes which are not mentioned.

antielectronbomb:
> Is this really a bug? It looks completely fine to me, and the Wiki noting it as an undocumented change doesn't automatically mean it was unintended just that it was a change Valve didn't mention. There have been plenty of fully-intended changes that Valve hasn't mentioned, like the recent L'Étranger rename and the subsequent rename of the Tzar Athlete to the Tsar Athlete. Two intentional changes which are not mentioned.

It is, the original .wav sound files all have a higher pitch compared to the ones in-game. The firing sounds does not really match up with the ammo drain when you fire too because of this. Likely a side effect of a patch involving the MvM firing speed upgrade back in Two Cities or Love and War.

See https://github.com/ValveSoftware/Source-1-Games/issues/4734

FlaminSarge:
I think the fixed pitch sound file linked in the above ticket is a more prudent fix than fixing it in code.
You would want mult_postfiredelay to affect firing sounds normally for most guns, but the Tomislav's firing sounds were built with the lower fire rate in mind initially and were not updated to be a 'normal' fire rate sound (pitched up to compensate) when mult_postfiredelay was updated to affect firing sounds. It stands to reason that the sound file itself should be updated and the firing sound behavior in code maintained.

JoriKos:
> It is, the original .wav sound files all have a higher pitch compared to the ones in-game. The firing sounds does not really match up with the ammo drain when you fire too because of this. Likely a side effect of a patch involving the MvM firing speed upgrade back in Two Cities or Love and War.
> 
> See [ValveSoftware/Source-1-Games#4734](https://github.com/ValveSoftware/Source-1-Games/issues/4734)

Yeah this is just repeating what was already said, but it still doesn't really sound to me like it's a bug. The firing rate looks fine to me.

antielectronbomb:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/2046#issuecomment-5571525212

Could be another way, but I think changing the sound pitch via game_sounds_weapons.txt might be a simpler method. Adding a pitch adjustment key of 1.087 for every Tomislav sound scripts would do the trick.

FlaminSarge:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/2046#issuecomment-5577051640

That would work too. Depends on whether there are plans to reuse that sound script for anything (or that sound file, where they would want the fixed file rather than just the fixed script).

antielectronbomb:
> > It is, the original .wav sound files all have a higher pitch compared to the ones in-game. The firing sounds does not really match up with the ammo drain when you fire too because of this. Likely a side effect of a patch involving the MvM firing speed upgrade back in Two Cities or Love and War.
> > 
> > See [ValveSoftware/Source-1-Games#4734](https://github.com/ValveSoftware/Source-1-Games/issues/4734)
> 
> Yeah this is just repeating what was already said, but it still doesn't really sound to me like it's a bug. The firing rate looks fine to me.

Try removing the -20% firing speed attribute and listen to the firing sounds with the SDK or SourceMod plugin so that it's more obvious. The firing speed and pitch will not match. The pitch will change because of it. 


Besides, the firing speed penalty is still the same before and after the patch. If the pitch change was intentional, then how come the firing speed penalty was not adjusted?


If it was an intentional change, then why are there no patch notes about it? This was back when Valve was still more directly involved with the game's development.

JoriKos:
> Try removing the -20% firing speed attribute and listen to the firing sounds with the SDK or SourceMod plugin so that it's more obvious. The firing speed and pitch will not match. The pitch will change because of it.

I know the pitch changes, I'm saying the evidence for this being a bug is not very solid.

> Besides, the firing speed penalty is still the same before and after the patch. If the pitch change was intentional, then how come the firing speed penalty was not adjusted?

Why would the firing speed penalty be adjusted? That's a balance change, why change the balance just so a sound doesn't change from one update to another?

> If it was an intentional change, then why are there no patch notes about it? This was back when Valve was still more directly involved with the game's development.

This is very common though, there are loads of changes that just aren't mentioned and completely intentional. There are quite major changes that have gotten no patch notes (crafting recipe changes, viewmodel changes, new qualities being added, attributes being reworded). Hell, the Natascha's 50% extra ammo stat being removed wasn't even mentioned in any patch note. If an actual balance change isn't mentioned, why would a slight change in pitch be?

mastercoms:
@JoriKos When Valve authored the sound originally, they authored it with the Tomislav's fire rate (with penalty) already in mind. Then, with the MvM updates, the fire sound was updated to take into account fire rate attributes and modulate pitch based on that. This changes the intended pitch/speed of the fire sound that Valve had already authored for the weapon. I think what you're misunderstanding is that the Tomislav always had a 20% fire rate penalty, and the weapon sound was authored for that.

However, I do agree that this is probably not the best fix. It assumes that all sounds are authored for their default fire rate, which may not be true. For example, if the Tomislav's fire rate does change beyond 20%, then the sound would be incorrect again. It would be much better to have an attribute marking up the fire sound's base fire rate, and modulating based off of that.

JoriKos:
> When Valve authored the sound originally, [...]

No, I do understand what's happening, but to me the sound doesn't look desynced at all. The original sound actually sounds too fast to me.

mastercoms:
I mean, it is also potentially possible that the sound artist authored the sounds in a staging environment where this MvM scaling code was active already. But I also do think the demonstration videos of the fix sound a bit different from the original 15 year old video that demonstrates the Tomislav. Maybe it's because of the room DSP or something that it sounds different, but the sound in general sounds a lot sharper / grating to the ears for some reason.

antielectronbomb:
> Replying to https://github.com/ValveSoftware/source-sdk-2013/pull/2046#issuecomment-5619129344

Regardless I think this PR is best described as a reversion rather than a bug fix at this point.

PR #2047, opened 2026-09-02
[TF2] Fix Pomson 6000 draining mechanic only to work when the medic has charge.

ALIEN31ITA:
# Description
This fix is meant to fix the drain  of the ubercharge, only if the medic has atleast some charge in his medigun.
Both Visually and Technically.

https://github.com/user-attachments/assets/162e15b3-6fbe-47ed-bf61-c761bf9872aa


ALIEN31ITA:
Aside the rare occurrence, of finding both a pomson engineer and a medic with no charge at the same time, it still does make sense to fix, giving a clear feedback that your not draining said medic.

src/game/shared/tf/tf_weaponbase.cpp
mastercoms, line 5579: This can scale the drain down to 0 as well. Please add a check for that too. I recommend adding a new float variable flDrain for this, so that there is no integer and float mixing either.

mastercoms, line 5589: Also should check for spy here (and below, check if the falloff made it 0, avoid integer and float mixing)

PR #2054, opened 2026-09-04
[HL2MP] Modernize the chat looks

speedvoltage:
The chat in HL2MP has long been an ancient system. It uses a fixed legacy layout that does not scale cleanly, while its fade and rate-limit timers pause with simulation time. When the game is paused with `sv_pausable 1`, the chat depends on `gpGlobals->curtime`, which is, to me, wrong. It should be using `gpGlobals->realtime`, otherwise, you are stuck with a partial rendered UI, the inability to send another message beyond one due to the 0.66s second talk time delay and anti chat spam time. Also, everything is yellow and ugly.

So, let's switch it out for a sleek, modern panel that’s responsive and includes optional timestamps, all while keeping a subtle color hierarchy. Apply some team names and assign them distinct colors. Server messages remain neutral, and use the accent color for server and ConVar messages. For fading and rate limiting, move it to using `gpGlobals->realtime`. Finally, it is now possible to drag and resize the chat. This helps player move the chat where they want now rather than having to edit resource files to do just that.

End results:

https://github.com/user-attachments/assets/bded0311-441b-4f1a-8239-5b6e690c9697

image

### DIFF
--- a/src/game/client/clientmode_shared.cpp
+++ b/src/game/client/clientmode_shared.cpp
@@ -37,6 +37,9 @@
 #include "hud_vote.h"
 #include "ienginevgui.h"
 #include "sourcevr/isourcevirtualreality.h"
+#ifdef HL2MP
+#include "hl2mp_gamerules.h"
+#endif
 #if defined( _X360 )
 #include "xbox/xbox_console.h"
 #endif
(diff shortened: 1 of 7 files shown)


blackletum:
Ok, but what's the song name?

And maybe have this as an option, but not default? Even if it's pretty neat. 

Does the chat box retain its position when you move it, even through map changes/etc?

speedvoltage:
It is a random tropical house mix https://www.youtube.com/watch?v=CkD08yt1B3Y

Eventually could have a VGUI option tied to a cvar, maybe a reset size and position button.

Yes, even game restarts, it basically updates:
```cpp
ConVar cl_chat_window_x( "cl_chat_window_x", "-1", FCVAR_CLIENTDLL | FCVAR_ARCHIVE, "Normalized horizontal position of the chat window." );
ConVar cl_chat_window_y( "cl_chat_window_y", "-1", FCVAR_CLIENTDLL | FCVAR_ARCHIVE, "Normalized vertical position of the chat window." );
ConVar cl_chat_window_w( "cl_chat_window_w", "-1", FCVAR_CLIENTDLL | FCVAR_ARCHIVE, "Normalized width of the chat window." );
ConVar cl_chat_window_h( "cl_chat_window_h", "-1", FCVAR_CLIENTDLL | FCVAR_ARCHIVE, "Normalized height of the chat window." );
```

PR #2055, opened 2026-09-04
[HL2MP] Modernize the scoreboard

speedvoltage:
The scoreboard was updated ever so slightly for the anniversary update, but I think so much more could have been done to make it better. I tried to modernize the scoreboard a bit by having it responsive, adjusting its size dynamically based on the number of players, and added some additional server information containing current map, mode, player-count, frag-limit, and time information. The hostname also updates immediately when it is changed by the server op. Spectators are now shown as a dim entry.

### DIFF
--- a/src/game/client/hl2mp/ui/hl2mpclientscoreboard.cpp
+++ b/src/game/client/hl2mp/ui/hl2mpclientscoreboard.cpp
@@ -6,34 +6,31 @@
 //=============================================================================//
 
 #include "cbase.h"
-#include "hud.h"
 #include "hl2mpclientscoreboard.h"
 #include "c_team.h"
 #include "c_playerresource.h"
-#include "c_hl2mp_player.h"
 #include "hl2mp_gamerules.h"
 
 #include <KeyValues.h>
+#include <tier1/strtools.h>
 
 #include <vgui/IScheme.h>
 #include <vgui/ILocalize.h>
 #include <vgui/ISurface.h>
-#include <vgui/IVGui.h>
+#include <vgui_controls/Label.h>
+#include <vgui_controls/ScrollBar.h>
 #include <vgui_controls/SectionedListPanel.h>
 
-#include "voice_status.h"
-
 using namespace vgui;
-
-#define TEAM_MAXCOUNT			5
 
 // id's of sections used in the scoreboard
 enum EScoreboardSections
 {
-	SCORESECTION_COMBINE = 1,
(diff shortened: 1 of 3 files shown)


inco-cc:
Looks clean, and the additional info is much appreciated.

However, I do think it clashes with the rest of the HUD, especially in the first screenshot. The current scoreboard, while flawed, is at least consistent with the standard HL2 HUD elements, with the rounded borders, more transparent background etc.

Honestly, I think if you brought the changes like dynamic sizing, the additional info and some of the text formatting to the current scoreboard, that would be sufficient.

speedvoltage:
> Looks clean, and the additional info is much appreciated.
> 
> However, I do think it clashes with the rest of the HUD, especially in the first screenshot. The current scoreboard, while flawed, is at least consistent with the standard HL2 HUD elements, with the rounded borders, more transparent background etc.
> 
> Honestly, I think if you brought the changes like dynamic sizing, the additional info and some of the text formatting to the current scoreboard, that would be sufficient.

Tell you what. I pushed some changes that keeps the classic style while the UX is improved. I have seen the same scoreboard for over two decades, so that is why I pushed the original idea of changing it a bit for something more modern, but your feedback is totally valid here. I added `cl_scoreboard_style` here, defaulting to 0 for the classic style. I think this can be a good middle ground between keeping the classic style for cohesive style, and modern style for those who want a fresh look. I am totally open to additional feedback if this seems all unnecessary. I pushed similar changes to #2054 and #2056.

Updated the screenshots here. Will follow up on the other two PRs with fresh screenshots a little later.

inco-cc:
Looks good! The opt-in ConVar is a great solution.

It might be best to leave the modern style for a future PR, though. Valve tends to merge “safer” changes, so they might be more likely to merge this PR if it was simply an upgraded version of the existing scoreboard, without introducing too many stylistic changes, even if they are optional. Otherwise, we'll have to hope that Valve also likes the modern style enough to include it. Just a thought.

While I do like the modern style, I think having the other changes merged first should be higher priority, especially things like the spectator menu fix and movable/resizable chat from the other PRs.

speedvoltage:
You are probably right. I updated the PR so that only the classic style remains, and a future PR may be opened to add an additional modern style. This way, we keep the classic style with UX improvements. I updated the spec GUI and chat GUI PRs as well to retain classic style while fixing bugs and improving UX. This should make all 3 PRs more cohesive this time. Thanks for the feedback!

PR #2058, opened 2026-09-06
[HL2MP] Fix & predict missing grenade roll sound

speedvoltage:
Issue:

Over a year ago, I opened a PR that aimed at fixing the missing roll sound for the frag grenade (see https://github.com/ValveSoftware/source-sdk-2013/pull/1101). The original fix was to use `CDisablePredictionFiltering disablePred;` so that the server played the sound for the client. This is still a valid fix for dedicated servers looking at fixing this with only server-side changes or plugins. The problem is that the client never hears that sound because the roll sound is triggered only by a server animation event, but prediction filtering excludes the owning player. Since the client never handles that event, the owner hears nothing.

Fix:

The fix here is to handle pullback completion and the roll-release event during client prediction, using the model’s existing animation timing. The local client hears the sound locally, while the server plays it for nearby players without duplicating it for the owner.

### DIFF
--- a/src/game/shared/hl2mp/weapon_frag.cpp
+++ b/src/game/shared/hl2mp/weapon_frag.cpp
@@ -546,6 +546,9 @@ void CWeaponFrag::RollGrenade( CBasePlayer *pPlayer )
 
 #endif
 
+#ifndef CLIENT_DLL
+	CDisablePredictionFiltering disablePred;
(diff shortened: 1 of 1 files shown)


PiMoNFeeD:
this is some of the worst code I've seen on this repository, is this AI generated?

PR #2059, opened 2026-09-07
[CSS, DODS, HLDMS] Shut up `info_ladder` console spam

SirYodaJedi:
## Issue
VBSP creates `info_ladder` entities for `func_ladder` after moving the brushes to worldspawn. While this was intended to help navmesh generation, it never actually did this, so the associated CInfoLadder class was removed sometime after the Orange Box SDK was released. This removal causes maps that used func_ladder to spam the console with "Attempted to create unknown entity type info_ladder!", though, as the entity no longer exists. This occurs for many vanilla and custom maps in CS:S, DoD:S, and HLDM:S.

## Solution in this PR
By treating `info_ladder` as an alias for `info_null`, the console spam is shut up. This can be combined with PR #910 to avoid creating an extra edict for a brief period at map spawn.

I have committed same fix in [Fortress Forever 2013](https://github.com/fortressforever-2013/fortressforever-src/pull/83), and no ill effects have been noticed. 

## Reviewed alternatives
It is possible to edit VBSP to not generate these info_ladder entities, however this requires recompiling maps to take any effect. It is possible to create ladders CS:GO-style, without func_ladder, however this requires recompiling maps to take any effect.
It is possible to re-add the CInfoLadder entity class from the Orange Box SDK, however this seems unnecessary, as it never did anything.

## Notes
Some third-party bot plugins try to read info_ladder to determine where ladders are. This is bad practice, as ladders can be created without func_ladder by using a `%compileladder` material such as toolsinvisibleladder, a prop that was compiled with `$contents ladder`, or a material with `$surfaceprop` `Ladder` or `WoodLadder`. Such plugins should be modified to look for CONTENTS_LADDER and/or the aforementioned surfaceprops.

## Legalese
I have read the contribution rules, and can affirm that the contents of this PR are based entirely upon existing SDK code and does not contain any code from third parties, other developers (the similar commit to Fortress Forever was my own), or LLM code generators.


blackletum:
This doesn't compile as-is, FYI. But wouldn't it be better to simply add

`LINK_ENTITY_TO_CLASS( info_ladder, CNullEntity );`

in subs.cpp?

SirYodaJedi:
> This doesn't compile as-is, FYI. But wouldn't it be better to simply add
> 
> `LINK_ENTITY_TO_CLASS( info_ladder, CNullEntity );`
> 
> in subs.cpp?

You're probably right.

PR #2061, opened 2026-09-07
[ALL] Fix `TestActivator` crash when caller and/or activator is null

SirYodaJedi:
Per the VDC:
> Calling this input will cause the server to crash if the [!activator](https://developer.valvesoftware.com/wiki/!activator) entity no longer exists. Do not use this input to test any entities which may be asynchronously deleted, such as players or projectiles, or in an i/o chain which might be initiated by entity deletion, such as the OnEndTouch output of a [trigger](https://developer.valvesoftware.com/wiki/Triggers). This crash may be more likely to occur if a delay is set on the calling output.

This PR fixes the aforementioned crashes by aborting early if the !activator and/or !caller is NULL (and therefore, doesn't exist).

Based upon a solution shared by @ALIEN31ITA on the TF2Maps discord, from [their TF2 mod](https://github.com/ALIEN31ITA/Better-Fortress-2); as such, I have set them as a co-author.


SirYodaJedi:
Caller generally shouldn't be deleted before this runs, but there are some entities that improperly set the caller as null when firing outputs (env_wind, game_round_end, logic_eventlistener), just like there are way too many entities that improperly set activator as null.

PR #2064, opened 2026-09-08
[ALL] Fix env_sun disappearing at certain view angles

### LINKED ISSUE #7405: (Feature Request) Backport env_sun disappearing fix
In game branches post-L4D2, `env_sun` entities will disappear when looking directly at them even if there isn't a crosshair overlay. This was fixed in L4D2 (and as a result subsequent games after it).

The fix is very simple, and involves editing [src/game/client/glow_overlay.cpp](https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/game/client/glow_overlay.cpp) to change the line referencing 0.999f to 0.99f.

```Vector pos = CurrentViewOrigin() + m_vDirection * zFar * 0.999f;```

is changed to

```Vector pos = CurrentViewOrigin() + m_vDirection * zFar * 0.99f;```

This is a relatively simple fix and would be ideal to backport to games like Half Life 2 and Team Fortress 2 considering they are receiving more update attention lately.

Thanks!

chocochocotorta:
The PR changes the final multiplier of the pos value in glow_overlay.cpp from 0,999f to 0.99f.
This fixes the env_sun entity dissapearing when looking at specific angles.

Tested both on TF2 and HL2:MP on the same map looking at the same position and angles.
### **Before:**
image

### **After:**
image

Closes https://github.com/ValveSoftware/Source-1-Games/issues/7405.


chocochocotorta:
The Fix was proposed at [Valve Developer Community](https://developer.valvesoftware.com/wiki/General_SDK_Snippets_%26_Fixes#Fix_env_sun_sprite_disappearing_in_sky_when_you_look_at_it) and was already opened in the past by #1785, but the person deleted it.

PR #2068, opened 2026-09-08
[TF2] Updated resistance pierce attribute to pierce powerups and Sapper resistance

### LINKED ISSUE #4412: [TF2] Please fix the Enforcer's "piercing" issues.
Apparently the Enforcer has a few bugs relating to it's "piercing" feature.
Doesn't pierce the Sentry's "33% damage resistance from the Spy", doesn't pierce the immunity to critical hits with the Battalion's Backup affect, doesn't pierce through certain Mannpower powerups, and it doesn't work with the Pain Train's increased bullet vulnerability. Maybe it'd be a slightly better weapon to use if these'd get fixed...

Nah, but they should be fixed regardless.

JoriKos:
Pretty much as the title describes, the Enforcer's resistance piercing attribute does not work on sapper Sentry Guns, nor does it work on any of the Mannpower powerups with a resistance (Plague victims vs Plague powerup holders, Resistance powerup, and the Vampire powerup). This pull request fixes that, making it deal full damage in all those situations.

There is one other resistance it does not pierce, which is the Crit immunity of the Battalion's Backup. However, seeing as the check for piercing resistances is specifically after this, it feels intended to me.

Fixes https://github.com/ValveSoftware/Source-1-Games/issues/4412 and https://github.com/ValveSoftware/Source-1-Games/issues/8262 (latter is closed, but I made a comment on there related to making a pull request)


num-get:
I don't think sapped sentry resistance not being pierced is a bug at all. More of a balancing act the same way level 2/3 sentry gun has innate resistance against Minigun.

JoriKos:
The difference is that Miniguns deal an insane amount of damage, while a Sapper deals very little damage and the Spy is incredibly weak. It's also only applicable to Sentry Guns as well, which means shooting a sapped Dispenser and Teleporter (arguably more important buildings) gives you the full damage regardless and means you can easily see the resistance being applied to Sentry Guns without visiting the Wiki article for it. Thus, you can reasonably expect the Enforcer to be piercing this resistance, but because it does not it's inconsistent with the attribute and player experience.

All that being said, it is possible that it is intentionally excluded, but I don't see enough evidence that it is excluded (unlike the Crit resistance for the Battalion's Backup). If it is intentionally excluded, then consider this a request at a balance change instead.

num-get:
My reasoning is that the resistance is only applicable to that specific spy that placed the sapper, other spy from the same team still does full damage just fine.

PR #994, opened 2025-03-12
[TF2] Change max FOV from 90 to 130

blackletum:
This PR is made to increase the max field of view within Team Fortress 2 from 90 to 130. 

This would help players on standard widescreen monitors be able to see as much as players on ultrawide displays, help reduce nausea, to have parity with Valve's other titles such as Left 4 Dead 2 as well as the Team Fortress modding communities typically allowed max FOV settings.

PR #2069, opened 2026-09-09
[ALL] Add 70–110 FOV options

speedvoltage:
At the risk of dividing the community on this one, my reasoning of pushing this is because, in HL2DM at least, we have been using a server-side FOV plugin or custom server binaries implementing FOV options for the past decade. Most have been using 110 as they feel more comfortable with this setting with their screen resolution.

TF2 already seems to allow 75 to 90, but HL2DM just flat out does not have any FOV options and everyone is stuck to 90. I have enabled support for FOV with `fov_desired` and added the updated values in the options as well. I also made it so that spectators and SourceTV receive proper view model FOV when in IN_EYE mode.

This PR also takes https://github.com/ValveSoftware/source-sdk-2013/pull/994 and https://github.com/ValveSoftware/source-sdk-2013/pull/1912 into consideration.

The following screenshots in TF2 were taken at 5120 x 1440. The HL2DM ones were at a standard 1920 x 1080.

TF2:
`fov_desired 70`

`fov_desired 90`

`fov_desired 110`

HL2DM:
`fov_desired 70`

`fov_desired 90`

`fov_desired 110`


JoriKos:
You could probably optionally lock these behind convars, off by default, if there really is going to be a divide in a community.

speedvoltage:
> You could probably optionally lock these behind convars, off by default, if there really is going to be a divide in a community.

To be determined. This division discussion was a while back. I would not want to add cvars if there is no reason to. We will see what other folks think and adapt if need be.

Niterux:
High fov's break Source's default 2d skybox rendering impl due to it not using the camera's real fov for the frustum culling of the 2d skybox model. [TF2 Classified had to implement their own renderer](https://store.steampowered.com/news/app/3545060/view/704392974023263097?snr=).

speedvoltage:
I agree this is a problem. I was hesitant on opening this PR but did so anyway, because it can easily be fixed if Valve does something about it. I included the skybox fix in a separate branch, but that too, I am not too sure this should be a PR if this can be fixed in the engine.

Issue #447, opened 2018-03-06
Predicted sequences problem.


Mince228:
So I've been trying to make the hl2mp player animations nicely in sync with both client and the server by predicting them. After 'implementing' the SetAnimation code from CHL2MP_Player without errors into the client side of the class the system fires up but the client always has its m_nSequence set to 0 and I cant seem to find the reason for why its happening. The server seems to have normal sequence numbers but the client always outputs zeros when tested. The player model appears to be in the T-Pose most of the time but seems like its trying to correct itself towards some kind of animation, perhaps from the server value.

If anyone has any experience with this or perhaps some helpful info it would be appreciated.

azuisleet:
The HL2MP implementation should actually be based on CMultiPlayerAnimState instead. The SetAnimation function is a red herring, as CMultiPlayerAnimState will reset the sequence anyway. 

There's a quick overview on the wiki: https://developer.valvesoftware.com/wiki/M_PlayerAnimState

You can also diagnose issues with anim_showstate 1

Mince228:
Well I tried "implementing" CMultiPlayerAnimState but its giving me some errors: [https://imgur.com/a/z3iF2](url)
CBasePlayer for example seems to be declared inside this header file but it doesnt seem to want to recognise it along with some others. If you tried implementing this animstate did you run into the same problems?

Mince228:
Got it fixed based on code from Tony Sergi's hub page. 

Closing

Issue #119, opened 2013-07-22
Jittery animations.


NFMynster:
It seems like the thirdperson animations are bugged out.
Anyone know how to fix this?
Video here:
http://www.youtube.com/watch?v=5VXkS9FGRMA&feature=youtu.be

Issue #35, opened 2013-06-28
Choppy player animations in MP


TurboBroadway:
It seems that Source 2013 MP is no longer using HL2MPPlayerAnimState. Without it, player animations are choppy and left unpredicted. 

While this is an easy issue to fix, I still think it should be reimplemented again.

PR #2071, opened 2026-09-10
[HL2MP] Migrate player animations to CMultiPlayerAnimState

inco-cc:
This converts the old `CPlayerAnimState` in HL2:DM to the newer `CMultiPlayerAnimState` used in later games like TF2.

Currently, player animations are calculated on the server, which is then sent to clients for interpolation. 

`CMultiPlayerAnimState` makes it possible to also calculate player animations on the client and have it match the server. This resolves the issue where animations appear choppy and gestures do not play properly in thirdperson, since animations are not interpolated for the local player.

This is based on Tony Sergi's unfinished implementation from the 2007 SDK, adapted for the current SDK. It appears at some point HL2:DM was going to transition to 9-way blending for player animations. If HL2:DM ever receives newer animations, this implementation would make it a lot easier.

I didn't implement the automatic head looking behavior because I don't think it's necessary. The aim pose parameters are sufficient, and they more accurately represent where the player is actually looking.

Related issues: #35, #119, #447

Before:

https://github.com/user-attachments/assets/50ff4b6b-0ead-4aab-aaf4-9d620060f3f4

After:

https://github.com/user-attachments/assets/37642a1a-87ef-4ba9-b719-265754789c7a


speedvoltage:
Great work on that one! Looks clean! Glad to see HL2DM is still getting some love!

PR #2074, opened 2026-09-12
Reworked default settings (mostly for new players)

pr0jesse:
# Description
https://steamcommunity.com/sharedfiles/filedetails/?id=3800210535

Updates several default TF2 client settings.

Changes include:
- FOV default from 75 to 90
- Viewmodel FOV from 54 to 70
- Fast weapon switch enabled by default
- Hit and kill sounds enabled by default
- Automatic Medic caller indicators enabled by default
- Medic heal target marker enabled by default

Update 1:
- Full charged shot indicator enabled by default
- Added a enable viewmodel toggle
- Added a slider for viewmodel transparency

The existing `fov_desired` and `hud_combattext_healing` settings are now also exposed through Advanced Options.

The corresponding options remain configurable through the existing settings and ConVars.


JoriKos:
I think minmode should be default too. A higher viewmodel fov is always recommended, but it blocks more of the screen.

pr0jesse:
(Im new to github, PRs, and coding as a whole, sorry for all these tweaks I'm still learning)

Pygargue00fr:
This default setting changes NEEDS to be added in.

KimmyTF2:
Personally I would leave `tf_medigun_autoheal` off by default. Healing feels more impactful when you actively press and hold M1 instead of clicking between targets. Just an opinion of someone playing a lot as the Medic.

pr0jesse:
> Personally I would leave `tf_medigun_autoheal` off by default. Healing feels more impactful when you actively press and hold M2 instead of clicking between targets. Just an opinion of someone playing a lot as the Medic.

Yeah I think I'm going to agree with you guys on that, it was more my personal preference but after researching a bit on steam forums and reddit I'll toggle it back off.

JoriKos:
I prefer the medigun autoheal because holding M1 is just a strain, and switching targets is faster with toggle because you just have to click once instead of letting go then clicking again (making it tactically superior).

src/game/client/tf/clientmode_tf.cpp
```diff
@@ -111,7 +111,7 @@ static Color colorEyeballBossText( 134, 80, 172, 255 );
 static Color colorMerasmusText( 112, 176, 74, 255 );
 
 ConVar default_fov( "default_fov", "75", FCVAR_CHEAT );
-ConVar fov_desired( "fov_desired", "75", FCVAR_ARCHIVE | FCVAR_USERINFO, "Sets the base field-of-view.", true, 20.0, true, MAX_FOV );
+ConVar fov_desired( "fov_desired", "90", FCVAR_ARCHIVE | FCVAR_USERINFO, "Sets the base field-of-view.", true, 20.0, true, MAX_FOV );
```
blackletum: You'll want to change default_fov to 90 here, not fov_desired, if you're hoping for 90 to be the default field of view

src/game/client/c_baseviewmodel.cpp
```diff
@@ -42,6 +42,7 @@
 
 #ifdef TF_CLIENT_DLL
 	ConVar cl_flipviewmodels( "cl_flipviewmodels", "0", FCVAR_USERINFO | FCVAR_ARCHIVE | FCVAR_NOT_CONNECTED, "Flip view models." );
+	ConVar cl_viewmodel_opacity( "cl_viewmodel_opacity", "100", FCVAR_ARCHIVE, "Viewmodel opacity percentage.", true, 0.0f, true, 100.0f );
```
blackletum: This entire section should be its own PR, since this is adding entirely new functionality and not in line with the goal of simply changing default settings.

src/game/client/weapon_selection.cpp
```diff
@@ -21,7 +21,7 @@
 #define HISTORY_DRAW_TIME	"5"
 
 ConVar hud_drawhistory_time( "hud_drawhistory_time", HISTORY_DRAW_TIME, 0 );
-ConVar hud_fastswitch( "hud_fastswitch", "0", FCVAR_ARCHIVE | FCVAR_ARCHIVE_XBOX );
+ConVar hud_fastswitch( "hud_fastswitch", "1", FCVAR_ARCHIVE | FCVAR_ARCHIVE_XBOX );
```
blackletum: This should be put into a TF ifdef so that it doesn't affect other source games

ShowinGrowin:
Many things here are undeniable improvements, but I think the viewmodel fov is a bad inclusion.

The original viewmodel fov is the intended visual look for the view models. The higher you go the more buggy and broken the visuals will get. Even at 70 there are noticeable issues. I don't see any reason to change that from the default setting.

Even ignoring the issues like arms being cut off, (which does still happen for some animations at 70) the view models are still artistically designed around the default value. Some look very odd at higher values, such as the minigun. Some things aren't meant to be visible. I think it's important to retain the original artistic intent. It's ok as an option for people to experiment with, though by default people should have the expectation that issues like this won't be present, and they will see the visuals that the developer's actually intended.

JoriKos:
70 is the max value limited by the game, and higher viewmodel fov is generally recommended by the entire playerbase. I have used 70 FOV for a few years now, and I haven't noticed a single artifact or cut-off arm with 70 viewmodel fov.

Pygargue00fr:
The only issue with 70 weapon fov is the sandman and wrapped assassin animations. 

But but this is an animation bug that should be fixed and has nothing to do with your fov whatsoever.

ShowinGrowin:
clip1
clip2

While it is far more minimal than higher fovs, it does exist. Just because it's not as noticeable, doesn't necessarily mean it's not a problem, that would ideally need to be fixed. This is probably the furthest valve even allows simply because it has the least problems.

minigun1
minigun2

But like I said, its not just the fact that there's issues like that. The animations are literally designed for that fov, which is an important thing to consider. Some things are intentionally hidden from your view. And you're kinda messing with the intention by changing this setting. Even a little bit. And I wouldn't say the "art" aspect to it is unimportant.

I'll acknowledge that many do play around with this setting. But that's exactly why its fine as an advanced setting. You can say "people prefer this" but you can't really quantify that. That's exactly the problem with determining what the default settings should be in the first place. Because it's not that simple. A lot of people customize the game to extremes and make it look completely different. While others change barely anything and stick to the vanilla look.

Imo it would be best to retain the original vanilla experience as close as possible, and when it comes to customization, improve that to be a lot better. I think we can all agree that the options UI is very bad in its current state. I'm not saying there shouldn't be any changes to the default settings, because some do have a big impact. But other things like this are more in the realm of preference, with some problems.

JoriKos:
At its furthest, 70 FOV on the Medi Gun looks like this:
afbeelding

So your FOV must've been set higher.

The Heavy fists one is only visible for a few frames when using the left hand punches. The Minigun one I also don't see what the issue is, that's just part of the Minigun that's visible (the drum is rendered at all times, as seen on FOV 150, if that's what you were referring to):
afbeelding

> You can say "people prefer this" but you can't really quantify that.

If you look up any recommended general settings, it will always talk about a high weapon fov. 90 FOV is often recommended, but /that/ is the value that actually causes a ton of these artifacts (like the Original viewmodel's extra hand). That's why it's changed to 70 here.

> The only issue with 70 weapon fov is the sandman and wrapped assassin animations.

I'm not sure what specifically is meant by this, but both animations look fine to me. I can't find an issue.

70 viewmodel FOV is fine, and it's intended by the game to be used, otherwise it would limit the fov to something smaller. 
The only real issue listed here is the punch animation, which is incredibly minor and should be fixed as opposed to not having the setting changed. That being said, pretty much all issues would be fixed with 70 fov + minmode (as I suggested before). Min mode is always better, because it hides most of the weapon while 70 FOV ensures that a weapon like the Shotgun doesn't end up becoming a tiny barrel.

ShowinGrowin:
It was not set further. I should had specified but the medigun example is just one of many potential animation problems that exist where you can see problems during a transition. They all usually only exist briefly. These issues are not present 100% of the time and either happen during the attack animation or when you pull it out. But I do think it would need to be fixed. While they are minimal, it is possible to notice them on occasion, and I think presentation wise it makes sense to hold that up to a higher standard. Would they be easy to fix, maybe, I'm sure people have already with mods, but overall there needs to be more polish. Which is technically less simple than just changing the value.

I'm not referring to the rest of the minigun being visible in general. Obviously it's meant to be visible briefly during the transition animations, but the point I'm trying to make is it's not meant to be visible there. The animation is designed for that part to specifically obscured in that animation. The only reason you can see it is because it's moved forward. Which results in what looks like a few black boxes in the corner of the minigun. It looks better the way its intended to be seen imo. My whole point is there's probably a bunch of little unintended things like this, that conflict with what they intended.

We can agree to disagree on whether or not that actually matters. But I believe all of this would need to be redone/fixed in someway to be more presentable. It doesn't matter to much when its a hidden option for advanced users, but if it was how the game looked normally, then I would hold all of these tiny issues up to more scrutiny.

I feel like we simply value different things considering you think minmode should be default as well. But the problem here is things like that are good options for you. But that's why they should exist as options. Because minmode changes the intended look of the game even further. I think the game should remain as close as possible to it's original intended art direction by default. Any deviations, even small should be options. A big aspect of my opinion here is purely down to the art aspect of this. Even if it works, something such as minmode would absolutely alter the visual direction of the view models. And it's fine if you personally enjoy it, but I'm thinking purely about what valve envisioned the game to look like originally. Which I think should be preserved. Some things are ok as options.

JoriKos:
> It was not set further. I should had specified but the medigun example is just one of many potential animation problems that exist where you can see problems during a transition. [...]

Not improving the default settings for a few, very rare instances of animation bugs which can already happen under normal circumstances is just ridiculous. You're being contrarian.

> but the point I'm trying to make is it's not meant to be visible there. The animation is designed for that part to specifically obscured in that animation.

Based on what? You're just making things up now and grasping at straws. Pointing at parts of the Minigun and saying it's not meant to be there is just straight up a fabrication because your point makes no sense.

> But the problem here is things like that are good options for you. 

Less space taken up by a weapon on the screen is an objective improvement. Minmode means you won't have the weapon take up space, 70 FOV means you will still see the weapon (which is also important).

> A big aspect of my opinion here is purely down to the art aspect of this. Even if it works, something such as minmode would absolutely alter the visual direction of the view models. And it's fine if you personally enjoy it, but I'm thinking purely about what valve envisioned the game to look like originally. Which I think should be preserved. Some things are ok as options.

Literally what are you even saying. Visual direction? Art aspect? You think part of the visual direction was to have your screen blocked by a weapon? Do you also run 70 field of view with 54 viewmodel field of view? I'm guessing you don't use fast switch either? No hit- or killsound? Do you run default HUD with no config as well? I sure hope so, otherwise you're not getting the real TF2 experience! If you want to have a reactionary vision of in-game settings where it must all remain as 2007 TF2's default settings then go ahead, but why are you dragging the rest of the community down with you? A new player doesn't care that it's not 'the original vision' (which is a made-up thing btw, no true Scotsman and all) if they get nauseous because of the default field of view. Are you gonna tell them it's actually how the game is supposed to be played? Please get a serious argument instead of reactionary kneejerk reactions, man. I don't see a point in even discussing this anymore.

ShowinGrowin:
The viewmodels are designed for 54. So anything that does look different isn't what you're intended to see. I don't need to make more arguments about what the intention is, I feel like that was already obvious. That part of the minigun isn't visible with the default settings. That's something you aren't meant to see in that specific animation. The viewmodel fov was merely a console command when these animations were made, so it makes sense that they didn't factor in how it would look when changed. I imagine they made sure that part was hidden because it doesn't look very good and looks like a couple of blocks on the side. But that's not a problem because you aren't meant to see it normally. It may not bother you, but it is something you cannot normally see in that specific animation.

This is even more so the case for the minmode viewmodels. That is not how the game was designed to look. The people who made those animations and models designed them for the a specific position. So by changing that you are heavily altering the intended design. It's fine if you don't mind or even prefer it but that doesn't mean it's objectively better.

It's not like I'm saying these shouldn't be options but the original devs didn't just slap them together. They designed them to look like this. So I think it's important to retain their original vision.

I could care less how an individual person prefers their settings. But the default settings should match how they wanted the game to look within reason. I understand that the viewmodel fov is a more minor thing (I'm just pointing out that it still has conflicts with their design. And sometimes small things do matter.) but minmode would be absolutely changing a lot. I think that would be a much bigger issue.

If you don't consider the original presentation important to preserve, or that screen space matters more, then that's just your opinion. But it's also exactly why valve is probably hesitant to change anything. (Even though I'd argree that some things are important to change.) Since nothing is truly objective. I would agree that some changes are good but some may also conflict with how the game is meant to look by the artists. Which is why they were optional. Minmode is a pretty big example of that.

If you don't think that's a real argument then I don't know what to tell you. We value different things.

AmandaRoseChaqueta:
For the model issues, it should be get a fix when this get merged. This new defaults improve the experience for most users that use 16:9 24 inches monitors or bigger.

PR #2080, opened 2026-09-14
[TF2] Enable sprays

num-get:
# Description
I like sprays. Since all the official Valve servers already disable sprays through sv_allowupload 0 command, I don't see why this new default setting should affect newer players that play on community servers.
The cfg/user_default.src needs to set cl_spraydisable to 0 as well after this PR is accepted.


KimmyTF2:
They could be reenabled ONLY if Valve added self-moderation options for players.
Like a server-wide vote to block sprays that are unacceptable to the majority of the server. Or have an option to hide individual sprays that you don't want to see.
Without any kind of moderation, sprays are doomed to be used as a shocker (mostly porn) rather than players using them to express themselves.

loporlp:
I don't think this is a wise choice, from the sprays I've seen I think it could be bad for a new player to enter a community server and see some of that stuff.

AmandaRoseChaqueta:
I would pretty much prefer for this be kept as opt-in. It should be user's choice if they want to enable them in the first place, cuz leaving it enabled by default can just allow people to post vile stuff. I don't think this PR should be merged.

PR #2087, opened 2026-09-15
[TF2] Fix MvM health upgrades for carried buildings

### LINKED ISSUE #8188: [TF2] [MVM] Teleport Health Multiplication Exploit
**Explanation**
There was an exploit I discovered out of curiosity.
Using a bind, it multiplies the health of a teleporter (either of the two) to over 1,270,000,000, with a limit determined by luck. At this point, it becomes very weak, and the skull symbol appears on the HUD upon death (though it's not visible).

**step-by-Step Instructions**
1. You must have a bind
to place and destroy the entrance teleporter.
2. Place both teleporters
when both are placed, the health is added or multiplied.
3. Grab the exit teleporter and use the bind to destroy the entrance.

4. Place the exit teleporter after grabbing it and placing the entrance. When there are two teleporters, they'll have 216 health. When you grab one and destroy the teleporter you want, it will retain its original health (216).
5. Repeat and you're done.

__(Valve, fix this before a YouTuber like Delfy tries it ;-;)__

SanyaKor:
# Fix MvM health upgrades for carried buildings

Fixes incorrect building health when buying or refunding Building Health while carrying a building.

Issue : [Youtube demo](https://youtu.be/jX4p5GgiIgg?si=UkmuicyKs3pqQf1q)

## Cause

Carried buildings temporarily reset to Level 1 for the redeployment animation. Their previous level is saved in `m_iHighestUpgradeLevel` and restored as they rebuild after placement.

However, `ApplyHealthUpgrade()` calculates health using the temporary Level 1, even when the saved level is 3.

## Why 450 HP instead of 648 HP?

A Level 1 building has 150 HP, while Level 3 has 216 HP. With the second Building Health upgrade (+200%):

```text
Incorrect: 150 + 150 * 2 = 450 HP (+200% from Level 1)
Correct:   216 + 216 * 2 = 648 HP (+200% from Level 3)
```

The third upgrade gives +300%: `216 + 216 * 3 = 864 HP` for Level 3.

## Change

Health upgrades now account for the building's saved level while it is being carried. A carried Level 3 building therefore receives the correct health for Level 3.

The saved pickup health is also updated when buying or refunding upgrades, so placing the building does not restore its old HP.

This applies to sentries, dispensers, and teleporters. Mini-sentries keep their existing health calculation and also receive the saved-health update.

## Related exploit

Also fixes the longstanding exploit where refunding Building Health while carrying any building could preserve its upgraded HP after placement until it was destroyed.

Additionally, it fixes an exploit that allowed teleporters to gain effectively unlimited HP. 
Fixes https://github.com/ValveSoftware/Source-1-Games/issues/8188


JoriKos:
Is there a reason to remove the commented out DevMsg?

SanyaKor:
> Is there a reason to remove the commented out DevMsg?

I used several dev message calls while debugging and removed them afterwards. Since this one was already commented out, I assumed that i could remove it as well to keep the code clean. I can restore it if needed .

HyperionCoding:
Could be worth looking into how these two interact together https://github.com/ValveSoftware/source-sdk-2013/pull/2026

